PR #213, opened 2024-04-09
fix: docker-compose without building

### LINKED ISSUE #211: Clarify docs for docker compose set up
Just checking but for Docker compose instructions, 
See https://github.com/grafana/explore-logs?tab=readme-ov-file#test-out-with-docker-compose

I see the following error, when attempting to spin up Grafana, Loki, and the Logs Explore App using the given compose file

```
[+] Building 0.0s (0/0)                                                                                                                                                                                     
unable to prepare context: path "./generator" not found
```
Can you double check this is correct? https://github.com/grafana/explore-logs/blob/main/docker-compose.yaml#L32
as I think there is an expectation that a generator directory already exists, that contains the Dockerfile file.

Thanks

svennergr:
We claimed that running `docker compose up` would not require the repo to be needed to checkout, but we were missing to use a pushed version of the `fake-logs-generator`. Fixing that by using a recent version I just pushed.

Fixes #211

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+mkdir -p /tmp/explore-logs/config
+mkdir -p /tmp/explore-logs/provisioning/datasources
+mkdir -p /tmp/explore-logs/provisioning/plugins
+
+# Download files
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/config/loki-config.yaml -o /tmp/explore-logs/config/loki-config.yaml
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/provisioning/datasources/default.yaml -o /tmp/explore-logs/provisioning/datasources/default.yaml
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/provisioning/plugins/app.yaml -o /tmp/explore-logs/provisioning/plugins/app.yaml
+
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/docker-compose.yaml -o /tmp/explore-logs/docker-compose.yaml
+
+docker compose -f /tmp/explore-logs/docker-compose.yaml up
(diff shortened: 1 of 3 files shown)


svennergr:
> Nice not seeing the generator issue anymore.
> 
> I see this, but it might be because I'm in an empty directory to test Screenshot 2024-04-09 at 11 08 07 AM

Ah, no, we are referencing the config file - great catch!

svennergr:
Needed to go a different route and use a shell script to just download the needed files.

scripts/run.sh
```diff
@@ -0,0 +1,14 @@
...
+mkdir -p /tmp/explore-logs/provisioning/datasources
+mkdir -p /tmp/explore-logs/provisioning/plugins
+
+# Download files
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/config/loki-config.yaml -o /tmp/explore-logs/config/loki-config.yaml
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/provisioning/datasources/default.yaml -o /tmp/explore-logs/provisioning/datasources/default.yaml
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/provisioning/plugins/app.yaml -o /tmp/explore-logs/provisioning/plugins/app.yaml
+
+curl https://raw.githubusercontent.com/grafana/explore-logs/main/docker-compose.yaml -o /tmp/explore-logs/docker-compose.yaml
+
```
L2D2Grafana, line 13: Locally I changed this to `cp ../docker-compose.yaml -o /tmp/explore-logs/` but I still see
Screenshot 2024-04-09 at 1 54 40 PM

svennergr: Should work with `cp ../docker-compose.yaml /tmp/explore-logs/`. Note the line in your output:
```
cp: -o: No such file or directory
```

L2D2Grafana: Yeah I might be having some strange permission issues according to google.

svennergr: Did you remove the `-o` from the `cp` command you ran?

L2D2Grafana (approved):
I think just my permissions are being weird creating directories

PR #221, opened 2024-04-11
PoC - Logs-app merge + table UI

gtk-grafana:
WIP: Adding the table from logs app to explore-logs.

image

The base functionality is working, but there's some UI fixes and changes we'll need to introduce.
1. Absolute time ranges take up too much space
image
2. Cannot differentiate between negative and positive filters
image
3. The filters take up a lot of space
image

I'm going to add a toggle to the oldschool logs panel next

@svennergr @L2D2Grafana I'd love to get this up on a dev instance so @zizzpudding can start playing around with it?


L2D2Grafana (approved):
Thank you for adding webpack and jest improvements! I like that this PR adds the table, but has an escape hatch back to the Log Lines. I'm unsure about the context/scenes state mapping but as I have no better ideas at this time for the sake of 'progress' 🚀

PR #222, opened 2024-04-12
Remove single cardinality graphs

gtk-grafana:
Fixes https://github.com/grafana/explore-logs/issues/130

Labels/detected fields with single cardinality cannot be added to filters, so let's remove any graph with a single non time field

FieldsBreakdownScene and LabelBreakdownScene were doing the same work in the buildAllLayout functions, I created a new parent class and moved the removeErrorsAndSingleCardinality and hideField functions to the new parent class.

KNOWN BUG: 
The labels count doesn't decrease like the detected fields does. Haven't figured out why yet.

### DIFF
--- a/src/components/Explore/LogsByService/Tabs/GenericBreakdownScene.tsx
+++ b/src/components/Explore/LogsByService/Tabs/GenericBreakdownScene.tsx
@@ -2,14 +2,14 @@ import { SceneCSSGridItem, SceneObjectBase, SceneObjectState, SceneQueryRunner }
 import { ALL_VARIABLE_VALUE } from '../../../../utils/shared';
 import { LoadingState, SelectableValue } from '@grafana/data';
 
-export interface AbstractBreakdownSceneState extends SceneObjectState {
+export interface GenericBreakdownSceneState extends SceneObjectState {
   labels: Array<SelectableValue<string>>;
   changeLabels?: (n: string[]) => void;
 }
 
-export abstract class AbstractBreakdownScene<
+export class GenericBreakdownScene<
   TState extends SceneObjectState = SceneObjectState
-> extends SceneObjectBase<AbstractBreakdownSceneState> {
+> extends SceneObjectBase<GenericBreakdownSceneState> {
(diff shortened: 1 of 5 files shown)


src/components/Explore/LogsByService/Tabs/GenericBreakdownScene.tsx
```diff
@@ -0,0 +1,39 @@
...
+        gridItem.setState({ isHidden: true });
+      } else if (result.data.state === LoadingState.Done) {
+        // Hide panels with single cardinality
+        if (result.data.series.length < 2) {
+          const val = result.data.series?.[0]?.refId;
+          gridItem.setState({ isHidden: true });
+
+          if (val) {
+            this.hideField(val);
+          }
```
L2D2Grafana, line 27: Still seeing No Labels for detected fields 
Screenshot 2024-04-12 at 11 40 39 AM I'd think the only change would need to be in the data response here and the rest could be left as is. I wonder if the label count doesn't change because they are there just hidden. @sven did something like this before I thought where he just remove data from the response  instead of hiding`const detectedFields = res.fields.filter((f) => !disabledFields.includes(f)).sort((a, b) => a.localeCompare(b));`

gtk-grafana: Yeah actually that would be a much better approach in general, we could just remove single cardinality/error dataframes at the top level after we do the extractFields transformation. That would fix some UX weirdness too with the graphs disappearing on load.

I'll close this and try that approach on monday! Thank you @L2D2Grafana

gtk-grafana: So that won't work as we found out in https://github.com/grafana/explore-logs/pull/226, the labels from the logs panel dataframe are only the labels of the log lines, not the complete set of labels for the query. So if we calculate the cardinality before we execute each sub-query aggregated by each label, we could be wrong, particularly in cases where the query result is a tiny subset

PR #223, opened 2024-04-13
chore: Remove duplicate flag

ystkfujii:
It appears to be configured in loki-config.
see : https://github.com/grafana/explore-logs/blob/e938af139fba465a9bd2fa135d0a9a8734fa567b/config/loki-config.yaml#L20-L21


cyriltovena (approved):
LGTM

PR #226, opened 2024-04-15
Remove single cardinality graphs

gtk-grafana:
Fixes https://github.com/grafana/explore-logs/issues/130

Labels/detected fields with single cardinality cannot be added to filters, so let's remove any graph with a single non time field

FieldsBreakdownScene and LabelBreakdownScene were doing the same work in the buildAllLayout functions, I created a new parent class and moved the removeErrorsAndSingleCardinality and hideField functions to the new parent class.

### DIFF
--- a/src/utils/fields.ts
+++ b/src/utils/fields.ts
@@ -29,29 +29,26 @@ type labelValue = string;
 export function getCardinalityMapFromFrame(frame: DataFrame) {
   const labels = frame.fields.find((f) => f.name === 'labels')?.values as Labels[];
 
-  const cardinalityMap = new Map<labelName, { valueSet: Set<labelValue>; maxLength: number }>();
-  labels.forEach((fieldLabels) => {
+  const cardinalityMap = new Map<labelName, { valueSet: Set<labelValue> }>();
+  for (let i = 0; i < labels.length; i++) {
+    const fieldLabels = labels[i];
     const labelNames = Object.keys(fieldLabels);
-    labelNames.forEach((labelName) => {
+    for (let j = 0; j < labelNames.length; j++) {
+      const labelName = labelNames[j];
       if (cardinalityMap.has(labelName)) {
         const setObj = cardinalityMap.get(labelName);
         const values = setObj?.valueSet;
-        const maxLength = setObj?.maxLength;
 
         if (values && !values?.has(fieldLabels[labelName])) {
           values?.add(fieldLabels[labelName]);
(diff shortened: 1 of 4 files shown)


src/utils/fields.ts
```diff
@@ -23,6 +23,39 @@ export function extractFields(data: DataFrame) {
   return result;
 }
 
+type labelName = string;
+type labelValue = string;
+
+export function getCardinalityMapFromFrame(frame: DataFrame) {
```
gtk-grafana, line 29: copy/paste from the table UI code, so it's doing some stuff right now that we don't need, like returning the length of the longest string for that field.

There is a performance cost with doing this, typically <3ms, and the table will need this map as well, so ideally we'd be able to expose the result of the calculation to the table

gtk-grafana: Cleaned it up, still seeing <3ms with worst case data, but it will save more then it uses by preventing more calls to the datasource for single cardinality time series panels, besides avgFields like 'duration', 'count', 'total', and 'bytes', which are included to allow time range selection.

gtk-grafana:
Actually it looks like we'll need to do this for indexed labels as well? The detected_labels api doesn't currently get passed the parsed filters, so if you have filter with a parsed/detected field the cardinality count won't be correct.

svennergr:
Couldn't we "just" count the frames of the metric query responses? Everything with more than 1 frame will be shown? I think, right now, you are basing the calc on the logs result, right? So that might not cover the whole timerange.

gtk-grafana:
Yeah but then we still have to make that call to the datasource, was hoping we could eliminate the number of requests we make

gtk-grafana:
@sven that was my original approach https://github.com/grafana/explore-logs/pull/222/files#diff-646ad8f3b71a7deb2c93974f312e6f7ff0880a97b13c5a341f85db95fba633dcR10

gtk-grafana:
But good point: If we only visualize labels that we get back from the initial log query as time-series, could we get a situation when running the same query but aggregated by the label, where the cardinality of the aggregated query is different then the labels we get back on the original dataframe? 

I think so: if we're seeing 1000 lines from 1000 different services with unique labels, we'd count the cardinality of everything as 1.

Ok, so yeah the usefulness of this feature scales with what percentage the returned logs are of the total result :/

gtk-grafana:
Makes incorrect assumptions about the data. Nice try me

PR #231, opened 2024-04-15
Clean up and unify subscriptions

gtk-grafana:
Some tidying up! Might be some performance benefit but I wasn't able to see any.

### DIFF
--- a/src/components/Explore/ByLabelRepeater.tsx
+++ b/src/components/Explore/ByLabelRepeater.tsx
@@ -44,6 +44,8 @@ export class ByLabelRepeater extends SceneObjectBase<ByLabelRepeaterState> {
       if (data.state.data) {
         this.performRepeat(data.state.data);
       }
+
+      return () => this._subs.unsubscribe();
     });
   }
 
--- a/src/components/Explore/ByLabelRepeater.tsx
+++ b/src/components/Explore/ByLabelRepeater.tsx
@@ -44,8 +44,6 @@ export class ByLabelRepeater extends SceneObjectBase<ByLabelRepeaterState> {
       if (data.state.data) {
         this.performRepeat(data.state.data);
       }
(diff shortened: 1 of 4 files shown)


src/components/Explore/ByLabelRepeater.tsx
```diff
@@ -44,6 +44,8 @@ export class ByLabelRepeater extends SceneObjectBase<ByLabelRepeaterState> {
       if (data.state.data) {
         this.performRepeat(data.state.data);
       }
+
+      return () => this._subs.unsubscribe();
```
gtk-grafana, line 48: RTFM https://grafana.com/developers/scenes/advanced-data#use-data-in-a-custom-object-class. you don't need to return any subs added to this._subs!

> The subscription returned from sourceData.subscribeToState is added to this._subs. Because of this, you don't need to do any cleanup when the custom object is destroyed, as the library will take care of unsubscribing.

So this sounds like as long as we add everything to this._subs we are good to go

src/pages/Explore/LogExploration.tsx
```diff
@@ -115,33 +115,36 @@ export class LogExploration extends SceneObjectBase<LogExplorationState> {
...
+        if (patternsVariable instanceof CustomVariable) {
+          const patternsLine =
+            newState.patterns?.map((p) => `${p.type === 'include' ? '|> ' : '!> '} \`${p.pattern}\``)?.join(' ') || '';
+          patternsVariable.changeValueTo(patternsLine);
+        }
+      })
+    );
 
     return () => {
+      this._subs.unsubscribe();
```
gtk-grafana, line 147: If we are returning a custom deactivation handler, do we also need to manually unsubscribe this._subs?

gtk-grafana:
Stale, closing

PR #232, opened 2024-04-16
Update installation requirements

cyriltovena:
Unfortunately this is not on by default in 3.0

This can't also be done via a flag only...


Vesther:
Thank you for the comments on the related issues and this PR @cyriltovena as i've also run into the missing of this setting!

I'd like to add that from my point of view these settings should maybe not be titled as "For an optimal experience" and rather clearly as required. That would be clearer guidance and avoid potential confusion such as mine 😅

cyriltovena (commented):
emphasis even more

Issue #67, opened 2024-03-28
nit: remove any legends that don't provide any additional context


matryer:
e.g. the **Value** legend here is unnecessary

<img width="1057" alt="image" src="https://github.com/grafana/loki-explore/assets/101659/6129e991-8154-46f2-95bb-30bb417c8a72">

svennergr:
I think we should actually also add a filter to only match for values in the metric queries powering the panels.

PR #233, opened 2024-04-16
Remove errors from detected field queries

gtk-grafana:
This fixes time series graphs in detected fields from showing a "Value" time series, which is the volume of the logs that are not aggregated by the label of the current query.

fixes: #67

Before:
image

After:
image


svennergr (approved):

src/components/Explore/LogsByService/Tabs/FieldsBreakdownScene.tsx
```diff
@@ -369,7 +369,7 @@ function buildNormalLayout(variable: CustomVariable) {
               body: new SceneReactObject({
                 reactNode: <LoadingPlaceholder text="Loading..." />,
               }),
-            })
```
gtk-grafana, line 372: I think webstorm has a bug where you cannot turn off autoformatting on save. Very frustrating

PR #236, opened 2024-04-17
Table: Refactor url state to use react-router 

gtk-grafana:
Syncing the selected columns between scenes and react was causing bugs, and scenes doesn't need to know which columns are selected within the react table. Using react router to manually set url state keeps viz specific logic out of scenes, and is easier to maintain

### DIFF
--- a/src/components/Table/ColumnSelection/ColumnSelectionDrawerWrap.tsx
+++ b/src/components/Table/ColumnSelection/ColumnSelectionDrawerWrap.tsx
@@ -15,7 +15,6 @@ export function getReorderColumn(setColumns: (cols: FieldNameMetaStore) => void)
     }
 
     const pendingLabelState = { ...columns };
-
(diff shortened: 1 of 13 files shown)


src/components/Table/ColumnSelection/ColumnSelectionDrawerWrap.tsx
```diff
@@ -40,6 +40,7 @@ export function getReorderColumn(setColumns: (cols: FieldNameMetaStore) => void)
 
 export function ColumnSelectionDrawerWrap() {
   const { columns, setColumns, setVisible, filteredColumns, setFilteredColumns } = useTableColumnContext();
+  const [searchValue, setSearchValue] = useState<string>('');
```
gtk-grafana, line 43: moved search state up so we can clear search text after a selection is made

gtk-grafana:
LGTM

PR #237, opened 2024-04-17
Table: Force panel re-render when timerange changes in scenes

gtk-grafana:


gtk-grafana:
So this fixes the bug where grabbing a link to a log line would have a stale timerange, but it causes a complete re-render of the table whenever the timerange is changed. If users use auto-refresh they'll get the UI state reset every time refresh.

gtk-grafana:
A workaround for this could be to set the absolute time range in the url, and pull it down in the react component

PR #238, opened 2024-04-18
chore: move/delete/rename files to get a better code organization

svennergr:
Housekeeping and quality-of-life changes to reorganize the project folder structure and better reflect the responsibility and rationale behind each component in relationship with the app.


matyax:
Would you mind adding a short description? 🙏 Thanks!

src/components/ByService/LogsVolumePanel.tsx
```diff
@@ -10,17 +10,17 @@ import {
   SceneQueryRunner,
   SceneDataTransformer,
 } from '@grafana/scenes';
```
matyax, line 12: Not sure I understand why LogsVolumePanel is inside ByService.

matyax: Also why ByService is outside of Explore, or maybe what does Explore mean in the hierarchy.

src/components/misc/FieldSelector.tsx
```diff
@@ -11,7 +11,7 @@ type Props = {
   onChange: (label: string | undefined) => void;
```
matyax, line 11: Misc also contains components, so it should be capitalized. I'd maybe consider UiElements or put them inside `components`, which should also be capitalized.

src/utils/fields.ts
```diff
@@ -2,7 +2,7 @@ import { DataFrame, PanelData } from '@grafana/data';
 import { DrawStyle, StackingMode } from '@grafana/ui';
```
matyax, line 2: "services" instead of utils? <3

src/pages/Explore/LogExplorationPage.tsx
```diff
@@ -1,15 +1,16 @@
...
+import { MainComponent } from '../../components/Main/MainComponent';
+import { newLogsExploration } from 'utils/utils';
 
 export const LogExplorationPage = () => {
   const [exploration] = useState(newLogsExploration());
 
   return <LogExplorationView exploration={exploration} />;
 };
 
-export function LogExplorationView({ exploration }: { exploration: LogExploration }) {
```
matyax, line 12: I would rename this ExploreLogs, because it's not Explore.

src/components/ByService/LogsVolumePanel.tsx
matyax: Further thinking, I would remove this `Explore` folder.

src/utils/fields.ts
ivanahuckova: <iframe src="https://giphy.com/embed/J336VCs1JC42zGRhjH" width="480" height="339" frameBorder="0" class="giphy-embed" allowFullScreen></iframe>

ivanahuckova: YES

src/components/Main/ServiceSelectionComponent.tsx
```diff
@@ -70,7 +61,7 @@ export class SelectStartingPointScene extends SceneObjectBase<LogSelectSceneStat
   });
   private _services: Record<string, ServiceItem> = {};
 
-  constructor(state: Partial<LogSelectSceneState>) {
+  constructor(state: Partial<ServiceSelectionComponentState>) {
     super({
       // $variables: state.$variables ?? getVariableSet(),
```
gtk-grafana, line 66: remove commented line?

L2D2Grafana (approved):
🚀 ty for this cleanup

src/components/ByService/LogsVolumePanel.tsx
svennergr: I removed the `Explore` from `components` completely. So it's just the "page" now. All separate components should be in `/components`.

It is in "ByService" because it's only used after you select a service. Basically not used in this main starting page.

src/components/misc/FieldSelector.tsx
svennergr: renaming `misc` to `UiElements`? Fine with that.

src/utils/fields.ts
svennergr: Yep, that's something I would've done next. Also inspect all those files in `utils` if they can be arranged better. Feels like too much right now.

src/pages/Explore/LogExplorationPage.tsx
svennergr: What do you mean by "this"? The route is called `explore` I think that's why the page is called like that.

src/components/misc/FieldSelector.tsx
matyax: I don't like it either honestly. Maybe I would let them just be under Components, or create a few folders that are more specific (misc is ~= utils).

For example, form or form controls:
- AddToFiltersButton
- FieldSelector
- GoToExploreButton
- SelectFieldButton
- ShareExplorationButton
- LiveTailCONTROL
- LayoutSwitcher
- LineFilter
- GiveFeedback (it's a button)

Ui/UIState/Messages
- GrotError
- StatusWrapper
-

src/components/ByService/LogsVolumePanel.tsx
matyax: I see. But there's also this:

imagen

LogsVolume there is also confusing. Everything after the initial page is "by service".

src/pages/Explore/LogExplorationPage.tsx
matyax: If, for some reason, we move the link from Explore / Logs to, I don't know, App / Logs, should we refactor the app?

I think it's mixing concerns.

src/components/ByService/LogsVolumePanel.tsx
svennergr: > Further thinking, I would remove this Explore folder.

Just not sure I understand it. The `/pages/Explore` folder? Should we remove `/pages` at all?

> I see. But there's also this:

But that's an old view, not from this PR. I deleted the `DetailsScene.tsx`.

matyax: > But that's an old view, not from this PR. I deleted the DetailsScene.tsx

True! Missed the -

> /pages/Explore

I would make it /pages/ExploreLogs, and then remove or rename the `src/Explore` folder. For renaming `src/App` would be a good candidate, or `src/ExploreLogs` and we follow the convention of sharing the name with the `pages/` folder. We can chat more about it later.

matyax:
Fine by me. I would also happily do a follow up with moar changes.

svennergr:
> Would you mind adding a short description? 🙏 Thanks!

Thank you 🙇

PR #241, opened 2024-04-18
chore: explore context

L2D2Grafana:
### Description
This is a POC trying to figure out how to use react context with scene state. Most of the magic 🪄 happens in `ScenesContext.tsx` by dumping the Scenes state in there and constantly updating via subscriptions. The goal was to get() the latest scenes state into context and also set() scenes state through context. Right now everything is in scenes components, but ideally as an application grows a react component could import the context and update scenes through that.  

1. Context updates whenever scenes state updates and components seem to always get the latest ✅
2. I was able to update scenes state by cloning the class add the new state, however I could not repeat this magic with the time range. 🚫

Is there a better way people have found to mix scenes state and with larger app react context?

Go to [http://localhost:3000/a/grafana-lokiexplore-app/explore?logId=&var-fields=&var-filters=service_name%7C%3D%7Ctempo-distributor&var-ds=gdev-loki&var-patterns=&var-lineFilter=&mode=logs&actionView=logs&var-logsFormat=%20%7C%20logfmt](http://localhost:3000/a/grafana-lokiexplore-app/explore?logId=&var-fields=&var-filters=service_name%7C%3D%7Ctempo-distributor&var-ds=gdev-loki&var-patterns=&var-lineFilter=&mode=logs&actionView=logs&var-logsFormat=%20%7C%20logfmt) and click a tab 'labels', 'detected fieds', 'patterns' 
Screenshot 2024-04-18 at 12 41 20 PM
The tab value changes when clicking on it by updating context
Screenshot 2024-04-18 at 12 41 07 PM

### DIFF
--- a/src/components/Explore/LogsByService/LogsByServiceScene.tsx
+++ b/src/components/Explore/LogsByService/LogsByServiceScene.tsx
@@ -1,5 +1,5 @@
 import { css } from '@emotion/css';
-import React from 'react';
+import React, { useContext } from 'react';
 
 import { DashboardCursorSync, GrafanaTheme2, LoadingState } from '@grafana/data';
 import {
@@ -18,6 +18,7 @@ import {
   SceneVariable,
   SceneVariableSet,
   VariableDependencyConfig,
+  sceneUtils,
 } from '@grafana/scenes';
 import { Box, Stack, Tab, TabsBar, useStyles2 } from '@grafana/ui';
 
@@ -51,6 +52,7 @@ import { GoToExploreButton } from './GoToExploreButton';
 import { GiveFeedback } from './GiveFeedback';
 import { renderLogQLLabelFilters } from 'pages/Explore';
 import { DetectedLabelsResponse } from '../types';
+import { ScenesContext } from '../../../context/ScenesContext';
 
 interface LokiPattern {
   pattern: string;
@@ -303,6 +305,7 @@ export class LogsByServiceScene extends SceneObjectBase<LogSceneState> {
   }
 
   updateFromUrl(values: SceneObjectUrlValues) {
+    console.log('Action view updateFromUrl :>> ');
     if (typeof values.actionView === 'string') {
       if (this.state.actionView !== values.actionView) {
         const actionViewDef = actionViewsDefinitions.find((v) => v.value === values.actionView);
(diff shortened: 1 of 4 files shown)


matyax:
> Why do we need to have context around the scene? Are there specific use cases you feel aren't supported by the scene state management that requires a separate context?

I haven't seen the PR but can provide extra color about possible reasons behind it, in addition to this PR being a POC.

The quid of the question is around:

> aren't supported by the scene state management that requires a separate context

One of the biggest sources of friction around Scenes emerges from the situation where everything either needs to be a Scenes object or depends on behaviors that are abstracted and provided by Scenes. This creates a framework-within-a-framework situation where, in what is supposed to be a React environment, becomes Scenes or depend on Scenes.

> Why do we need to have context around the scene? 

My answer would be because, at the end of the day, this is a React application, where React components should be able to communicate with each other using a common language, in this case the Context API. Encapsulating state inside Scene object instances creates a new and diffuse contract where the internal life cycle is, for the lack of a better term, a mistery.

Another pattern that we've seen is storing state or context inside variables, and specifically invisible variables. In my opinion that is not a good approach, where variables are required to do a lot of heavy lifting that is not part of its purpose (key => value pairs for interpolation), creating an anti-pattern and another diffuse contract that developers need to interact with.

---

To be clear, I think your question was great and completely fair; we ourselves need to ask more of those kinds of questions. I just wanted to share some "context" and background behind our rationale.

In any case, I my team needs to spend some time with experiments and POC like this one, to further explore Scenes its surface of contact with React and other components, maybe discovering new patterns to work with it or around it.

L2D2Grafana:
> Why do we need to have context around the scene? Are there specific use cases you feel aren't supported by the scene state management that requires a separate context?

We have done a lot of work on components that used react context heavily [table poc](https://github.com/grafana/explore-logs/pull/221/files#diff-d1e5b9a9c52d84c8931bfa98c0d195e88abfb11541cdb8dbec0783093fdced02 ) and we're looking for easy ways to link things without a full rewrite. It's good to know from your perspective it would be better to put all those context values into a scene class component state and create something like a scenes table instead.

L2D2Grafana:
Thanks for the feedback everyone, closing for now!

Issue #258, opened 2024-04-19
Landing page wrapping issue


cyriltovena:
On rare occasion and very weird ration the landing page doesn't wrap nicely

<img width="746" alt="image" src="https://github.com/grafana/explore-logs/assets/1053421/d7b620a9-63cb-4d1d-a011-d7d70df1bc9c">

We most likely could do better.

PR #261, opened 2024-04-22
fix: landing page wrap

L2D2Grafana:
### Description
- #258 
Set the width to 100% in md breakpoint. Setting the height or minHeight did not seem to work, but it would have been nice to make them taller. 

Screenshot 2024-04-22 at 2 54 25 PM

- [Scenes breakpoints docs](https://github.com/grafana/scenes/pull/156/files#diff-0619e5bc001b9b055c71b89fc9f9f9638fb7373bdd5f9ceaeaa03161216f46e7)


matyax:
I think we noticed this during some demos today. Good catch!

matyax (approved):

PR #262, opened 2024-04-23
feat: volume 404 grot error

L2D2Grafana:
### Description
- #246 
Screenshot 2024-04-24 at 8 12 41 AM

 - lmk if there is better copy for this 😉


svennergr (approved):
🚀

PR #267, opened 2024-04-29
Services: refactor and reorganize

matyax:
Series of refactors to resolve ambiguity in the current implementation, providing more clarity of purpose and intent, and helping future refactors.

- Fields: renamed functions
- Plugins: unused, removed
- Shared: moved types to the corresponding usage places and renamed to variables
- Types: dissolved module, as types are defined close to their source

Bonus track:
- Return a redirect instead of `null` when the user has no permissions (they will otherwise just see an empty screen).
- Moved `pages` inside `Components`


L2D2Grafana (approved):
LGTM I like the tabs and Redirect 👍

PR #268, opened 2024-04-29
chore: change GAR to `prod`

svennergr:
We won't use the `dev` environment for now. Going to destroy any allocated resources in a PR in deployment_tools.


L2D2Grafana (approved):
👍

PR #286, opened 2024-05-03
feat(service-selection): disable log details

svennergr:
Disable log details on service selection screen. Given the small size of the logs panel, I don't believe it's useful and rather misleading, as scrolling in the logs panel is difficult with it being enabled.


matyax (approved):
Agree

PR #291, opened 2024-05-06
feat(query-tracking): add `supportingQueryType` header

svennergr:
image

Adds the plugin ID as `supportingQueryType` header, which translates into the `X-Query-Tags` in Grafana, and then in the `source` label in Loki (see screenshot).

NB: this does only work for "queries" not for resource calls for now.


cyriltovena (approved):
LGTM

PR #303, opened 2024-05-07
Chore: rename function and remove duplicated code

matyax:
Follow up of https://github.com/grafana/explore-logs/pull/301


matyax:
https://github.com/grafana/explore-logs/pull/301#discussion_r1592148218

My thoughts on that:

We could, but then the function signature would be wrong, because why would you call it with an undefined parameter. I would consider it a bad practice in the function definition.

I understand the purpose, but the correct sematic approach, IMHO, should be to not call something when it should not be called, rather than have the function do nothing when nothing is passed.

matyax:
In other words: is not the function responsibility that you're passing a potentially undefined value to it. We, the users or consumers, should not be calling it if we don't have anything to pass.

ivanahuckova (approved):

PR #311, opened 2024-05-07
feat(queryExpressions): add `buildBaseQueryExpression` to create the base LogQL expression

svennergr:
`buildBaseQueryExpression` will create the basic LogQL expression instead of relying on AdhocFilters to do the trick. This allows to more easily adapt expressions where needed.

The PR also renames the `VAR_FILTERS` to `VAR_LABELS` to make it clear that that variable stores all indexed labels. 

Additional side effect: this fixes the "Go to Explore" button, that did not include `| logfmt` in some cases.

Fixes https://github.com/grafana/explore-logs/issues/257

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -24,7 +24,7 @@ import {
   explorationDS,
   VAR_DATASOURCE,
   VAR_FIELDS,
-  VAR_FILTERS,
+  VAR_LABELS,
   VAR_LINE_FILTER,
   VAR_LOGS_FORMAT,
   VAR_PATTERNS,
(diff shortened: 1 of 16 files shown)


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -303,9 +305,10 @@ function getStyles(theme: GrafanaTheme2) {
       '&:first-child': {
         // The wrapper of each filter
         '& > div': {
-          // the 'service_name' filter wrapper
+          // wrapper for all indexed labels
           '&:nth-child(2) > div': {
-            gap: 0,
+            // remove the `gap` from the last label
+            'margin-right': theme.spacing(-2),
```
gtk-grafana, line 311:
```suggestion
            'marginRight': theme.spacing(-2),
```
Kebab-case not supported warning

svennergr: Done in [d46bc6b](https://github.com/grafana/explore-logs/pull/311/commits/d46bc6bb779e6e83eeb606ba9ea87874c02c4478)

svennergr:
> Summarizing our slack thread: Seeing a host of bugs with queries not being updated when variables they depend on have changed, for example the logs volume histogram not updating when the text query is updated, etc.
> 
> Sven mentioned we might be able to fix this with _variableDependency on the scene classes.

thanks galen. I'll continue to look into that.

PR #313, opened 2024-05-08
refactor(IndexScene): move LogsExplorationScene to a module and rename as LayoutScene

matyax:
This PR extracts `LogsExplorationScene` out of `IndexScene`, and renames `LogsExplorationScene` to `LayoutScene`. The reason for the new name is that `LayoutScene` contains the header contents (Variable controls + pattern filters) and a variable body, which can be the `ServiceScene` or the `ServiceSelectionScene`, so there's no "logs exploration", but rather a combination of fixed contents + a variable body.

Other changes:
- Renamed a few `onActivate()` methods and removed the private underscore suffix. It used to be a mix of with and without underscore, and I don't think this is a pattern that we'd like to follow.
- Within `LayoutScene`, instead of accessing `IndexScene` internal state thorugh `sceneGraph`, we now use `model.parent.useState/setState`.

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -74,7 +74,7 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
       ...state,
     });
 
-    this.addActivationHandler(this._onActivate.bind(this));
+    this.addActivationHandler(this.onActivate.bind(this));
   }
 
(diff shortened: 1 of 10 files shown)


matyax:
@grafana/observability-logs Mind taking a look? It's a relatively simple changeset.

ivanahuckova (approved):
nice change!

PR #343, opened 2024-05-16
fix(patterns): only allow a single include pattern

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/218

This PR does 2 things:
1. When a pattern is already excluded, it will hide the exclude pattern button. Same with included patterns
2. When a pattern is included, all other patterns are removed from state.

https://github.com/grafana/explore-logs/assets/109082771/2e8afe1d-53d1-4933-a0fa-fb0ec1b0b3ca


matyax (approved):
Great improvement! I wonder if users will understand that you can only have one active pattern filter active at a given time.

PR #344, opened 2024-05-16
feat(serviceIndex): add crosshair sync to panels within service index

gtk-grafana:
Fixes https://github.com/grafana/explore-logs/issues/207


matyax (approved):
👌

PR #345, opened 2024-05-17
fix(ServiceSelectionScene): disable hover tooltips

matyax:
This PR disables the hover tooltips that are shown in the following video:

https://github.com/grafana/explore-logs/assets/1069378/efc0e1f9-7c36-4dc9-b09a-f7b116dfbcae

The reason is that the behavior and information it shows is potentially confusing, and the tooltip is very hard to dismiss.

### DIFF
--- a/src/Components/ServiceScene/ServiceScene.tsx
+++ b/src/Components/ServiceScene/ServiceScene.tsx
@@ -444,7 +444,7 @@ const MAIN_PANEL_MAX_HEIGHT = '30%';
 function buildGraphScene() {
   return new SceneFlexLayout({
     direction: 'column',
-    $behaviors: [new behaviors.CursorSync({ key: 'metricCrosshairSync', sync: DashboardCursorSync.Crosshair })],
+    $behaviors: [new behaviors.CursorSync({ key: 'logsCrosshairSync', sync: DashboardCursorSync.Crosshair })],
     children: [
       new SceneFlexItem({
         minHeight: MAIN_PANEL_MIN_HEIGHT,
--- a/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
+++ b/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
@@ -22,6 +22,7 @@ import {
   LegendDisplayMode,
   LoadingPlaceholder,
   StackingMode,
+  TooltipDisplayMode,
   useStyles2,
 } from '@grafana/ui';
 import { getLokiDatasource } from 'services/scenes';
@@ -237,6 +238,9 @@ export class ServiceSelectionComponent extends SceneObjectBase<ServiceSelectionC
           calcs: ['sum'],
           placement: 'right',
           displayMode: LegendDisplayMode.Table,
+        })
+        .setOption('tooltip', {
+          mode: TooltipDisplayMode.None,
         })
         .setHeaderActions(new SelectServiceButton({ service }))
         .build(),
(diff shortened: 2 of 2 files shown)


svennergr:
But now there are no tooltips at all hovering the graphs.

matyax:
That is true, although at this point, anything is better than the current behavior. I'll see if I can spot the point where the tooltips started to appear when hovering the logs, which is not happenning in the version deployed to OPS.

matyax:
Couldn't find any way to control this the app, but it's currently very broken:

https://github.com/grafana/explore-logs/assets/1069378/e7a49d0a-fce4-4a02-aaec-b1a958a57497

I'm not sure where to report this.

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -246,10 +247,15 @@ export class ServiceSelectionComponent extends SceneObjectBase<ServiceSelectionC
   // Creates a layout with logs panel
   buildServiceLogsLayout(service: string) {
     return new SceneCSSGridItem({
+      $behaviors: [new behaviors.CursorSync({ sync: DashboardCursorSync.Off })],
```
matyax, line 250: @svennergr Disabled it using this. Took inspiration from here: https://github.com/grafana/explore-logs/blob/main/src/Components/ServiceScene/ServiceScene.tsx#L447

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -444,7 +444,7 @@ const MAIN_PANEL_MAX_HEIGHT = '30%';
 function buildGraphScene() {
   return new SceneFlexLayout({
     direction: 'column',
-    $behaviors: [new behaviors.CursorSync({ key: 'metricCrosshairSync', sync: DashboardCursorSync.Crosshair })],
+    $behaviors: [new behaviors.CursorSync({ key: 'logsCrosshairSync', sync: DashboardCursorSync.Crosshair })],
```
gtk-grafana, line 447: better name, but what are we trying to sync since this is only a single panel?

gtk-grafana (approved):
Can confirm this fixes the bug locally. Thank you!!

src/Components/ServiceScene/ServiceScene.tsx
matyax: It has no effect. Removing it causes the same bug we see in the service selection page, so what I would do is to also set it as `DashboardCursorSync.None`.

This comes from the hackathon: https://github.com/grafana/hackthon-2024-03-logger-info/blob/main/src/components/Explore/LogsByService/LogsByServiceScene.tsx#L349

@adrapereira Do you remember if we have any reason to keep this?

gtk-grafana: I'd set it to none, or figure out if we can sync crossHairs between logs volume and the logs panel (but in another issue/PR), but I'm not sure that would work since the buckets from the aggregation wouldn't line up with the timestamps on the logs?

matyax: Agree. I'll do it in a follow-up because, for some reason, setting it to `Off` somehow enables it.

matyax: Ahhh good point. I never payed attention to it, I'll see if that actually ever happens. Thank you!

matyax: Yep, it's adding a crosshair between Logs Volume and the other panels. Keeping it then, thank you again @adrapereira

PR #346, opened 2024-05-20
fix(LogsVolumePanel): remove layout wrappers and superfluous height code

matyax:
Looking at the behavior of the Logs Volume Panel while resizing, you can see that it disappears from the DOM below a certain width:

https://github.com/grafana/explore-logs/assets/1069378/c407600e-58b8-49c2-96af-24ac0d836605

It happens in code that we can't control from the application, so after many attempts (from changing widths attributes to change from Flex to Grid layout), I realized that we were wrapping the panel with an unnecessary FlexLayout and FlexItem. Removed.

Also found some code setting heights in a very sketchy way, which didn't seem to have any visual effect or regression.

Sections and reloads:

https://github.com/grafana/explore-logs/assets/1069378/dc8d903c-90bc-4fac-b486-5e60f77ed23b

Empty state:

https://github.com/grafana/explore-logs/assets/1069378/1f383ae8-488e-4c6f-8bbc-727493ca8fe5


gtk-grafana (approved):
Awesome and thanks for figuring this out, this was driving me batty!

PR #347, opened 2024-05-20
Patterns: change label to singular and increase preview size

matyax:
After https://github.com/grafana/explore-logs/pull/343 we no longer allow multiple pattern filters. For that, in this PR:
- Added `lg` and `sm` sizes to `PatternTag`
- Use `lg` "Pattern" when filtering by a single pattern
  Pattern
- Use `lg` "Excluded pattern" when excluding a single pattern
  Excluded pattern
- Use `sm` "Excluded patterns" when excluding multiple patterns
  Excluded patterns
- Define the `lg` pattern tag preview size based on the screen size. The chosen value seems to nicely behave under different viewport sizes:

https://github.com/grafana/explore-logs/assets/1069378/82d9e59c-f028-4bdd-b444-df007146436b

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@swc/jest": "^0.2.26",
     "@testing-library/jest-dom": "6.1.4",
     "@testing-library/react": "14.0.0",
+    "@testing-library/user-event": "^14.5.2",
     "@types/jest": "^29.5.0",
     "@types/lodash": "^4.14.194",
     "@types/node": "^20.8.7",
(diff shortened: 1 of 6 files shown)


matyax:
Iterated a bit on this PR to better fit the current implementation of pattern filtering.

gtk-grafana (approved):
Looks good and works well locally. The changes in https://github.com/grafana/explore-logs/pull/342 might conflict with the e2e tests, but it should be quick fixes. Lemme know if you run into any trouble

PR #348, opened 2024-05-22
fix(LineFilter): keep track of search length and attributes but not contents

matyax:
I was looking at usage tracking, and realized we were storing the contents of the line filter searches. We don't even store original queries, which are [obfuscated](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/tracking.ts#L185). Ammending that in this PR.


gtk-grafana (approved):
Good catch!

PR #369, opened 2024-05-31
feat(move-log-volume): moves log volume panel inside logs tab

svennergr:
Moves the log volume panel inside the logs tab.

https://github.com/grafana/explore-logs/assets/8092184/eac90eeb-5582-465a-8838-b708415d34f9


matyax (approved):
Nice change 👍

PR #377, opened 2024-06-03
fix(patterns): fix bug introduced by using viewport units in styles

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/373

Small viewports will scroll horizontally instead of pushing out the container width.

To reproduce on main: Find a really long pattern without any spaces to break, or go into the inspector and change the content of a pattern string within the table to something really long without any spaces.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/PatternsViewTableScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/PatternsViewTableScene.tsx
@@ -39,6 +39,7 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
     });
   }
 
+  // The component PatternTableViewSceneComponent contains hooks, eslint will complain if hooks are used within typescript class, so we work around this by defining the render method as a separate function
   public static Component = PatternTableViewSceneComponent;
 
   /**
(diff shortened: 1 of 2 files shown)


src/Components/ServiceScene/Breakdowns/PatternsViewTableScene.tsx
```diff
@@ -40,35 +41,17 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
...
-    const columns = model.buildColumns(total, appliedPatterns);
-
-    return (
-      <div data-testid={testIds.patterns.tableWrapper} className={renderStyles}>
-        <InteractiveTable columns={columns} data={tableData} getRowId={(r: WithCustomCellData) => r.pattern} />
-      </div>
-    );
-  }
+  // The component PatternTableViewSceneComponent contains hooks, eslint will complain if hooks are used within typescript class, so we work around this by defining the render method as a separate function
+  public static Component = PatternTableViewSceneComponent;
```
matyax, line 45: <3

matyax (approved):
Tested it and looks good.

PR #378, opened 2024-06-03
Revert "fix: skip all e2e tests"

gtk-grafana:
Reverts grafana/explore-logs#376

Turn e2e tests back on, removes hardcoded strings that were changing since the datagenerator used in CI is returning different patterns then the generator ran locally.


matyax (approved):
👌 👌

PR #381, opened 2024-06-03
fix: use content irrelative unit to get more consistent calculations

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/380

Note for reviewers: As far as I can tell this doesn't change the layout in chrome/ff, but it fixes the bug where it looks like safari is calculating a different 70% then the other browsers

image


svennergr (approved):
nice!

PR #383, opened 2024-06-03
chore(patterns): update percent column to include total count

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/382

image


svennergr (approved):
should we align the text to the right? i think it makes it a bit easier to read. after that, lgtm

gtk-grafana:
@svennergr good catch, fixed 
image
I also enabled numeric sort on that column while I was in there.
image

PR #392, opened 2024-06-04
Chore: Refactor patterns breakdown

gtk-grafana:
Having a new container gets rid of the need for useMeasure hook, which reduces the number of re-rendering we're doing in scenes. Also this tees up further work in the patterns view like text search.,

Shouldn't be much change in functionality, although now the column headers are sticky when scrolling through the patterns, and the header has been removed from the main pattern timeseries up top.

Double vertical scrollbar is still possible, for devices with small screen heights, and when many filters are applied pushing down the top of the page. Perhaps in https://github.com/grafana/explore-logs/issues/396 we should address this via measuring the height of the sticky header, or by adding the ability for the user to manually resize how much of the viewport height is dedicated to each section.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/PatternsBreakdownScene.tsx
@@ -1,33 +1,25 @@
 import { css } from '@emotion/css';
 import React from 'react';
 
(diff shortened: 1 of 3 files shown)


gtk-grafana:
Need to review with @zizzpudding about any style/design changes, but this should make it a bit easier adding more complex functionality to the patterns

matyax (approved):
Tried it locally and looks great!

PR #409, opened 2024-06-05
feat(LabelBreakdownScene): display an error message when the request fails

matyax:
Closes https://github.com/grafana/explore-logs/issues/402

Message:

imagen

Open for suggestions for the error message. The important part is that:
- It should let the user know what happened.
- It should offer at least one workaround.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
@@ -20,9 +20,9 @@ import {
   SceneVariableSet,
   VariableDependencyConfig,
 } from '@grafana/scenes';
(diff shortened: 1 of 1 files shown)


matryer:
nice @matyax - this is an improvement. I will open a separate issue to discuss auto-retries.

matryer (approved):
LGTM

svennergr:
Just a nit, but do we have a way to make the `Alert` less alerty and more warningy? What I mean is less red, more orange/informative?

matyax:
Good idea. Made it a warning instead of error.

matyax:
Good idea. Made it a warning instead of error. We should keep this type in mind to be consistent.

svennergr (approved):

PR #410, opened 2024-06-05
Service Search: Use search input to query volume API

svennergr:
Added a quick feature to use the input in "Search services" to be used in the Volume API request. This enables users to search for their exact service name.

https://github.com/grafana/explore-logs/assets/8092184/3f6d6e86-9852-43ef-b2c1-a82ebe3d6fba


cyriltovena (approved):
LGTM

PR #411, opened 2024-06-06
fix(service-selection): only show favorite services if there is data

svennergr:
"Favorite" or rather previous selected services will be shown at the top of the service selection. This PR fixes a behavior where favorite services are shown even if they are not included in the response of the /volume API:

image


matyax (approved):
Great catch!

PR #418, opened 2024-06-06
feat(breakdowns): refactor top section

matyax:
Closes https://github.com/grafana/explore-logs/issues/398

This PR refactors breakdowns by:
- Removing field labels in general for cleaner UI and saving space (`By Field` and `View`)
- Adopting `Select` as the default field/label selector.

Additional change: improved accessibility of the label and fields "select" buttons.

Demo:

https://github.com/grafana/explore-logs/assets/1069378/dfa57b2c-7b9c-453c-a6a8-c36987c3092a

Labels:

Labels landing

Fields:

Fields landing

Empty states:

No labels

No fields

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/FieldSelector.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldSelector.tsx
@@ -2,8 +2,7 @@ import { css } from '@emotion/css';
 import React from 'react';
 
 import { GrafanaTheme2, SelectableValue } from '@grafana/data';
-import { Select, useStyles2, InlineField, Stack, IconButton } from '@grafana/ui';
-import { ALL_VARIABLE_VALUE } from 'services/variables';
+import { Select, useStyles2, InlineField } from '@grafana/ui';
 
 type Props = {
   options: Array<SelectableValue<string>>;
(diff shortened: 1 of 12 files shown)


gtk-grafana:
When I have "All" selected in a breakdown, I don't see the dropdown.
image

But it shows in the label value breakdowns:
image

matyax:
@gtk-grafana That's a good point. I'll iterate on that, thanks!

matyax:
@gtk-grafana What do you think?

https://github.com/grafana/explore-logs/assets/1069378/2ef3b705-f3a3-430d-a8db-94da976cf6fc

matyax:
Total reset of this PR. Please check the issue description for the latest snapshot of the changes.

Thanks @gtk-grafana for the feedback, as I originally intended to remove the selector from the "all" views.

gtk-grafana (approved):
Awesome, yeah this is what I expected, now we can even search labels/detected_field names by name in the dropdown if users know what they are looking for, and don't want to scroll through results to find it!

My only note, but we probably want to address this in another issue/PR
image

Is it would be awesome if we could have the same behavior for values (crappy mockup)
image
Where users could search the values in dropdown/textsearch if there are many results.
However this gets more complicated when negative filters are working, so yeah, let's not address it in this PR.

LGTM

matyax:
Oh yes, it would be cool to have something for values. I can follow this up with something, great idea.

PR #424, opened 2024-06-07
fix(FieldsBreakDownScene): add loading state and empty state

matyax:
Fixes https://github.com/grafana/explore-logs/issues/393

This PR:
- Adds a missing loading state to Detected Fields
- Adds an empty state to Detected fields

Message

https://github.com/grafana/explore-logs/assets/1069378/4a132ace-561f-4437-85aa-7c8c1f2bf0fe

### DIFF
--- a/src/Components/ServiceScene/ServiceScene.tsx
+++ b/src/Components/ServiceScene/ServiceScene.tsx
@@ -68,6 +68,8 @@ export interface ServiceSceneState extends SceneObjectState {
   patterns?: LokiPattern[];
 
   detectedFieldsCount?: number;
+
+  loading?: boolean;
 }
 
 export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
@@ -81,6 +83,7 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     super({
       body: state.body ?? buildGraphScene(),
       $data: getQueryRunner(buildLokiQuery(LOG_STREAM_SELECTOR_EXPR)),
+      loading: true,
       ...state,
     });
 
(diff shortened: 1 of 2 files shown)


matyax:
The "by field" will be gone in https://github.com/grafana/explore-logs/pull/418

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -81,6 +83,7 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     super({
       body: state.body ?? buildGraphScene(),
       $data: getQueryRunner(buildLokiQuery(LOG_STREAM_SELECTOR_EXPR)),
+      loading: true,
```
gtk-grafana, line 86: instead of passing a new loading prop, couldn't we just grab the $data state from the ServiceScene in the child scene, and then use the `LoadingState` instead of a bool?

matyax: This boolean is what makes the missing `loading` message work.

gtk-grafana: The `loading` prop is only set to false when the data state is `LoadingState.Done` or `LoadingState.Error`, why do we need a new prop when we could check for these data states in the child?

matyax: It is possible, but I don't see what it would solve. What would be the advantage of the children holding implementation details of the parent instead of exposing a bolean through the state. Seems like more propagation of an antipattern.

matyax: In addition, these `loading` booleans also control the display of loading indicators.

gtk-grafana: Gotcha, I was just thinking we should get the loading state directly from the queryRunner that's executing the query so there's a single source of truth, and then if we wanted to differentiate error states in the children, like error vs done vs loading we could, but I'll defer to your expertise on this since it sounds like you've dug in a bit more then I have :)

gtk-grafana (approved):
LGTM, but I'd suggest checking with design or docs to double check the new copy in the message

src/Components/ServiceScene/ServiceScene.tsx
matyax: It's not a bad idea at all, but right now I just don't see any advantage on changing it. I specially dislike having to know that your parent has a $data attribute that is the query and it has states, etc.

matyax:
Good idea. Last version of this new message:

imagen

PR #430, opened 2024-06-07
Patterns: Logs sample

gtk-grafana:
This cannot be released until 11.1.1 is released, and the dependencies have been updated due to [conflict in InteractiveTable](https://github.com/grafana/grafana/pull/88948#pullrequestreview-2109724671) with the logs panel. 
Users running < 11.1.1 will see the logs table all ****ed up, so we might want to wait to deploy this until we can deprecate those versions of Grafana

Aug 01, 2024 Update: Looks like all cloud instances are on some version of 11.1.1 or higher now, so we should be good to release this now.

Functionality:
This got a bit complicated to try and work around the fact that patterns only filter on indexed filters in the users query context.
We first attempt to run the pattern sample query with all of the line filters and field filters. If that comes back with no results, we show a warning message that the patterns being displayed do not intersect with your current query results, and we remove the ability to include/exclude by that pattern.

image

We also prompt the user to clear their current filters, in case they do want to filter by that pattern. Unfortunately this will probably route the user away from the patterns view (unless we are just clearing the line filter), as right now we always route to the logs tab after change of variables. We should deal with this in another issue/PR: (https://github.com/grafana/explore-logs/issues/640).

Ideally we would not show patterns that don't apply to the users current filters, but that needs to be addressed in the loki API: https://github.com/grafana/explore-logs/issues/641.

If something else goes wrong: we show an actionless error state, and log the request and traceIds to console.error (`Pattern sample query returns no results`), so we can use faro logs to later debug what's going on. We probably want to add a panel to our tracking dashboard for this error.
image

If the active filters do not exclude all results, and the first query comes back successful, we allow users to filter by this pattern, and the results are filtered by the users current query context.
image


cyriltovena:
I would not apply line filters for showing sample of patterns !

matyax:
Will be taking a look at this one today. Apologies for taking so long.

src/Components/ServiceScene/Breakdowns/PatternsLogsSampleScene.tsx
```diff
@@ -0,0 +1,265 @@
...
+      lineFilterVariable.changeValueTo('');
+
+      const noticeFlexItem = this.getNoticeFlexItem();
+
+      // The query we just fired is already correct after we clear the filters, we just need to hide the warning, and allow filtering
+      noticeFlexItem?.setState({
+        isHidden: true,
+      });
+
+      this.removeThisPatternFromFilterExclusion();
```
matyax, line 109:
```suggestion
      this.removePatternFromFilterExclusion();
```

matyax (approved):
Is this tooltip coming from the UI library?

imagen

I initially thought there was a missing loading state, so I was going to suggest one, but it turns out with these many patterns (300) the performance is awful. Adding the logs panel takes a significant amount of time. Not sure if there's anything we can do now, maybe something to follow up.


https://github.com/user-attachments/assets/e0b27531-b7b4-4b4b-a059-5a943bd333a8

Besides these comments and other nitpicks, it works very well, great job!

gtk-grafana:
@matyax thanks for taking a look, the tooltip is indeed in the library, it's pretty obnoxious, but it should be relatively quick to update the base component to allow passing in of a custom title tag.

For performance, it would be nice to build our own scenes table that can be virtualized, but I agree that should be addressed in another PR

I looked again into why we didn't have a loading state, and it's because we need to render the logs panel to subscribe to the query runner, which executes the query, so the panel is the loading state, it just takes a while to render. Maybe worth revisiting in the future.

gtk-grafana:
I'll merge this in on Monday. CC: @svennergr @matyax

PR #444, opened 2024-06-12
Bug: Table - Add indexed labels to stream selector

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/443

Also removed unused filters from QueryContext, we might want to add them back in when negative filters are supported to prevent people from seeing the context menu item to add a filter that's already been added.


matyax (approved):
👌

PR #453, opened 2024-06-13
Logs panel: allow filtering from log lines

matyax:
This PR enables log line filtering for the Logs and Table visualizations, using a recently updated LogsPanel API.

- Reused and extracted `AddToFiltersButton` logic.
- Added `getLabelTypeFromFrame()` to add filters to the appropriate filter type.

Demos:

https://github.com/grafana/explore-logs/assets/1069378/f69e6c92-d5c9-4f2e-aeb5-16f9cf66acc5

Only works for the latest latest Grafana, but degrades gracefully if not supported (i.e. buttons are not shown).


gtk-grafana:
Removing the filters isn't quite working as expected.
* Removing the service name will break the query, but the variable still shows up in the UI. Have to refresh to fix.
* Negative filters show up as positive filters in the UI, when clicking on the "remove filter" it will add a positive filter, and not remove the negative filters

matyax:
Great catches @gtk-grafana .

See:

https://github.com/grafana/explore-logs/assets/1069378/24acfb47-f0d3-4b51-bddb-38b87cf181a5

Changing service_name is now a noop. We need to flash a message to the user but I didn't find a way to trigger notifications from apps.

matyax:
Changelog:
- Unified the implementation with the table
- Added comments for the filter types
- Refactored `reset` to `clear` for clarity

gtk-grafana:
Working much better now! Do we want to update the title tags? 
image

src/Components/FilterButton.tsx
```diff
@@ -6,14 +6,14 @@ import { testIds } from 'services/testIds';
 
 type Props = {
   onInclude: () => void;
-  onReset: () => void;
+  onClear: () => void;
```
ivanahuckova, line 9: nice!

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -16,6 +16,11 @@ export interface AddToFiltersButtonState extends SceneObjectState {
 
 export type FilterType = 'include' | 'reset' | 'exclude' | 'toggle';
 
+export function addAdHocFilter(filter: AdHocVariableFilter, variableName: string, scene: SceneObject) {
+  const type: FilterType = filter.operator === '=' ? 'toggle' : 'exclude';
```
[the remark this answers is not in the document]
ivanahuckova: Why  in `const type: FilterType = filter.operator === '=' ? 'toggle' : 'exclude';` we are doing `toggle` instead of `include`, as the UI doesn't show that you could toggle it (no highlight). Seems like change of functionality, which I am not sure is intentional.

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -99,6 +104,43 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
     });
   }
 
+  private handleLabelFilter(key: string, value: string, frame: DataFrame | undefined, operator: FilterType) {
+    // NOOP. We need a way to let the user know why this is not possible.
```
ivanahuckova, line 108:
```suggestion
    // @todo: NOOP. We need a way to let the user know why this is not possible.
```

src/services/fields.ts
```diff
@@ -67,3 +67,26 @@ export function getFilterBreakdownValueScene(
     });
   };
 }
+
+export enum LabelType {
```
ivanahuckova, line 71:
```suggestion
// copied from public/app/plugins/datasource/loki/types.ts
export enum LabelType {
```

ivanahuckova (commented):
Lots of great refactors in here!! 🤩

matyax:
> Do we want to update the title tags?

That will be automagically picked up from Grafana.

imagen

ivanahuckova:
>> Do we want to update the title tags?

> That will be automagically picked up from Grafana.

Not sure if we want to update `refId` to something else than `A` to make it less confusing on what is query A. `Filter for value in query in explore logs` ? :D 
image

matyax:
Changelog:
- Simplified the implementation of the table filtering. Now the table doesn't need to know about filters, variables, etc. It can just call a function.
- Unified the check for LEVEL
- Made sure we only display active state of positive filters

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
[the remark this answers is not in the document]
matyax: Toggle is a "smarter" behavior when the filter already exists, so the change was intentional. But, since this function is only used by the Table and the table has no "active" status of filters, I better revert this change.

We intentionally use `toggle` for log details, which shows the active state.

src/Components/ServiceScene/LogsListScene.tsx
matyax: Good idea.

src/services/fields.ts
matyax: Good idea.

matyax:
> Not sure if we want to update refId to something else than A to make it less confusing on what is query A. Filter for value in query in explore logs ? :D Or Filter for value in query filters. :D

To me just `filter for value` `filter out value` is fine. In any case, we don't control that message from the app. Maybe it would be an interesting idea to have default messages but also expose a simple API for customizing the title of those options.

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -14,57 +14,81 @@ export interface AddToFiltersButtonState extends SceneObjectState {
...
     reportAppInteraction(
       USER_EVENTS_PAGES.service_details,
       USER_EVENTS_ACTIONS.service_details.add_to_filters_in_breakdown_clicked,
       {
         filterType: this.state.variableName,
-        key: selectedFilter.name,
+        key: filter.name,
         action: type,
-        filtersLength: variable.state.filters.length,
+        filtersLength: variable?.state.filters.length || 0,
```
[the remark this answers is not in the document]
ivanahuckova: @gtk-grafana trend setter

gtk-grafana (approved):
Looks good, works well locally, and you fixed a bug in the table view where you can no longer add conflicting filters:
image

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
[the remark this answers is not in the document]
gtk-grafana: 😎

ivanahuckova (approved):
👏

matyax:
Added unit tests. Merging. Thanks for the collab and feedback!

PR #467, opened 2024-06-18
fix(service-name-op): hide operator on `service_name`

svennergr:
It's not pretty, but it works...
Hiding the operator for `service_name`, as this one should always be an `equal` operator.

### DIFF
--- a/src/Components/IndexScene/LayoutScene.tsx
+++ b/src/Components/IndexScene/LayoutScene.tsx
@@ -117,6 +117,15 @@ function getStyles(theme: GrafanaTheme2) {
           },
         },
       },
+      // the `service_name` filter is a special case where we want to hide the operator
+      '[data-testid="AdHocFilter-service_name"]': {
+        'div[class*="input-wrapper"]:first-child': {
+          display: 'none',
+        },
+        'div[class*="input-wrapper"]:nth-child(2)': {
+          marginLeft: 0,
+        },
+      },
 
       ['div >[title="Add filter"]']: {
         border: 0,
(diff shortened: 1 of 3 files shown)


src/Components/IndexScene/LayoutScene.tsx
```diff
@@ -117,6 +117,15 @@ function getStyles(theme: GrafanaTheme2) {
           },
         },
       },
+      // the `service_name` filter is a special case where we want to hide the operator
+      '[data-testid="AdHocFilter-service_name"]': {
```
matyax, line 121: 😬

matyax:
I would get rid of this code which is not really doing anything: https://github.com/grafana/explore-logs/blob/main/src/Components/IndexScene/IndexScene.tsx#L164-L174


https://github.com/grafana/explore-logs/assets/1069378/f3e88fa0-aff9-4958-8bbb-c76ee48bb0a4

matyax (approved):

svennergr:
> I would get rid of this code which is not really doing anything: https://github.com/grafana/explore-logs/blob/main/src/Components/IndexScene/IndexScene.tsx#L164-L174

You mean get rid of the condition?

matyax:
Yeah, because only works when there's only 1 filter, otherwise it does nothing.

svennergr:
> Yeah, because only works when there's only 1 filter, otherwise it does nothing.

This: https://github.com/grafana/explore-logs/pull/467/commits/2bbd528c15ab6f3f56a4d76c556df8efb0635f81 ?

matyax (approved):
:cohete:

PR #473, opened 2024-06-19
Service selection scene: filter logs by the selected level in the chart

matyax:
This PR adds support for applying a level filter to the logs preview we show in the service selection.

Demo:

https://github.com/grafana/explore-logs/assets/1069378/34bb039a-ce56-457f-9ad1-27d452a5ac75

### DIFF
--- a/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
+++ b/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
@@ -303,10 +303,16 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
 
   // Creates a layout with logs panel
   buildServiceLogsLayout = (service: string) => {
-    const serviceLevel = this.state.serviceLevel.get(service);
-    const levelFilter = serviceLevel
-      ? ` | logfmt | json | drop __error__, __error_details__ | level=\`${serviceLevel}\` or detected_level=\`${serviceLevel}\` `
-      : '';
+    let serviceLevel = this.state.serviceLevel.get(service);
+    let levelFilter = '';
+    if (serviceLevel) {
+      let operator = 'or';
+      if (serviceLevel === 'logs') {
+        serviceLevel = '';
+        operator = 'and';
+      }
+      levelFilter = ` | logfmt | json | drop __error__, __error_details__ | level=\`${serviceLevel}\` ${operator} detected_level=\`${serviceLevel}\` `;
+    }
     return new SceneCSSGridItem({
       $behaviors: [new behaviors.CursorSync({ sync: DashboardCursorSync.Off })],
       body: PanelBuilders.logs()
--- a/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
+++ b/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
@@ -39,6 +39,7 @@ import { getQueryRunner, setLeverColorOverrides } from 'services/panel';
 import { ConfigureVolumeError } from './ConfigureVolumeError';
 import { NoVolumeError } from './NoVolumeError';
 import { PluginPage } from '@grafana/runtime';
+import { toggleLevelFromFilter } from 'services/filters';
 
 export const SERVICE_NAME = 'service_name';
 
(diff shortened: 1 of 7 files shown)


gtk-grafana:
Looks good but I'm seeing a bug when querying a non-standard level name: 
image
(DB: grafana logging dev, service: `grafana/grafana` or `cortex-gw`)

matyax:
Yeah. Not really a bug. I'm not sure if we should do nothing when "logs" level is queried or what else.

matyax:
Yeah. Not really a bug. I'm not sure if we should do nothing when "logs" level is queried or what else. ("logs" level means no known level).

gtk-grafana:
To the user it is a bug, I would expect to see logs that don't have a level. Looks like this is an issue in explore as well

matyax:
Agree. What I mean is that there's nothing broken, just "logs" isn't a level.

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -216,60 +221,110 @@ export class ServiceSelectionComponent extends SceneObjectBase<ServiceSelectionC
...
+      body: panel,
     });
   }
 
   // Creates a layout with logs panel
-  buildServiceLogsLayout(service: string) {
+  buildServiceLogsLayout = (service: string) => {
+    const serviceLevel = this.state.serviceLevel.get(service);
+    const levelFilter = serviceLevel
+      ? ` | logfmt | json | drop __error__, __error_details__ | level=\`${serviceLevel}\` or detected_level=\`${serviceLevel}\` `
```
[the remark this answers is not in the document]
gtk-grafana: This is a good question, I feel like this or could slow the query down a bit

matyax:
When the level is no level/missing level (logs), we can query for `level != "" and detected_level != ""`. We still need to determine if we want to use both or just detected.

> when selecting multiple, or excluding via holding down command

Did't know that existed. Will take a look, thanks!

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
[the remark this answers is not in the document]
svennergr: > Should we always just use detected_level?

Let's just use `detected_level`.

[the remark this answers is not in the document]
matyax: Using `detected_level` now. Thanks!

matyax:
It was a small challenge, but I managed to support the panel interactions in both modes (I was about to suggest not supporting append).

https://github.com/grafana/explore-logs/assets/1069378/465ac6f5-bc39-4f81-a757-f6c5358a7b62

I moved the logic to a service that I'd like to reuse for the Logs Volume, should we decide to do the same there.

src/services/levels.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const ALL_LEVELS = ['logs', 'debug', 'info', 'warn', 'error', 'crit'];
+
+export function toggleLevelFromFilter(
+  level: string,
+  serviceLevels: string[] | undefined,
+  mode: SeriesVisibilityChangeMode
+) {
+  if (mode === SeriesVisibilityChangeMode.ToggleSelection) {
+    const levels = serviceLevels || [];
```
gtk-grafana, line 11: ??

src/services/levels.ts
```diff
@@ -0,0 +1,27 @@
+import { SeriesVisibilityChangeMode } from '@grafana/ui';
+
+export const ALL_LEVELS = ['logs', 'debug', 'info', 'warn', 'error', 'crit'];
```
gtk-grafana, line 3: Does this work with numeric log lines as well?

gtk-grafana (approved):
Works great, and looks great. Thanks Matias!

src/services/levels.ts
matyax: That's a good question. I worked on the assumption that these are the levels we will get from `detected_level`, but it wouldn't hurt to double check.

svennergr: Well, there's no defined list. Every value of a `level` label can be the value of `detected_level`.

matyax: I was just writing. Based on [this code](https://github.com/grafana/loki/blob/dbc0e2489cb7d88f04155f4087e6e40a7dc92ff3/pkg/distributor/distributor.go#L402-L416), it would grab the level 1) from a label 2) from metadata 3) detect it (returning normalized names). 

I have an idea, but otherwise we can't make assumptions about possible levels, so we need to adjust.

matyax: Updated. We now read all the levels from the series, so it should support whatever Loki sends as `detected_level`.

PR #474, opened 2024-06-20
LogsVolumePanel: support filtering logs from the selected level

matyax:
Closes https://github.com/grafana/explore-logs/issues/422

This PR adds support to apply level filters from the Logs Volume panel.

https://github.com/user-attachments/assets/58d6978d-430b-4c32-9f41-3ceda8a3867e


matyax:
When the user is `appending` levels, we don't change the filters because we don't yet support filters with multiple values.

cyriltovena (approved):
LGTM

matyax:
Thanks @cyriltovena ! What does @grafana/observability-logs / @zizzpudding  this of this? How does this behavior feels to you?

gtk-grafana:
@matyax what if we just ignored level/detected_level in the logs volume, although then we'd want to be able to set legend visibility when a level/detected_level filter is applied, but it looks like we can do that with series visibility overrides? 🤔 

```
"matcher": {
          "id": "byNames",
          "options": {
            "mode": "exclude",
            "names": [
              "A-series1"
            ],
            "prefix": "All except:",
            "readOnly": true
          }
        },
```

cyriltovena:
I haven't follow the tech side of it but I love this feature.

matyax:
I've added a new variable, dedicated to hold the level, that is used everywhere but in the logs volume query. This should take this feature pretty much where we wanted.

src/services/panel.ts
```diff
@@ -30,6 +29,27 @@ export function setLeverColorOverrides(overrides: any) {
   });
 }
 
+interface TimeSeriesFieldConfig extends FieldConfig {
+  hideFrom: HideSeriesConfig;
+}
+export function setLevelSeriesOverrides(levels: string[], overrides: FieldConfigOverridesBuilder<FieldConfig>) {
```
matyax, line 35: This override is not setting the visibility of levels in the Logs Volume viz and I don't understand why. Need to further investigate.

gtk-grafana: create a panel in a dashboard where it works and compare the JSON data of your viz

matyax: Good idea. I applied the same structure and values that Grafana was internally setting, but maybe I missed something.

matyax:
```
"overrides": [
          {
            "matcher": {
              "id": "byName",
              "options": "{level=\"info\"}"
            },
            "properties": [
              {
                "id": "custom.hideFrom",
                "value": {
                  "legend": false,
                  "tooltip": false,
                  "viz": true
                }
              }
            ]
          }
        ]
```

https://obslogs.grafana-dev.net/d/cds8vpood924ga/new-dashboard?tab=queries&viewPanel=panel-1&from=2024-07-19T11:48:02.437Z&to=2024-07-19T17:48:02.437Z

matyax: That's for one. Hiding multiple:

```
"overrides": [
          {
            "__systemRef": "hideSeriesFrom",
            "matcher": {
              "id": "byNames",
              "options": {
                "mode": "exclude",
                "names": [
                  "{level=\"error\"}"
                ],
                "prefix": "All except:",
                "readOnly": true
              }
            },
            "properties": [
              {
                "id": "custom.hideFrom",
                "value": {
                  "legend": false,
                  "tooltip": false,
                  "viz": true
                }
              }
            ]
          }
        ]
```

matyax: Ohhhh, it was "byNames" not "byName" ‼️  🙇 @gtk-grafana

matyax:
Not having to refresh logs volume seems to produce a bit more of responsiveness in the app. And it's one less query that we need to run. 💡

matyax:
@grafana/observability-logs Mind taking another look? Thanks!

matyax:
It seems that we've been bitten by the buggy interpolation error we recently fixed, as the metric query it's producing for for fields and values is like: `sum(count_over_time({service_name=`tempo-distributor`}   | logfmt   {service_name=\"tempo-distributor\"}| drop __error__ | namespace!=\"\" [$__auto])) by (namespace)`

src/services/variables.ts
```diff
@@ -12,7 +14,14 @@ export const VAR_LOGS_FORMAT = 'logsFormat';
 export const VAR_LOGS_FORMAT_EXPR = '${logsFormat}';
 export const VAR_LINE_FILTER = 'lineFilter';
 export const VAR_LINE_FILTER_EXPR = '${lineFilter}';
-export const LOG_STREAM_SELECTOR_EXPR = `${VAR_LABELS_EXPR} ${VAR_PATTERNS_EXPR} ${VAR_LOGS_FORMAT_EXPR} ${VAR_FIELDS_EXPR} ${VAR_LINE_FILTER_EXPR}`;
+let STREAM_SELECTOR_EXPR = `${VAR_LABELS_EXPR} ${VAR_PATTERNS_EXPR} ${VAR_LOGS_FORMAT_EXPR} ${VAR_LEVELS_EXPR} ${VAR_FIELDS_EXPR} ${VAR_LINE_FILTER_EXPR}`;
```
matyax, line 17: We either hold this PR until a new Grafana image is available, or we do something like this. Any suggestion will be much appreciated.

To reproduce, use the app with `yarn server` (as the e2e tests do), and browse labels.

imagen

matyax:
We will wait for a new Grafana release before merging these changes.

matyax:
Never stop believing.

svennergr:
I think with that we should also upgrade the grafana versions in `plugin.json` and `docker-compose.yaml`, WDYT?

matyax:
Definitely.

PR #477, opened 2024-06-20
Pages: Switch to useSceneApp

gtk-grafana:
Currently not working as expected, the `useSceneApp` cache always misses, forcing every class to re-instantiate on anything that changes the url params.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "grafana-lokiexplore-app",
-  "version": "0.0.11",
+  "version": "0.0.12",
   "description": "Query less exploration of log data stored in Loki",
   "scripts": {
     "build": "webpack -c ./.config/webpack/webpack.config.ts --env production",
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "@emotion/css": "^11.10.6",
     "@grafana/data": "^11.0.0",
     "@grafana/runtime": "^11.0.0",
-    "@grafana/scenes": "4.32.0",
+    "@grafana/scenes": "5.2.0",
     "@grafana/ui": "^11.0.0",
     "@playwright/test": "^1.43.1",
     "@types/react-table": "^7.7.20",
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -161,16 +161,7 @@ function getVariableSet(initialDS?: string, initialFilters?: AdHocVariableFilter
     key: 'adhoc_service_filter',
   });
 
-  filterVariable._getOperators = () => {
-    // Only allow equal operator if there is only one filter
-    if (filterVariable.state.filters.length === 1) {
-      return [
-        {
-          label: FilterOp.Equal,
-          value: FilterOp.Equal,
-        },
-      ];
-    }
+  filterVariable._getOperators = function () {
     return operators;
   };
 
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -72,6 +72,7 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
       body: new LayoutScene({}),
       ...state,
     });
+    console.log('indexscene render');
 
     this.addActivationHandler(this.onActivate.bind(this));
   }
--- a/src/Components/ServiceScene/ServiceScene.tsx
+++ b/src/Components/ServiceScene/ServiceScene.tsx
@@ -1,11 +1,11 @@
 import { css } from '@emotion/css';
 import React from 'react';
 
-import { GrafanaTheme2, LoadingState } from '@grafana/data';
+import {GrafanaTheme2, LoadingState, PageLayoutType} from '@grafana/data';
 import { locationService } from '@grafana/runtime';
 import {
   AdHocFiltersVariable,
-  CustomVariable,
+  CustomVariable, SceneAppPage,
   SceneComponentProps,
   SceneFlexItem,
   SceneFlexLayout,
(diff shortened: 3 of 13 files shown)


gtk-grafana:
This is currently in a somewhat working state. Caveats are browser back button can now take you back to the index page, but filters are not cleared (but they don't exist in the URL):

image

And breadcrumbs are a bit buggy:
image

Hopefully we can fix these bugs and merge this in before refactoring the code to support new routes.

package.json
```diff
@@ -74,7 +74,7 @@
     "@emotion/css": "^11.10.6",
     "@grafana/data": "^11.0.0",
     "@grafana/runtime": "^11.0.0",
-    "@grafana/scenes": "4.32.0",
+    "@grafana/scenes": "5.2.0",
```
gtk-grafana, line 77: Need 5.2.0 to support the layout prop in the sceneAppPage

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -78,70 +79,91 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
...
       getUrlSyncManager().cleanUp(this);
     };
   }
 
+  /**
+   * @todo why do we need to manually sync fields, but nothing else?
+   * @param fieldsVariable
+   * @private
+   */
+  private syncFieldsWithUrl(fieldsVariable: AdHocFiltersVariable) {
```
gtk-grafana, line 127: I think this is causing the double back bug.

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -195,6 +217,7 @@ function getVariableSet(initialDS?: string, initialFilters?: AdHocVariableFilter
       dsVariable,
       filterVariable,
       fieldsVariable,
+      // @todo where is patterns being added to the url? Why do we have var-patterns and patterns?
```
gtk-grafana, line 220: address in another PR, this is already too big

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -47,16 +49,14 @@ export interface IndexSceneState extends SceneObjectState {
   // contentScene is the scene that is displayed in the main body of the index scene - it can be either the service selection or service scene
   contentScene?: SceneObject;
```
matyax, line 50: Now that we're refactoring this, I find this concept of `body` and `contentScene` confusing. Body seems to be fixed to always be a `LayoutScene`, so maybe we should remove it? Or at least create the instance in the constructor?

https://github.com/grafana/explore-logs/blob/3da4446d8b66abc58e187826f587d8e4130e5500/src/Components/IndexScene/IndexScene.tsx#L91

Or maybe the `body` should either be ServiceSelectionScene or ServiceScene. I think it would simplify this component a bit more.

src/Components/Pages.tsx
```diff
@@ -0,0 +1,128 @@
...
+        routePath: ROUTE_DEFINITIONS.fields,
+        getPage: (routeMatch, parent) => makeBreakdownPage(routeMatch, parent, SLUGS.fields),
+      },
+      {
+        routePath: '*',
+        getPage: () => makeRedirectPage(),
+      },
+    ],
+  });
+}
```
matyax, line 66: 👌

src/Components/Pages.tsx
```diff
@@ -0,0 +1,128 @@
...
+    preserveUrlKeys: DRILLDOWN_URL_KEYS,
+    getParentPage: () => parent,
+    getScene: () => getServicesScene(),
+  });
+}
+
+export function getSlug() {
+  const location = locationService.getLocation();
+  const slug = location.pathname.slice(location.pathname.lastIndexOf('/') + 1, location.pathname.length);
+  return slug as SLUGS;
```
matyax, line 114: I think we can do a bit better with this type. It's also unexpected to be all uppercase. Maybe `AppSection`, `AppViewType`, `LogsExplorationView` or something like this.

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -44,20 +40,18 @@ import { buildPatternsScene } from './Breakdowns/PatternsBreakdownScene';
...
   samples: Array<[number, string]>;
 }
 
-export type ActionViewType = 'logs' | 'labels' | 'patterns' | 'fields';
-
-interface ActionViewDefinition {
+interface BreakdownViewDefinition {
   displayName: string;
-  value: ActionViewType;
+  value: SLUGS;
```
matyax, line 54: Related with my previous comment; at first glance looks like some enum or constant, not a type.

src/Components/Pages.tsx
```diff
@@ -0,0 +1,128 @@
...
+}
+
+export function extractServiceFromRoute(routeMatch: SceneRouteMatch<{ service: string }>): { service: string } {
```
matyax, line 117: Maybe these can go to the routing service?

src/services/routing.ts
```diff
@@ -1,15 +1,137 @@
...
+  `var-${VAR_DATASOURCE}`,
+  `var-${VAR_FILTERS}`,
+  `var-${VAR_FIELDS}`,
+  `var-${VAR_FIELD_GROUP_BY}`,
+  `var-${VAR_LABEL_GROUP_BY}`,
+  `var-${VAR_DATASOURCE}`,
+  `var-${VAR_LOGS_FORMAT}`,
+  `var-${VAR_LINE_FILTER}`,
+];
+
```
matyax, line 78: A few unit tests for these wouldn't hurt.

matyax (approved):
Works perfectly for me, amazing job! Left mostly minor comments and suggestions.

src/Components/ServiceScene/ServiceScene.tsx
gtk-grafana: SLUGS is an enum

src/Components/Pages.tsx
gtk-grafana: Moved

src/Components/ServiceScene/ServiceScene.tsx
gtk-grafana: Renamed tho

src/Components/Pages.tsx
gtk-grafana: Updated

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: Yeah I'll take a look, the constructor isn't between routes if the scene is cached, but the activation function is, so as long as there's nothing in the layoutScene that needs to be re-inited we should be fine to move this to the constructor

gtk-grafana: Fixed.

PR #479, opened 2024-06-24
[POC] Value breakdowns: calculate and sort by standard deviation

matyax:
POC of sorting, descending, by standard deviation.

For a given series, we can access calculations using `doStandardCalcs()` and transformations like `stdDev`. With that, we could offer the user different sorting options.

Demo:

Uploading Demo.mov…

Related with https://github.com/grafana/explore-logs/issues/316


src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
```diff
@@ -46,8 +46,23 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
   private performRepeat(data: PanelData) {
     const newChildren: SceneFlexItem[] = [];
 
-    for (let seriesIndex = 0; seriesIndex < data.series.length; seriesIndex++) {
-      const layoutChild = this.state.getLayoutChild(data, data.series[seriesIndex], seriesIndex);
+    const reducer = fieldReducers.get(ReducerID.stdDev);
+
+    const fieldCalcs = data.series.map((dataFrame) => ({
+      calcs: doStandardCalcs(dataFrame.fields[1], true, true),
```
matyax, line 52: I'm just leaving `calcs` here to show that we can use those values for sorting too, like we do with stdDev, but it's not needed for this POC.

cyriltovena (approved):
LGTM

So cool, what other function could we use ?

matryer:
Ohh this is exciting!

matyax:
I was just looking at that. Some low hanging criteria could be:

- Highest values
- All zero values
- Mean value

matyax:
I'm now working on adding a Select with a few sorting options, and direction (asc, desc).

stevendungan:
This looks really great!!! A few questions I had @matyax:

1. From the video, I assume this is addressing sorting of the graphs, but the UI suggests it is sorting of the Data source - I assume at this point the user has determined their data source
image

Do we think it's best here to:
* remove the ability to select the data source on this page (and make corresponding text change?)
* move the sort selector somewhere nearby?

2. Do we think "UNSORTED" is a bit of a distraction/papercut? I think we can replace "UNSORTED" with "SORTED BY [x]" (whatever our opinionated way that default should be)

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
```diff
@@ -43,11 +43,25 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
     });
   }
 
+  private getSortedSeries(data: PanelData) {
```
gtk-grafana, line 46: Can we add a comment that this is working under the assumption that each dataframe has a single time and value field

gtk-grafana:
is this supposed to be open for review, looks like it's still a WIP?

matyax:
> is this supposed to be open for review, looks like it's still a WIP?

Today I removed the PoC part and moved on to implement the sorting criteria and direction. Will ping back when it's ready for review.

matyax:
This should be ready to take a second look, now with more options for sorting, and direction. Ran out of time to add coverage, but will do tomorrow.

gtk-grafana:
Notes from our call earlier:

Not enough space, shrink search bar?
Pasted Graphic

Sort has no impact on single view:
Pasted Graphic 1

Add calculated value to UI somewhere (e.g. the legend)

matyax:
Made it bigger and added a tooltip.

imagen

If you don't mind, I'd prefer to take care of the other items in a follow-up PR.

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
matyax: Each data frame from this `PanelData` is a series, I'm not sure I understand the comment. In any case, I can add this to the follow up changes.

matyax:
Merging these changes to continue in follow UP prs.

PR #490, opened 2024-06-26
Table: Persist user column resizing

gtk-grafana:
Currently when users resize columns, any refresh in dataframe will reset the column widths to the default calculation. This PR aims to persist any user column width overrides set in the UI via dragging the column header divider.

This does not save the column widths to the URL, so changes will only be local for now, but it should be relatively simple to add width overrides to the URL in a follow up PR. However we are saving column widths to local storage, so any changes users make will persist in their browser until they are reset via the new option that shows up in the column context menu when a custom override has been added:

image

Also the user can clear overrides for specific columns in the column management UI:
image


src/Components/Table/ColumnSelection/LogsTableNavField.tsx
```diff
@@ -73,6 +79,19 @@ export function LogsTableNavField(props: {
               </div>
             </div>
           )}
+          {props.columnWidthMap && props.setColumnWidthMap && props.columnWidthMap?.[props.label] !== undefined && (
+            <div
+              onClick={() => {
+                const { [props.label]: omit, ...map } = { ...props.columnWidthMap };
+                props?.setColumnWidthMap?.(map);
```
matyax, line 86:
```suggestion
                props.setColumnWidthMap(map);
```

matyax (approved):
Looks great!

src/Components/Table/ColumnSelection/LogsTableNavField.tsx
gtk-grafana: This throws a type error, we'd need to add an additional `typeof props.setColumnWidthMap === 'function'` conditional outside of the react template conditional, i.e. 
```
if(typeof props.setColumnWidthMap === 'function'){
  props.setColumnWidthMap(map);
}
```

gtk-grafana: I was able to remove the first optional chaining operator however, just not the for the function call

matyax: Yeah, it's super annoying. Thanks for checking!

PR #493, opened 2024-06-27
E2E: reduce test flake

gtk-grafana:
Increasing expect timeout, some loki queries were taking longer then 5s locally, I assume this might be responsible for some of the flake in CI as well.

Also adding an assertion in a test that the loading state is undefined before opening the logs panel, were some executions that would run before the dataframe came back.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -20,7 +20,7 @@ export default defineConfig<PluginOptions>({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
+  retries: process.env.CI ? 2 : 1,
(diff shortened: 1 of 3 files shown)


playwright.config.ts
```diff
@@ -20,7 +20,7 @@ export default defineConfig<PluginOptions>({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
+  retries: process.env.CI ? 2 : 1,
```
matyax, line 23: Do we need this change?

matyax (approved):

playwright.config.ts
gtk-grafana: Hmm yeah, it might hide flake, I'll remove it for now.

PR #495, opened 2024-07-01
UX: add search icon to field/label autocomplete dropdown

gtk-grafana:
Just a small helper in the UI to let users know they can search by label name.

Main:
image

This PR:
image

This PR:
image
image

We don't have a lot of flexibility with the search icon already in the base select component, I would like to always show the search icon so users know it's searchable and not just a dropdown. This PR suggests adding the search icon as a prefix that is only visible when the select isn't open.

Also making the dropdown a bit bigger, the ALL value is selected by default and it's easy to miss this element in this state.

Main:
image

This PR: 
image


matyax (approved):
Great idea!

PR #496, opened 2024-07-01
chore: add npm script for generating local logs

trevorwhitney:
I sometimes like to have everything running locally (loki, grafana, explore plugin) without docker, and in those cases it's nice to still leverage the generator to ship directly to my local Loki instance. This PR adds a npm script for doing just that, and gitignores for the generator vendoring.


gtk-grafana (approved):

svennergr (approved):
Nice!

PR #500, opened 2024-07-02
add CODEOWNERS

svennergr:


matyax (approved):
👏

JStickler:
Would it make sense for me to be added as CODEOWNER for `/docs/sources` so that I get notified when there are docs to review?

svennergr:
> Would it make sense for me to be added as CODEOWNER for `/docs/sources` so that I get notified when there are docs to review?

Absolutely, https://github.com/grafana/explore-logs/pull/502

PR #512, opened 2024-07-03
Value breakdown: refactor sorting options and add changepoint detection sorting

matyax:
This PR introduces some minor improvements to the sorting options, but the main change is the integration of https://github.com/grafana/augurs .

Changes:
- Added changepoint detection as a sorting option (currently the default option, renamed as "Relevance").
- From the previous list of reducers/sorting options, moved stdDev to the second place and renamed it to "Dispersion".

Demo:

https://github.com/grafana/explore-logs/assets/1069378/50fe77f6-1a5a-4b27-887e-9591907a18be

### Notes

Currently a work in progress/POC.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
+++ b/src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { LoadingState, PanelData, DataFrame, FieldType, fieldReducers, doStandardCalcs } from '@grafana/data';
+import { LoadingState, PanelData, DataFrame } from '@grafana/data';
 import {
   SceneObjectState,
   SceneFlexItem,
@@ -10,7 +10,7 @@ import {
   SceneByFrameRepeater,
   SceneLayout,
 } from '@grafana/scenes';
-import { ChangepointDetector } from '@bsull/augurs';
+import { sortSeries } from 'services/sorting';
 
 interface ByFrameRepeaterState extends SceneObjectState {
   body: SceneLayout;
@@ -64,57 +64,9 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
     }
   };
 
-  private getSortedSeries = (data: PanelData) => {
-    const reducer = (dataFrame: DataFrame) => {
-      if (this.sortBy === 'changepoint') {
-        return this.calculateChangepointsValue(dataFrame);
-      }
-      const fieldReducer = fieldReducers.get(this.sortBy);
-      const value =
-        fieldReducer.reduce?.(dataFrame.fields[1], true, true) ?? doStandardCalcs(dataFrame.fields[1], true, true);
-      return value[this.sortBy] ?? 0;
-    };
-
-    const seriesCalcs = data.series.map((dataFrame) => ({
-      value: reducer(dataFrame),
-      dataFrame: dataFrame,
-    }));
-
-    seriesCalcs.sort((a, b) => {
-      if (a.value && b.value) {
-        return b.value - a.value;
-      }
-      return 0;
-    });
-
-    if (this.direction === 'asc') {
-      seriesCalcs.reverse();
-    }
-
-    return seriesCalcs.map(({ dataFrame }) => dataFrame);
-  };
-
-  private calculateChangepointsValue = (data: DataFrame) => {
-    const fields = data.fields.filter((f) => f.type === FieldType.number);
-
-    const dataPoints = fields[0].values.length;
-    let samplingStep = Math.floor(dataPoints / 100) || 1;
-    if (samplingStep > 1) {
-      // Avoiding "big" steps for more accuracy
-      samplingStep = Math.ceil(samplingStep / 2);
-    }
-
-    const sample = fields[0].values.filter((_, i) => i % samplingStep === 0);
-
-    const values = new Float64Array(sample);
-    const points = ChangepointDetector.defaultArgpcp().detectChangepoints(values);
-
-    return points.indices.length;
-  };
-
   private performRepeat(data: PanelData) {
     const newChildren: SceneFlexItem[] = [];
-    const sortedSeries = this.getSortedSeries(data);
+    const sortedSeries = sortSeries(data.series, this.sortBy, this.direction);
 
     for (let seriesIndex = 0; seriesIndex < sortedSeries.length; seriesIndex++) {
       const layoutChild = this.state.getLayoutChild(data, sortedSeries[seriesIndex], seriesIndex);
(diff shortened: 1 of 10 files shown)


src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
```diff
@@ -20,6 +21,7 @@ type FrameFilterCallback = (frame: DataFrame) => boolean;
 type FrameIterateCallback = (frames: DataFrame[], seriesIndex: number) => void;
 
 export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
+  public changepointDetector = ChangepointDetector.defaultArgpcp();
```
sd2k, line 24:
```suggestion
```

I'm now realising that the docs are really unclear about this, but this should be recreated afresh for each series, as the detector continues to build up a model of how a series looks every time `detectChangepoints` is called, and uses the existing state.

When I make this change locally I get decent looking results here - sample size of 1, of course :grin:

image

Another plus: it should be faster, since the time/memory complexity is > O(N) in the number of points the detector sees.

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
```diff
@@ -63,28 +65,42 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
...
 
-    return fieldCalcs.map(({ field }) => field);
-  }
+    return seriesCalcs.map(({ dataFrame }) => dataFrame);
+  };
+
+  private changePointsValue = (data: DataFrame) => {
+    const fields = data.fields.filter((f) => f.type === FieldType.number);
+    const values = new Float64Array(fields[0].values);
+    const points = this.changepointDetector.detectChangepoints(values);
```
sd2k, line 101:
```suggestion
    const points = ChangepointDetector.defaultArgpcp().detectChangepoints(values);
```

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
```diff
@@ -63,28 +65,42 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
...
     }
 
-    return fieldCalcs.map(({ field }) => field);
-  }
+    return seriesCalcs.map(({ dataFrame }) => dataFrame);
+  };
+
+  private changePointsValue = (data: DataFrame) => {
+    const fields = data.fields.filter((f) => f.type === FieldType.number);
+    const values = new Float64Array(fields[0].values);
```
sd2k, line 100: If it's still too slow, we could consider sampling these points before feeding them to the detector (something like this?)

```suggestion
    const values = new Float64Array(fields[0].values.filter((_, i) => i % 5 === 0));
```

Will need to experiment to see how it runs though.

sd2k (commented):
A few suggestions inline but it's working for me :tada: 

Another suggestion for a future PR: when sorting by 'Auto' we could consider separating series out into those _with_ changepoints and those _without_. (If we wanted to get super fancy, we could use an outlier detector on the changepoint counts to find any series which have an outlying number of changepoints :exploding_head:). That way, we highlight 'interesting' series more prominently - being at the front of the sort doesn't necessarily make them stand out, e.g. in the screenshot in the inline comment.

sd2k:
Another idea to avoid the UI freezing could be to run the changepoint detection in a web worker, but my webpack-foo isn't very strong... I struggled to get it working in another plugin :disappointed:

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
matyax: Ahhhh, I had my suspicions but assumed that instances could be reused. Thanks!

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
matyax: Sampling definitely makes a huge difference in performance, but also in results. When results are sampled, the order changes, sometimes significantly.

Tab on the left, sampling every 3, and on the right with no sampling (claerly the slowest tab), with data of the last 6 hours:

https://github.com/grafana/explore-logs/assets/1069378/3f66895e-4ab5-4c4d-915b-0e5558d2ace7

I'll see what happens using different sample sizes based on the data set length.

matryer:
@clord could probably help with the workers stuff, he did some of it recently.

matryer:
Loving this direction. Helping people find interesting stuff in the sea of graphs is a fantastic capabilities for these apps.

Some thoughts:
1. I wonder if `Auto` should be `Recently changed`? Is that correct (enough)?
2. Can we do away with Asc/Desc dropdown altogether?
3. What if we take away any sort option that we don't have a strong use case for? For now. And wait for someone to ask for something additional?

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
matyax: I updated this to define the "step" to take samples based on the time series length with a simple rule:
- If the series length is below 100, the step is 1
- Else, step will be taking in multiples of 100

For example:

L = series length
S = step

```
L   | S | Final sample size
1   | 1 | 1
361 | 3 | 121
127 | 1 | 127
324 | 3 | 108
43  | 1 | 3
229 | 2 | 115
3   | 1 | 3
```

With that, I'm getting more consistent sorting results between sampled and non sampled.

https://github.com/grafana/explore-logs/assets/1069378/80647282-70f4-4808-9d92-751e387ca553

sd2k: Nice, that sounds very sensible to me! :tada:

matryer:
1. What if `Auto` was `Most interesting`?

sd2k:
I tried to get web workers working in the past with mixed success - [this branch](https://github.com/grafana/scenes-ml/compare/main...web-workers) has some findings, most difficulties were to do with webpack config & how it interacts with plugin builds & public paths...

matryer:
Consider tweaking the help text to: `Best-effort identification of the most relevant graphs`.

sd2k (approved):
This is working well for me :rocket: not sure what's going on with the tests though?

matyax:
Thank you @sd2k ! I'm wrapping up this PR, so I need to update and fix the tests, it's an expected failure.

matyax:
@sd2k FYI, this is needed for Jest when one of the modules includes @bsull/augurs https://github.com/grafana/explore-logs/pull/512/commits/61804f0c5399432d9e848278249b806f6c981d89

matyax:
- Renamed from "Auto" to "Relevance"
- Renamed from "Standard deviation" to "Dispersion"
- Extracted and unit tested (and discovered a small bug in the process) the sorting code

I'm going to merge this so we can play with it a bit, let me know if I forgot something, and we can continue this in follow-up PRs. Thanks everyone!

PR #514, opened 2024-07-03
feat(show-number-of-logs): show the total number of logs

svennergr:


matryer:
@svennergr how far did you get with this?

PR #515, opened 2024-07-03
Services: Add autocomplete search

gtk-grafana:
Updates the UX of the services search. Results are now autocompleted, and new services are fetched as the user types into the dropdown.
Selecting an option in the dropdown will automatically take you to the service view, and typing a partial match and exiting the dropdown will add a custom option with a filter icon, showing that the user is filtering services matching the search text.

Fixes: https://github.com/grafana/explore-logs/issues/375

image
image
image

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/FieldSelector.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldSelector.tsx
@@ -2,16 +2,20 @@ import { css } from '@emotion/css';
 import React, { useState } from 'react';
 
 import { GrafanaTheme2, SelectableValue } from '@grafana/data';
-import { Select, useStyles2, InlineField, Icon } from '@grafana/ui';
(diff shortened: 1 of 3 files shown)


tests/exploreServices.spec.ts
```diff
@@ -13,7 +13,7 @@ test.describe('explore services page', () => {
 
   test('should filter service labels on search', async ({ page }) => {
     await explorePage.servicesSearch.click();
-    await explorePage.servicesSearch.fill('mimir');
+    await explorePage.servicesSearch.pressSequentially('mimir');
```
matyax, line 16: TIL

matyax (approved):
Much better!

svennergr (changes requested):
Can we wait for a small bit.

svennergr (approved):

PR #516, opened 2024-07-04
feat(initial-datasource): always select `grafanacloud-logs` as the default datasource

svennergr:
With this PR the default datasource will be the datasource with the `grafanacloud-logs` uid. This `uid` is set for the stack's default logs datasource.

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -51,7 +51,6 @@ export interface IndexSceneState extends SceneObjectState {
   controls: SceneObject[];
   body?: LayoutScene;
   initialFilters?: AdHocVariableFilter[];
-  initialDS?: string;
   patterns?: AppliedPattern[];
 }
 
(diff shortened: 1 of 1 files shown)


cyriltovena (approved):
LGTM

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -62,7 +61,8 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     super({
       $timeRange: state.$timeRange ?? new SceneTimeRange({}),
       $variables:
-        state.$variables ?? getVariableSet(state.initialDS ?? getLastUsedDataSourceFromStorage(), state.initialFilters),
+        state.$variables ??
+        getVariableSet('grafanacloud-logs' ?? getLastUsedDataSourceFromStorage(), state.initialFilters),
```
matyax, line 65: What would happen if, for whatever reason, there's no `grafanacloud-logs` in the instance? I think the first use could be one example, where you don't have either `grafanacloud-logs` and last used from storage.

svennergr: It just selects the first from the list.

image
-> 
image

matyax: Fantastic!

PR #518, opened 2024-07-04
feat(detected-fields): get fields from `/detected_fields` API

### LINKED ISSUE #245: [detected fields tab] implement detected_fields API

svennergr:
This PR uses Loki's `/detected_fields` API to get fields from there. The PR also contains a lot of renaming, thus there are quite a lot of small changes.

For the user, nothing really should change. 

Note: This does only work with a recent Grafana version that contains https://github.com/grafana/grafana/pull/89849

Fixes #245


src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -316,6 +327,10 @@ const emptyStateStyles = {
   }),
 };
 
+function getFieldByLabel(fields: Array<SelectableValue<DetectedField>>, fieldLabel: string | undefined) {
```
matyax, line 330: We should really not define functions with required arguments that can be undefined. Rather, we should not call this function if `fieldLabel` is undefined.

Random, but, for example, this would return a field with undefined label, if both things happen at the same time. In any case, even without running conditions, an undefined required parameter is not a good funciton signature.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -340,28 +355,23 @@ function getStyles(theme: GrafanaTheme2) {
   };
 }
 
-const avgFields = ['duration', 'count', 'total', 'bytes'];
+const avgFieldTypes = ['duration'];
```
matyax, line 358: I've no context, why did you remove the other fields? Do we need the array?

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -191,33 +183,32 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
       this.setState({
-        fields: [],
         loading: false,
+        fields: response.fields
+          .filter((field) => !disabledFields.includes(field.label) && field.cardinality > 1)
+          .sort((a, b) => a.label.localeCompare(b.label))
+          .map((field) => new DetectedField(field.type, field.label, field.parsers, field.cardinality)),
       });
+    } catch (error) {
+      console.error('Could not fetch detected_fields', error);
```
matyax, line 211: Probably set loading to false here.

src/services/variables.ts
```diff
@@ -17,3 +17,19 @@ export const EXPLORATION_DS = { uid: VAR_DATASOURCE_EXPR };
 export const ALL_VARIABLE_VALUE = '$__all';
 export const LEVEL_VARIABLE_VALUE = 'detected_level';
 export const PATTERNS_TEXT_FILTER = 'patternsFilter';
+
+export class DetectedField {
```
matyax, line 21: This should definitely not be here.

matyax (commented):
Changes look great! Left a few comments, but also locally I'm seeing some errors:

imagen

svennergr:
> Changes look great! Left a few comments, but also locally I'm seeing some errors:
> 
> imagen

probably based on the last main merge. thank you, i'll take a look!

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
svennergr: True! Fun fact, and this should not be an excuse, I used the "Refactor -> Extract as function" functionality for this.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
svennergr: Good shout. On thing that changed is that we now make a decision based on the "type" the API returns, not based on the field name anymore. But I was in fact missing "bytes" here.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -23,35 +24,51 @@ import { Alert, Button, DrawStyle, LoadingPlaceholder, StackingMode, useStyles2
...
 import { ByFrameRepeater } from './ByFrameRepeater';
 import { FieldSelector } from './FieldSelector';
 import { LayoutSwitcher } from './LayoutSwitcher';
-import { StatusWrapper } from './StatusWrapper';
-import { BreakdownSearchScene, getLabelValue } from './BreakdownSearchScene';
 import { SortByScene, SortCriteriaChanged } from './SortByScene';
-import { getSortByPreference } from 'services/store';
+import { StatusWrapper } from './StatusWrapper';
+
+export class DetectedField {
```
matyax, line 44: Apologies, I was pretty vague with my previous comment. Ideally, my suggestion would be to just move this to its own little module file. Can be under services, or feel free to introduce a `models` or `data` folder.

matyax:
I noticed that, during the initial logs load after selecting a service, the logs query fails.


https://github.com/grafana/explore-logs/assets/1069378/192dc998-4d68-4cf9-8915-a28ba2a7cc9c

The error is:

imagen

Only happens in this branch.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
svennergr: Makes total sense!

matyax (commented):
I noticed that, during the initial logs load after selecting a service, the logs query fails.


https://github.com/grafana/explore-logs/assets/1069378/192dc998-4d68-4cf9-8915-a28ba2a7cc9c

The error is:

imagen

Only happens in this branch.

matyax:
We can "ignore" ☝️ , as it also happens in main.

PR #526, opened 2024-07-08
fix(lazy-load): add missing `isLazy` prop

svennergr:
Fixes the missing lazy loading in fields and label breakdowns by adding the missing `isLazy` property.


src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -261,11 +261,13 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
       active: 'grid',
       layouts: [
         new SceneCSSGridLayout({
+          isLazy: true,
```
matyax, line 264: !!!

PR #539, opened 2024-07-09
Logs panel: implement wrap lines option

matyax:
"Wrap lines" has been implemented as the first visualization option for logs, and it's only visible when this viz type is used.

Additional changes:
- A few renames to respect conventions (class names, module names, storage keys).


gtk-grafana (approved):
Looks good and works well locally, nice touch adding unit tests!

PR #541, opened 2024-07-09
Sorting: filter some sorting options

matyax:
Rationale for percentiles:

> 10th Percentile: Indicates the value below which 10% of the data falls, useful for understanding the lower extremes of the data distribution.
> 
> 25th Percentile (First Quartile, Q1): TMrks the value below which 25% of the data falls. It helps to understand the lower range of the data distribution.
> 
> 50th Percentile (Median): Represents the middle value of a dataset, providing a measure of central tendency that is less affected by outliers than the mean.
>
> 75th Percentile (Third Quartile, Q3): Marks the value below which 75% of the data falls, providing insight into the upper range of the data distribution.
> 
> 90th Percentile: Indicates the value below which 90% of the data falls, providing insight into the upper extremes of the data distribution.
> 
> 1st Percentile and 99th Percentile: Extreme percentiles are often used to identify outliers and understand the extreme values in the distribution.

Removed every "all *" options (all zeros, all nulls, all values, all unique values)

Options:

**Most relevant:** Smart ordering of graphs based on the data
**Dispersion:** Standard deviation of all values in a field
**Last:** Last value
**First:** First Value
**Min:** Minimum Value
**Max:** Maximum Value
**Mean:** Average Value
**Variance:** Variance of all values in a field
**Total:** The sum of all values
**Count:** Number of values in response
**Range:** Difference between minimum and maximum values
**Delta:** Cumulative change in value
**Step:** Minimum interval between values
**Difference:** Difference between first and last values
**Min (above zero):** Used for log min scale
**Change Count:** Number of times the value changes
**Distinct Count:** Number of distinct values
**Difference percent:** Percentage difference between first and last values
**1st %:** 1st percentile value
**10th %:** 10th percentile value
**25th %:** 25th percentile value
**50th %:** 50th percentile value
**75th %:** 75th percentile value
**90th %:** 90th percentile value
**99th %:** 99th percentile value


matyax:
Shall we go with these options? @grafana/observability-logs @matryer @cyriltovena ?

gtk-grafana:
Maybe we can remove the null/NaN calculations as well as I don't think the count of logs for an interval is ever null or NaN?

matyax:
Good idea. Done.

svennergr (approved):

PR #545, opened 2024-07-10
fix(ds-selection): change precedence of selected datasources

svennergr:
This PR fixes the behavior that would always default to the `grafanacloud-logs` datasource and would never actually use the datasource saved in localstorage.

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -62,7 +62,7 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
       $timeRange: state.$timeRange ?? new SceneTimeRange({}),
       $variables:
         state.$variables ??
-        getVariableSet('grafanacloud-logs' ?? getLastUsedDataSourceFromStorage(), state.initialFilters),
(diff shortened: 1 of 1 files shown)


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -62,7 +62,7 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
       $timeRange: state.$timeRange ?? new SceneTimeRange({}),
       $variables:
         state.$variables ??
-        getVariableSet('grafanacloud-logs' ?? getLastUsedDataSourceFromStorage(), state.initialFilters),
+        getVariableSet(getLastUsedDataSourceFromStorage() ?? 'grafanacloud-logs', state.initialFilters),
```
matyax, line 65: I saw this in the PR from a few days ago but it didn't click.

matyax (approved):

src/Components/IndexScene/IndexScene.tsx
svennergr: I made that PR and it didn't click.

PR #548, opened 2024-07-10
Break down scenes: sort all layouts on sort change

matyax:
Before these changes, we were only sorting the first layout, grid, but not the rows view.


matryer (approved):
Well spotted.

PR #555, opened 2024-07-10
Services: State inconsistency across routes

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/554

This PR adds a singleton metadata service that stores the current service scene state to the singleton before navigation, which is then pulled after navigation has occurred in the activation handler.

This should fix the issue where drilldown views need a refresh to show the user the proper state as they are adding filters and navigating back to a drilldown.


gtk-grafana:
Added issue to add some test coverage for this case: https://github.com/grafana/explore-logs/issues/556

src/services/metadata.ts
```diff
@@ -0,0 +1,34 @@
+import { ServiceSceneCustomState, ServiceSceneState } from '../Components/ServiceScene/ServiceScene';
+
+let metadataService: MetadataService;
+
+export function initializeMetadataService(): void {
+  if (!metadataService) {
+    metadataService = new MetadataService();
+  }
+}
```
matyax, line 9: This should be enough and then you don't need any initialization, right?

```suggestion
const metadataService = new MetadataService();
```

matyax (approved):
Tested the most usual happy paths and workflows and seemed fine, even better than current main.

src/services/metadata.ts
gtk-grafana: Yeah this does seem to work, but in app-o11y and grafana core we always init singletons like this? 
Scares me a bit...

gtk-grafana: I guess it comes down to how the bundles are built? Using a scoped variable within the module works as long as the file is included once across all bundles? I'll ask someone smarter then myself

matyax: It's definitely not a big deal.

gtk-grafana: Great question though, but let's merge and I'll follow up

svennergr (approved):
LGTM

PR #562, opened 2024-07-10
Table: Use detected_level, fix low contrast background color & text

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/519

image
image


gtk-grafana:
Biggest complaint of mine is that the level column needs a bunch more space to render because of the longer title.

src/Components/Table/Constants.ts
```diff
@@ -0,0 +1 @@
+export const LEVEL_NAME = 'detected_level';
```
matyax, line 1: Minor, but this file should be lower case.

matyax (approved):
Looks great!

PR #568, opened 2024-07-10
Table: clear selected line from url state after the table is initialized

gtk-grafana:
Fix for table when viewing a link to a specific log line.
The problem was that we weren't clearing the url state after rendering the initial row highlight in the table, so after updating the time range, the selected row would still scroll into view.

Requires core change: https://github.com/grafana/grafana/pull/90308 to fix bug where passing in undefined 
 to rowHighlightIndex wouldn't update the state of the table if it was previously set in another render.


gtk-grafana:
Will need to wait until after public preview release next week since it requires core changes

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -92,6 +92,12 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
     }
   }
 
+  clearSelectedLine() {
+    this.setState({
+      selectedLine: undefined,
+    });
+  }
+
```
matyax, line 100: Is this being called from anywhere?

matyax (approved):
Works very well!

matyax:
I really like having these options always visible 👌 

imagen

src/Components/ServiceScene/LogsListScene.tsx
gtk-grafana: yeah from the LogsTableScene https://github.com/grafana/explore-logs/pull/568/files#diff-30db4d73ccca81096b38ec2e68d2c0ac8969ddf5c062c0ef01ba2dfbcebb1c0bR40-R45

PR #572, opened 2024-07-11
Detected fields: No fields with active filters message

gtk-grafana:
When no results are available in the detected fields, show a new message that prompts users to clear out the filters.

image

Also updated the view when no filters are applied:
image

Also fixed the Grot error overflowing on small viewports/mobile devices while I was in there.

Notes:
This shouldn't redirect you to the logs view after clicking on it, but it does because every change to a non-pattern filter redirects you to the logs view. I think we should change this globally, but that is out of scope for this issue (which would be nice to get out before public preview)

Doesn't solve 529, because that is for labels not fields, although labels can only be put in this state via patterns, when detected field filters remove all data we just get panels of no-data.
image

Fixes https://github.com/grafana/explore-logs/issues/536 tho maybe?


src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -199,13 +247,28 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
...
       ],
     });
   }
+  private buildClearFiltersLayout(clearCallback: () => void) {
+    return new SceneReactObject({
+      reactNode: (
+        <GrotError>
+          <Alert title="" severity="info">
+            No labels match these filters.{' '}
+            <a className={emptyStateStyles.link} onClick={() => clearCallback()}>
```
matyax, line 263: Anchor with no link is considered no bueno. Mind using a span that looks like a link or something similar?

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -179,14 +200,41 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
     this.setState(stateUpdate);
   }
 
+  private clearVariables(variablesToClear: Array<SceneVariable<SceneVariableState>>) {
+    return () => {
```
matyax, line 204: Minor, but I'm not a huge fan of a class method returning a callback, and it's more like a variables-clearing factory.

matyax (commented):
Works great! Let me know what you think of the comments.

This was supposed to be an "info", right?

imagen

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
gtk-grafana: Yeah or just a button. good catch

gtk-grafana: image
How's this?

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
gtk-grafana: Fixed!

gtk-grafana:
@matyax no clue why you didn't get the info alert, maybe your branch was stale?

matyax:
Not sure, but looks okay now! 
imagen

matyax (approved):

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -168,8 +172,25 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
...
+        if (variable instanceof AdHocFiltersVariable && variable.state.filters.length) {
+          variablesToClear.push(variable);
+        }
+        if (variable instanceof CustomVariable && variable.state.value && variable.state.name !== 'logsFormat') {
+          variablesToClear.push(variable);
+        }
+      }
+
+      if (variablesToClear.length > 1) {
+        stateUpdate.body = this.buildClearFiltersLayout(() => this.clearVariables(variablesToClear));
```
matyax, line 190: <3

PR #579, opened 2024-07-11
chore: replace react-beautiful-dnd with successor

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/573

Needs to wait until after 11.2 release of grafana/ui.
And then we need to update the package.lock in this PR before merging


svennergr (approved):
LGTM

PR #580, opened 2024-07-11
Sorting: update options and add alphabetical order

matyax:
Closes https://github.com/grafana/explore-logs/issues/574

- Made select input wider
- Added a "target" for the sorting event (so we know if we sorted fields or labels)
- Updated sorting options and descriptions

Available options:

**Most relevant**: Smart ordering of graphs based on the most significant spikes in the data
**Widest spread**: Sort graphs by deviation from the average value
**Name**: Alphabetical order
**Highest spike**: Show graphs with the highest values first (max)
**Lowest dip**: Show graphs with the smallest values first (min)
**10th %**: 10th percentile value
**25th %**: 25th percentile value
**75th %**: 75th percentile value
**90th %**: 90th percentile value
**99th %**: 99th percentile value

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
@@ -26,12 +26,12 @@ export class SortByScene extends SceneObjectBase<SortBySceneState> {
         {
           value: 'changepoint',
           label: 'Most relevant',
-          description: 'Smart ordering of graphs based on the data',
+          description: 'Smart ordering of graphs based on the most significant spikes in the data',
         },
         {
           value: ReducerID.stdDev,
           label: 'Widest spread',
-          description: 'Prioritize graphs that have changed the most',
+          description: 'Prioritize graphs that have changed the most (standard deviation)',
         },
         {
           value: 'alphabetical',
--- a/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
@@ -41,12 +41,12 @@ export class SortByScene extends SceneObjectBase<SortBySceneState> {
         {
           value: ReducerID.max,
           label: 'Highest spike',
-          description: 'Show graphs with the highest values first (max)',
+          description: 'Sort graphs by the highest values (max)',
         },
         {
           value: ReducerID.min,
           label: 'Lowest dip',
-          description: 'Show graphs with the smallest values first (min)',
+          description: 'Sort graphs by the smallest values (min)',
         },
       ],
     },
--- a/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
@@ -31,7 +31,7 @@ export class SortByScene extends SceneObjectBase<SortBySceneState> {
         {
           value: ReducerID.stdDev,
           label: 'Widest spread',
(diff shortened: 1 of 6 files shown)


src/Components/ServiceScene/Breakdowns/SortByScene.tsx
```diff
@@ -21,16 +21,39 @@ export class SortCriteriaChanged extends BusEventBase {
...
     {
-      value: 'changepoint',
-      label: 'Most relevant',
-      description: 'Smart ordering of graphs based on the data',
+      label: '',
+      options: [
+        {
+          value: 'changepoint',
+          label: 'Most relevant',
+          description: 'Smart ordering of graphs based on the data',
```
gtk-grafana, line 29: How about `Smart ordering of graphs based on the most significant spikes in the data'?

src/Components/ServiceScene/Breakdowns/SortByScene.tsx
```diff
@@ -21,16 +21,39 @@ export class SortCriteriaChanged extends BusEventBase {
...
+        },
+        {
+          value: ReducerID.stdDev,
+          label: 'Widest spread',
+          description: 'Prioritize graphs that have changed the most',
```
gtk-grafana, line 34: Can we at least let users know that this is stdDev? i.e. 'Prioritize graphs that have changed the most (standard deviation)'

gtk-grafana: Plus this description isn't accurate, standard deviation is a measure of variation from the mean, a single outlier with large magnitude of difference from the mean could give a particular graph the largest stdDev, but otherwise have all values with the same value, "changing the most" is subjective, and if I wanted to dig into why a particular graph is being sorted first with this description there's nothing I can google to figure it out

src/Components/ServiceScene/Breakdowns/SortByScene.tsx
matyax: Sounds good to me 👍

src/Components/ServiceScene/Breakdowns/SortByScene.tsx
matyax: Definitely.

matyax: "Deviated the most"?

gtk-grafana: "Sort graphs by deviation" would be better, also since there is a sort order I'm not sure we want to say "the most" because when sorting DESC you're seeing "the least"

matyax: Yeah, great point there. I removed the "first" from other descriptions, because first and last will be defined by the ascending or descending order. Thanks!

src/Components/ServiceScene/Breakdowns/SortByScene.tsx
```diff
@@ -21,16 +21,39 @@ export class SortCriteriaChanged extends BusEventBase {
 export class SortByScene extends SceneObjectBase<SortBySceneState> {
   public sortingOptions = [
     {
-      value: 'changepoint',
-      label: 'Most relevant',
-      description: 'Smart ordering of graphs based on the data',
+      label: '',
+      options: [
```
matyax, line 25: @matryer What do you think?

gtk-grafana (approved):

PR #583, opened 2024-07-12
feat(pattern-tokens): add better visibility to pattern tokens

svennergr:
Especially when there are a lot of patterns and a lot of tokens, the dotted underline is not enough to find those tokens. This PR adds a different background colour to those tokens, and emphasises that on hover.

image

On hover:
image

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternNameLabel.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternNameLabel.tsx
@@ -112,7 +112,6 @@ function getStyles(theme: GrafanaTheme2) {
 
       '&:hover': {
         textDecoration: 'underline',
-        // somehow this needs 3px when hovered
         textUnderlineOffset: '3px',
         backgroundColor: theme.colors.emphasize(theme.colors.background.primary, 0.2),
       },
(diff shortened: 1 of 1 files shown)


matyax:
Have you considered without underline? 

imagen

Just asking, not sure what I would prefer.

In parallel, it seems that Emotion doesn't like the kebab classes.

imagen

svennergr:
> Have you considered without underline?

In my opinion, the dotted underline emphasises a bit more that hose things are clickable. Well spotted, I changed to camel casing.

svennergr:
@zizzpudding we need you here 😅 Can you chime in if we should remove the dotted underline or keep it?

zizzpudding:
Yes, definitely. 

Here is a link to the [Figma design](https://www.figma.com/design/NT8kODiyLshr8S5WoXplIA/ExploreLogs?node-id=1288-15977&t=nCAzsdTWgfpdfK4E-1) for tokenizing the patterns better.

svennergr:
> Yes, definitely.
> 
> Here is a link to the [Figma design](https://www.figma.com/design/NT8kODiyLshr8S5WoXplIA/ExploreLogs?node-id=1288-15977&t=nCAzsdTWgfpdfK4E-1) for tokenizing the patterns better.
> 
> image
> 
> This is all based on work that was done in Adaptive Logs to make the patterns more readable. We can reach out to Travis Peterson and Ben Clive (who is doing the work on this).

I think we can defintely improve that in the future. Let's focus on just the iteration of this PR where we add some better styles to show users that the tokens are clickable.
For that specifically: should we keep the dotted underline or remove it?

gtk-grafana (approved):
Either way with the underline or not, I'm good with this change

matryer:
on the call we said,

1. leave `<_>` text for now
2. use background and darker on hover
3. remove underline from both non-hover and hover states

PR #584, opened 2024-07-12
Sorting: memoize sorting function

matyax:
For some criteria, sorting can be a very expensive client-side operation, because it will reduce a final value from all the time series values (or a sample of them), which was also being calculated twice for each "Grid" or "Rows" view.

These changes memoize the sorting function and implement a custom key based on the provided time series.

Before:

https://github.com/user-attachments/assets/7c0a662a-f280-4917-8a42-4a03d2c4d57f

After:

https://github.com/user-attachments/assets/5d8656e1-609a-499c-aede-9634dcc9566c


src/services/sorting.ts
```diff
@@ -1,35 +1,50 @@
...
+      seriesCalcs.reverse();
     }
-    return 0;
-  });
 
-  if (direction === 'asc') {
-    seriesCalcs.reverse();
+    return seriesCalcs.map(({ dataFrame }) => dataFrame);
+  },
+  (series: DataFrame[], sortBy: string, direction: string) => {
```
matyax, line 36: Might be a bit too much, but I want to be sure the sorting function is actually being called when a different time series is passed.

matyax:
This is a positive optimization, but we decided to hold until later to not introduce a last-minute surprise.

matyax:
@grafana/observability-logs mind looking at this one? It would be positive to optimize sorting a little bit.

svennergr (approved):
LGTM

PR #589, opened 2024-07-12
Sorting: check for wasm support or use stdDev

matyax:
Wasm may or may not be supported by the browser. When that happens, fall back to stdDev.

imagen


gtk-grafana (approved):
good catch!

PR #606, opened 2024-07-15
Routes: Adding routes for label values

gtk-grafana:
New routes for value breakdowns.
`/explore/service/:serviceName/label/:labelName`
`/explore/service/:serviceName/field/:fieldName`

Fixes: https://github.com/grafana/explore-logs/issues/587

Notes for reviewers:
* This is a big refactor, and I'm sorry
* We removed the duplicate detected_labels query in the LabelBreakdownScene, and removed duplicate labels/fields in the breakdown scene states
* * Only exception being the variable value, which needs to be stored outside of the variable
* Added a `CustomConstantVariable` extension of the base variable class, it doesn't do anything besides overload the `getValueOptions` function to just return the internal options state
* * This is required because the CustomVariable is not designed to have its options updated externally, how it was working before I'm not sure.
* Breadcrumbs are kinda funny, it's just using the label name for now and capitalizing the first letter, we might want to move forward on finding a way to pass the PageNav through the `SceneAppPage` to have more customization here.


src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -36,74 +34,103 @@ import { getLabelOptions, sortLabelsByCardinality } from 'services/filters';
...
 export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneState> {
   protected _variableDependency = new VariableDependencyConfig(this, {
     variableNames: [VAR_LABELS],
-    onReferencedVariableValueChanged: this.onReferencedVariableValueChanged.bind(this),
   });
 
-  constructor(state: Partial<LabelBreakdownSceneState>) {
+  // Labels/options can be passed in when instantiated, but should ONLY exist on the state of the variable
+  constructor(state: Partial<LabelBreakdownSceneState> & { options?: VariableValueOption[]; value?: string }) {
+    const variable = new CustomConstantVariable({
```
gtk-grafana, line 59: This can be moved inline

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -36,74 +34,103 @@ import { getLabelOptions, sortLabelsByCardinality } from 'services/filters';
...
-      if (
-        newState.options !== oldState.options ||
-        newState.value !== oldState.value ||
-        newState.loading !== oldState.loading
-      ) {
-        this.updateBody(variable);
-      }
-    });
+    // Pull the detected_labels from our service scene, update the variable when they change
+    this._subs.add(
```
gtk-grafana, line 95: refactor this like the fields breakdown scene please

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -369,8 +378,13 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
                 <search.Component model={search} />
               </>
             )}
-            {!loading && fields.length > 1 && (
-              <FieldSelector label="Field" options={fields} value={value} onChange={model.onFieldSelectorChange} />
+            {!loading && options.length > 1 && (
+              <FieldSelector
+                label="Field"
+                options={options}
+                value={String(value)}
```
matyax, line 385:
```suggestion
                value={String(value ?? '')}
```

Maybe this, to prevent `"undefined"`.

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -36,74 +34,113 @@ import { getLabelOptions, sortLabelsByCardinality } from 'services/filters';
...
+  };
+
+  /**
+   * Pull the detected_labels from our service scene, update the variable when they change
+   * @param newState
+   * @param prevState
+   */
+  private onServiceStateChange = (newState: ServiceSceneState, prevState: ServiceSceneState) => {
+    const variable = this.getVariable();
+    if (JSON.stringify(newState.labels) !== JSON.stringify(prevState.labels)) {
```
matyax, line 126: It would be so nice and elegant if we were able to trust these arrays and simply compare references, without having to stringify...

src/services/navigate.ts
```diff
@@ -0,0 +1,122 @@
...
+      locationService.push(fullUrl);
+      return;
+    }
+  }
+}
+
+/**
+ * Navigates to the drilldown view specified by the path slug
+ * Note: If the serviceScene is not provided we assume it is not a parent of the calling class, i.e. we're navigating from the service selection view, instead of a drilldown view
+ * Drilldown views should ALWAYS provide the serviceScene state
```
matyax, line 75: The only way to ensure this would be to come to this function and read this comment. Maybe we should split this function into two? One for navigating from the parent scene and another to be used from the service selection page?

Looking at the codebase, the only call that I found was `navigateToBreakdown(ROUTES.logs(service));` and as far as I can see it's not actually "navigating to a breakdown", it's navigating to the logs view.

That would also simplify this function a bit.

Similar to `navigateToIndex()`, maybe `navigateToService()`?

src/services/navigate.ts
```diff
@@ -0,0 +1,122 @@
...
+  DRILLDOWN_URL_KEYS,
+  PageSlugs,
+  prefixRoute,
+  replaceSlash,
+  ROUTES,
+  ValueSlugs,
+} from './routing';
+import { sceneGraph } from '@grafana/scenes';
+import { UrlQueryMap, urlUtil } from '@grafana/data';
+
```
matyax, line 17: I think it would be positive to cover these functions with a few test cases to ensure `locationService` is called with the expected routes. It should also surface any hidden bug.

matyax (commented):
Locally tested it and works as expected . Left a few comments and suggestions.

src/services/navigate.ts
matyax: Sounds good. In general, for "Drilldown views should ALWAYS provide the serviceScene state" I would have a function where the scene is not optional, and another one when the scene should not be passed, so we don't have to branch execution inside the function and the function-callers knows what to use and what to pass.

gtk-grafana: Yeah good call, I was still waking up when I first read this. I created a much more verbose `navigateToInitialPageAfterServiceSelection` function for this special case where you're navigating from the top-level to the service scene for the first time after selecting a service. Also removed the "breakdown" terminology except in regards to the label value views, and use drilldown when referring to generic sub-pages of the top page.

src/services/navigate.ts
gtk-grafana: Yeah good call, I also added some test cases to ensure the state in the metadata service is populated as well

src/services/navigate.ts
matyax: Perfect! Thanks.

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
gtk-grafana: Yeah a helper function is the only thing I can think of that might help, but that won't help if people don't remember that ```['a'] === ['a'] === false``` and call it in the first place.

gtk-grafana: Actually if we had a way to throw a linting error when comparing arrays/objects in the state subscription callback? Although that could get messy too

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
gtk-grafana: `VariableValue` cannot be undefined, at least in the types definition:

```
type VariableValue = VariableValueSingle | VariableValueSingle[];
type VariableValueSingle = string | boolean | number | CustomVariableValue;
```

I updated the types in the `CustomConstantVariable.ts` to only allow `isMulti` as false, and hinted that the value type is `VariableValueSingle`, so we have to address this when folks want to enable multi without pushing runtime errors.

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
matyax: Yeah, dw. I was mainly thinking out loud. I don't think this would be possible in this version of the app.

gtk-grafana: I created an issue to deal with this more generically: https://github.com/grafana/explore-logs/issues/614.

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.test.tsx
```diff
@@ -131,7 +131,7 @@ describe('addToFilters and addAdHocFilter', () => {
       const lookupVariable = jest.spyOn(sceneGraph, 'lookupVariable').mockReturnValue(adHocVariable);
       addToFilters('indexed', 'value', 'include', {} as SceneObject);
 
-      expect(lookupVariable).toHaveBeenCalledWith(VAR_LABELS, expect.anything());
+      expect(lookupVariable).toHaveBeenCalledWith(VAR_FIELDS, expect.anything());
```
svennergr, line 134: Isn't this change rather an indication that something is wrong? 
The test says
> identifies indexed lables and uses the appropriate variable

and with the change we seem to add it to the `FIELDS` variable and not `LABELS` anymore?

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -369,8 +378,13 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
                 <search.Component model={search} />
               </>
             )}
-            {!loading && fields.length > 1 && (
-              <FieldSelector label="Field" options={fields} value={value} onChange={model.onFieldSelectorChange} />
+            {!loading && options.length > 1 && (
```
svennergr, line 381: Btw, do you know why we do `options.length > 1` not `options.length > 0`?

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.test.tsx
svennergr: The app does create the correct query expression tho. Confusing

src/services/routing.ts
```diff
@@ -22,9 +20,22 @@ export enum PageSlugs {
   explore = 'explore',
   logs = 'logs',
   labels = 'labels',
+
```
svennergr, line 23:
```suggestion
```

src/services/routing.ts
```diff
@@ -39,6 +50,13 @@ export const ROUTES = {
   labels: (service: string) => prefixRoute(`${PageSlugs.explore}/service/${replaceSlash(service)}/${PageSlugs.labels}`),
 };
 
+export const SUB_ROUTES = {
+  label: (service: string, label: string) =>
+    prefixRoute(`${PageSlugs.explore}/service/${replaceSlash(service)}/${ValueSlugs.label}/${label}`),
```
svennergr, line 55: I guess we should call `replaceSlash` on `label` too?

```suggestion
    prefixRoute(`${PageSlugs.explore}/service/${replaceSlash(service)}/${ValueSlugs.label}/${replaceSlash(label)}`),
```

src/services/routing.ts
```diff
@@ -39,6 +50,13 @@ export const ROUTES = {
...
+  field: (service: string, label: string) =>
+    prefixRoute(`${PageSlugs.explore}/service/${replaceSlash(service)}/${ValueSlugs.field}/${label}`),
```
svennergr, line 57:
```suggestion
    prefixRoute(`${PageSlugs.explore}/service/${replaceSlash(service)}/${ValueSlugs.field}/${replaceSlash(label)}`),
```

svennergr (commented):
Very cool! Great stuff. Left a few comments, but looks good overall.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
gtk-grafana: Because the ALL value is always there

src/services/routing.ts
svennergr: Ah, it's label name. Yea, sure! Good call!

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.test.tsx
gtk-grafana: Good catch, looks like we were stubbing the object wrong in the test. Dunno how that was causing it to update the wrong variable, but after updating the stub to be a DetectedLabel[], the original assertions started working.

gtk-grafana:
Ok, this has been merged already, but github was having a moment when it was. Manually closing.

PR #609, opened 2024-07-16
Value breakdowns: maintain filters between value changes

matyax:
Summary of changes:
- Moved the filtering code from `BreakdownSearchScene` to `ByFrameRepeater`.
- Added the previously used filter to `ByFrameRepeater`
- `ByFrameRepeater` if has a filter, applies it after "repeating".
- Added a missing UI state for when there are no values matching your search:
  message

Closes https://github.com/grafana/explore-logs/issues/559

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/BreakdownSearchScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/BreakdownSearchScene.tsx
@@ -14,7 +14,6 @@ export class BreakdownSearchScene extends SceneObjectBase<BreakdownSearchSceneSt
     super({
       filter: '',
     });
-    console.log('new search');
   }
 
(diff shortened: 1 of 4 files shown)


gtk-grafana:
Wondering if we want to add an action to clear the search in the alert like we have when filters have removed all results?
image

matyax:
Oh, definitely. Good idea.

gtk-grafana (approved):
Works well, good addition, great work!

matyax:
I managed to update this PR with the route changes, to support keeping filters between value changes. Next, I would like to double check the combination of sorting + filtering + value changes + layout switcher (single, grid, row) to be sure we're not introducing (or carrying over) any inconsistency.

matyax:
Double checked for inconsistencies and looks good to go.

PR #611, opened 2024-07-16
Fields: Allow include/exclude of logs that don't have particular field

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/590

image

After excluding "Logs missing oldVersion"
image

When dealing with high-cardinality fields without descriptive labels (indexed fields), sometimes you want to include all fields that have any value, instead of trying to include each value.

Likewise, sometimes you want to exclude all fields that have a particular field. This feature saves the user from needing to include/exclude every graph to obtain this functionality.

I also tweaked the styles a bit and added a description so it stands out a bit


gtk-grafana:
Needs test coverage and a bit of discussion.

matyax:
In general I like this idea a lot, but I'm not really sure it's a good fit with the current breakdown UX. Even with the tooltip and the title change, I think it takes a lot of cognitive effort to understand the feature.

My initial reaction would be to maybe suggest turning this into a third button, that we show along the current ones, like: `[include] [exclude] [exclude all ¿?]`

imagen

matyax:
We probably can't implement ☝️ , but I feel that we should try a couple of alternatives based on this.

gtk-grafana:
After discussion with the squad, we're gonna remove the panel from the breakdown, but instead show new buttons on the aggregated field names view for include/exclude and rename "select" to "values". We'll rename "select" to "values" in the labels view as well.

For now we're going to limit the filtering of empty to fields (until we hear that users want this feature for labels as well).

CC: @zizzpudding

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -103,7 +104,12 @@ function resolveVariableNameForField(field: string, scene: SceneObject) {
 
 export class AddToFiltersButton extends SceneObjectBase<AddToFiltersButtonState> {
   public onClick = (type: FilterType) => {
-    const filter = getFilter(this.state.frame);
+    let filter = getFilter(this.state.frame);
+
+    if (this.state.labelName && !filter) {
+      filter = { name: this.state.labelName, value: '""' };
+    }
+
```
matyax, line 112: So we get this behavior when there's a label name but no frame name. If we could specify from the "outside" if the filter should include/exclude all, it would be ideal.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -283,7 +283,7 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
         continue;
       }
 
-      const query = buildDataQuery(getExpr(optionValue), {
+      const query = buildDataQuery(getExpr(optionValue, false), {
```
matyax, line 286: I'd suggest to create a new function like `getDrillDownExpr` instead of adding a boolean, as it helps a lot when reading the code.

matyax (commented):
The feature and the use case it aims to resolve is of great value. Left some small notes, let me know when it's ready for review again. Great work!

gtk-grafana:
Closing and starting over with refactor branch as base: https://github.com/grafana/explore-logs/pull/703

PR #618, opened 2024-07-17
Patterns: Show UI message when time range is too old to show patterns

gtk-grafana:
Users have expressed frustration when patterns disappear when looking for non-recent logs.
Let's add a UI state and link to the docs that specify how long patterns are around.


gtk-grafana:
image

The copy is drafty, and the link (https://grafana.com/docs/grafana/latest/explore/simplified-exploration/logs/patterns/) doesn't actually mention anything about how long patterns stick around for.

@JStickler @matryer @cyriltovena Can I get a little help?

matyax (approved):

PR #620, opened 2024-07-18
Fields: fix loading forever when fields return empty

gtk-grafana:
Introduced by refactoring due to label value routes.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
@@ -104,18 +104,14 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
   };
 
   private serviceFieldsChanged = (newState: ServiceSceneState, oldState: ServiceSceneState) => {
-    if (JSON.stringify(newState.fields) !== JSON.stringify(oldState.fields)) {
+    if (JSON.stringify(newState.fields) !== JSON.stringify(oldState.fields) || newState.loading !== oldState.loading) {
       this.updateFields(newState);
     }
   };
 
   private updateFields(state: ServiceSceneState) {
-    if (!state.fields || !state.fields.length) {
-      return;
-    }
-
     const variable = this.getVariable();
-    const options = getLabelOptions(state.fields);
+    const options = state.fields ? getLabelOptions(state.fields) : [];
 
     variable.setState({
       options,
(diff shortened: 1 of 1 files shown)


src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -174,6 +170,8 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
       loading: logsScene.state.loading,
     };
 
+    console.log('update body, loading?', logsScene.state.loading);
```
gtk-grafana, line 173: remove

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -104,18 +104,14 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
   };
 
   private serviceFieldsChanged = (newState: ServiceSceneState, oldState: ServiceSceneState) => {
-    if (JSON.stringify(newState.fields) !== JSON.stringify(oldState.fields)) {
+    if (JSON.stringify(newState.fields) !== JSON.stringify(oldState.fields) || newState.loading !== oldState.loading) {
```
matyax, line 107: Ohhhh, right. More reasons to move away from this type of checks.

matyax (approved):

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
gtk-grafana: Yeah, it's like a useEffect dependency, but there's no eslint rule to let you know when something is referenced but not in the dep array

matyax: What would happen if the fields are the same, but the loading state changed? I think the "or" will completely override this check.

gtk-grafana: should be fine, it'll just update the loading state in that case

gtk-grafana: But yeah, we'll call `getLabelOptions` uncessarily, splitting this out into two functions, for updating loading state, and updating options might be better

PR #622, opened 2024-07-18
Labels: Clear filters in value breakdown when in no-data state

gtk-grafana:
Also refactoring things a bit to share the clear-filters scene, and move the label values breakdown to a dedicated scene as well.

Also when in the labels view, you only want to clear fields, patterns, and line filters, as any label filters that remove data would be sent in the detected_labels call, and return no-data view before users can reach the breakdown.

Would be nice to do something when all panels are no-data in the all labels view, but should probably deal with that in another PR before this gets too big

WIP


gtk-grafana:
Closing for now, still something that needs to be done, but after our other in-flight refactors.

PR #625, opened 2024-07-23
feat: serve service histograms from aggregated metrics

trevorwhitney:
This depends on https://github.com/grafana/loki/pull/13621. Once that is merged, this change allows us to query those aggregated metrics to serve the service landing page histograms.

### DIFF
--- a/src/Components/ServiceScene/LogsVolumePanel.tsx
+++ b/src/Components/ServiceScene/LogsVolumePanel.tsx
@@ -113,37 +113,6 @@ export class LogsVolumePanel extends SceneObjectBase<LogsVolumePanelState> {
     return panel;
   }
 
-  // private hasSingleServiceSelector(): boolean {
-  //   const fields = sceneGraph.lookupVariable(VAR_FIELDS, this) as AdHocFiltersVariable;
-  //   const patterns = sceneGraph.lookupVariable(VAR_PATTERNS, this) as CustomVariable;
-  //   const lineFilter = sceneGraph.lookupVariable(VAR_LINE_FILTER, this) as CustomVariable;
-
-  //   if (fields.state.filters.length !== 0 || patterns.state.value !== '') {
-  //     return false;
-  //   }
-
-  //   const labels = sceneGraph.lookupVariable(VAR_LABELS, this) as AdHocFiltersVariable;
-  //   if (labels.state.filters.length > 1) {
-  //     return false;
-  //   }
-
-  //   const filter = (lineFilter.state.value as string).trim();
-  //   if (labels.state.filters[0].key === SERVICE_NAME) {
-  //     if (filter === '|~ `(?i)`' || !filter) {
-  //       return true;
-  //     }
-  //   }
-
-  //   return false;
-  // }
-
-  // private service(): string {
-  //   const labels = sceneGraph.lookupVariable(VAR_LABELS, this) as AdHocFiltersVariable;
-  //   const filters = labels?.state.filters ?? [];
-
-  //   return filters.find((filter) => filter.key === SERVICE_NAME)?.value ?? '';
-  // }
-
   private extendTimeSeriesLegendBus = (context: PanelContext) => {
     const originalOnToggleSeriesVisibility = context.onToggleSeriesVisibility;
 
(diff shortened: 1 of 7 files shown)


matyax (commented):
👏 👏 

How would we handle backwards compatibility?

trevorwhitney:
@matyax I see 2 options

1. put this behind a feature flag, turn the backend on a few weeks before enabling the frontend
2. make a second query to the traditional query path if the first one comes back empty (would be more complicated, how to figure out which times are empty?)

src/Components/ServiceScene/LogsVolumePanel.tsx
```diff
@@ -23,21 +41,59 @@ export class LogsVolumePanel extends SceneObjectBase<LogsVolumePanelState> {
...
+  public updateVolumePanel = () => {
+    this.setState({
+      panel: this.getVizPanel(),
+    });
+  };
+
   private getVizPanel() {
+    const service = this.service();
+    const queryString = this.hasSingleServiceSelector()
+      ? `sum by (${LEVEL_VARIABLE_VALUE}) (sum_over_time({__aggregated_metric__=\`${service}\`} | logfmt | unwrap count [$__auto]))`
```
trevorwhitney, line 89: the `_aggregated_metric__` query is still getting fields injects as line filter statements. I'm surprised that would happen since I'm not using the `$LOG_STREAM_SELECTOR_EXPR`. Where's that coming from, and is there a way to disable it in this case?

svennergr: We use `service_name` as an "AddHocVariable": https://github.com/grafana/explore-logs/blob/main/src/Components/IndexScene/IndexScene.tsx#L180-L189

The way scenes works, it will try to add all those variables as add hoc filters to every query run. We would need to change the variable type in order for this to work. As a workaround for your PR you could change
https://github.com/grafana/explore-logs/blob/main/src/Components/IndexScene/IndexScene.tsx#L182 to something like `datasource: { type: 'fake' },`.
With that you wouldn't see those `service_name` labels included, however it would break showing the values in the label dropdowns.

project-words.txt
```diff
@@ -428,6 +428,7 @@ udpa
 ufuzzy
 ugorji
 unsubs
+unsub
```
matyax, line 431: What is this?

src/Components/ServiceScene/LogsVolumePanel.tsx
```diff
@@ -23,21 +41,59 @@ export class LogsVolumePanel extends SceneObjectBase<LogsVolumePanelState> {
...
     }
+
+    const fields = sceneGraph.lookupVariable(VAR_FIELDS, this) as AdHocFiltersVariable;
+    const patterns = sceneGraph.lookupVariable(VAR_PATTERNS, this) as CustomVariable;
+    const lineFilter = sceneGraph.lookupVariable(VAR_LINE_FILTER, this) as CustomVariable;
+    const labels = sceneGraph.lookupVariable(VAR_LABELS, this) as AdHocFiltersVariable;
+
+    this._subs.add(
+      fields.subscribeToState((newState, prevState) => {
+        if (newState.filters.length !== prevState.filters.length) {
```
matyax, line 52: Filters may have the same length but different contents. What it's usually done, which is "not great", is comparing serialized strings with `JSON.parse()`.

svennergr:
> > make a second query to the traditional query path if the first one comes back empty (would be more complicated, how to figure out which times are empty?)
> 
> I think we need to do something like this, because otherwise the app will not work for older Loki versions.

When we move to `/detected_fields` we would require a different Loki version anyways. I think it's fine to not be backwards compatible in the state we are currently, as long as we support the latest released Loki version. But that's totally up for discussion.

svennergr (changes requested):
fyi, we should release a new version without this change, so just marking this accordingly so it doesn't get merged by accident.

matyax:
Agree.

project-words.txt
trevorwhitney: no idea, but CI was failing without it

matyax: Ahhhh, I see now. Thanks!

src/Components/ServiceScene/LogsVolumePanel.tsx
gtk-grafana: Helper function in https://github.com/grafana/explore-logs/pull/633 is being proposed as solution for this

matyax:
@trevorwhitney Do you mind if we clean up the frontend part of this PR or would you prefer that we add comments?

trevorwhitney:
Feel free to clean it up! Thank you

PR #627, opened 2024-07-23
Labels: Improve query performance

gtk-grafana:
Removes parsers and parser filters from the queries in the labels breakdowns, to improve performance of those queries.

Note:
This PR introduced custom variable overrides when instantiating the SceneQueryRunner. This allows for adding additional stream selectors that aren't currently in the UI state, in this case to exclude labels that have empty values.

Fixes: https://github.com/grafana/explore-logs/issues/626

This might conflict with https://github.com/grafana/explore-logs/pull/611 a bit, as that PR wants to introduce the empty values back in, but this PR should be the priority to help keep loki load down.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
@@ -11,6 +11,7 @@ import {
   SceneFlexItemLike,
   SceneFlexLayout,
   sceneGraph,
+  SceneObject,
   SceneObjectBase,
   SceneObjectState,
   SceneReactObject,
@@ -25,7 +26,7 @@ import { getQueryRunner, setLeverColorOverrides } from 'services/panel';
 import { buildLokiQuery } from 'services/query';
 import { ValueSlugs } from 'services/routing';
 import { getLokiDatasource } from 'services/scenes';
-import { ALL_VARIABLE_VALUE, LOG_STREAM_SELECTOR_EXPR, VAR_LABEL_GROUP_BY, VAR_LABELS } from 'services/variables';
+import { ALL_VARIABLE_VALUE, VAR_LABEL_GROUP_BY, VAR_LABELS, VAR_LABELS_EXPR } from 'services/variables';
 import { ByFrameRepeater } from './ByFrameRepeater';
 import { FieldSelector } from './FieldSelector';
 import { LayoutSwitcher } from './LayoutSwitcher';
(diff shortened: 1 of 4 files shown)


src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -341,11 +349,44 @@ function buildLabelsLayout(options: VariableValueOption[], scene: LabelBreakdown
 }
 
 function getExpr(tagKey: string) {
-  return `sum(count_over_time(${LOG_STREAM_SELECTOR_EXPR} | drop __error__ | ${tagKey}!="" [$__auto])) by (${tagKey})`;
+  // Is there a better way to check for all detected metadata?
+  if (tagKey === LEVEL_NAME) {
+    return `sum(count_over_time(${VAR_LABELS_EXPR} | ${LEVEL_NAME}!="" [$__auto])) by (${tagKey})`;
+  }
+  return `sum(count_over_time(${VAR_LABELS_EXPR} [$__auto])) by (${tagKey})`;
```
svennergr, line 356: We would lose filtering by lines, fields and patterns with that. Not sure that is foreseen?

svennergr:
Opened an alternative approach in https://github.com/grafana/explore-logs/pull/628

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -341,11 +349,44 @@ function buildLabelsLayout(options: VariableValueOption[], scene: LabelBreakdown
...
 
 const GRID_TEMPLATE_COLUMNS = 'repeat(auto-fit, minmax(400px, 1fr))';
 
+/**
+ * AdHoc variables are bound to the UI state, so if we want to add a specific override for a query without updating the UI, we'll need to clone the original and make updates.
+ * Allows adding custom filters that are specific to a single query, when passed into the SceneQueryRunner $variables prop.
+ * @param scene
+ * @param keyName
+ */
+function cloneVariablesToExcludeEmptyForKey(scene: SceneObject, keyName: string) {
```
matyax, line 367: I'm having a hard time understanding "to exclude empty for key".

gtk-grafana:
Closing in favor of approach in #628, that won't break existing functionality. 
I'm a little surprised the e2e tests passed, we might need to add some coverage for breakdown views and the various variables.

PR #628, opened 2024-07-24
Label queries: remove unneccessary filters and parsers in query expression

svennergr:
This change looks larger than it is. I moved the `buildLabelsLayout` and `buildLabelValuesLayout` into the class, so I can access a sceneObject from `getExpr`.

In `getExpr` you will find the important change: https://github.com/grafana/explore-logs/compare/svennergr/better-label-queries?expand=1#diff-0c7c5b52164d9762335a7b37b5b4822ed449d0a60ea4dd765c8ee29efad26ae1R367-R390

We now create the stream selector expression in the app, to add filtering for only existing items. To prevent the parser from being added, we check for `fields.length` and only add the parser if we have a fields filter expression.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
@@ -25,7 +25,7 @@ import { DetectedLabel, getFilterBreakdownValueScene } from 'services/fields';
 import { getQueryRunner, setLeverColorOverrides } from 'services/panel';
 import { buildLokiQuery } from 'services/query';
 import { ValueSlugs } from 'services/routing';
-import { getLokiDatasource } from 'services/scenes';
+import { getLokiDatasource, isDefined } from 'services/scenes';
 import {
   ALL_VARIABLE_VALUE,
   VAR_LABEL_GROUP_BY,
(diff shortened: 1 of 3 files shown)


src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -201,12 +212,190 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
+    let fieldExpressionToAdd = '';
+    // `LEVEL_VARIABLE_VALUE` is a special case where we don't want to add this to the stream selector
+    if (tagKey !== LEVEL_VARIABLE_VALUE) {
+      labelExpressionToAdd = { key: tagKey, operator: '!=', value: '' };
+    } else {
+      fieldExpressionToAdd = `| ${LEVEL_VARIABLE_VALUE} != ""`;
+    }
+    const streamSelectors = [...labelsVariable.state.filters, labelExpressionToAdd]
+      .filter((f) => !!f)
+      .map((f) => `${f!.key}${f!.operator}\`${f!.value}\``)
```
gtk-grafana, line 388: If these were undefined, would we rather throw an error at runtime, or execute a query that would likely throw an error?

```suggestion
      .map((f) => `${f?.key}${f?.operator}\`${f?.value}\``)
```

gtk-grafana (approved):
LGTM, one minor nit/question, but otherwise it works well for me locally, seems to interpolate labels with special chars properly with logfmt and json.

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
matyax: Oh, I missed this. I wouldn't mind a linting rule to avoid non-null assertions.

svennergr: Well, I do `.filter((f) => !!f)` before. I don't see how `f` would still be `null | undefined` here. Let me fix it with a type predicate, so we can spare both `!` and `?`.

svennergr: WDYT https://github.com/grafana/explore-logs/pull/628/files#diff-9ef2f732762145f9751be0e3c3cc4ce83b3ad2f7fbff9563aeea16cb67d3033fR62-R64

PR #629, opened 2024-07-24
Service views: Prevent extra queries

gtk-grafana:
There are two sets of extra queries that are being run that this PR attempts to fix.
1. When selecting a service, after updating the services variable, for every SceneQueryRunner loaded in the grid layout, we currently re-query, but add in the selected service, leading to a bunch of queries that return nothing, and look like ```{service_name=`mimir-distributor`, service_name=`just-selected-service`}```. Currently you will get 2 of these queries re-run for every single service that you have scrolled to, sometimes leading to dozens or hundreds queries being blasted out when landing on the ServiceScene

This PR simply clears the body of the service selection scene before navigation, which clears out all of the SceneQueryRunners which would have otherwise re-queried on change of the ad-hoc filter variable.

2. On loading the ServiceScene after selecting a service, we run the logs-volume query, and then check the log lines to see if we want to use json or logfmt parser, and then re-run the query with the parser. 

This PR checks the dataframe we have in the services scene, and updates the parser variable before clearing out the body state for the fix in 1 (since we don't use a parser in the service selection scene, it shouldn't trigger a re-query), IFF we have already queried the logs for the given service. If the user uses the dropdown, or the log panel query errored, or failed, then we'll fall-back to running the queries twice on load of the service scene.

### DIFF
--- a/src/Components/ServiceScene/ServiceScene.tsx
+++ b/src/Components/ServiceScene/ServiceScene.tsx
@@ -4,7 +4,6 @@ import React from 'react';
 import { GrafanaTheme2, LoadingState } from '@grafana/data';
 import {
   AdHocFiltersVariable,
-  CustomVariable,
   SceneComponentProps,
   SceneFlexItem,
   SceneFlexLayout,
@@ -17,7 +16,7 @@ import {
 } from '@grafana/scenes';
 import { Box, Stack, Tab, TabsBar, useStyles2 } from '@grafana/ui';
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from 'services/analytics';
-import { DetectedLabel, DetectedLabelsResponse, extractParserAndFieldsFromDataFrame } from 'services/fields';
+import { DetectedLabel, DetectedLabelsResponse, updateParserFromDataFrame } from 'services/fields';
 import { getQueryRunner } from 'services/panel';
 import { buildLokiQuery, renderLogQLStreamSelector } from 'services/query';
 import { getDrilldownSlug, getDrilldownValueSlug, PageSlugs, PLUGIN_ID, ValueSlugs } from 'services/routing';
@@ -29,7 +28,6 @@ import {
   VAR_DATASOURCE,
   VAR_FIELDS,
   VAR_LABELS,
-  VAR_LOGS_FORMAT,
   VAR_PATTERNS,
 } from 'services/variables';
 import {
@@ -195,16 +193,7 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
       });
   }
 
-  private getLogsFormatVariable() {
-    const variable = sceneGraph.lookupVariable(VAR_LOGS_FORMAT, this);
-    if (!(variable instanceof CustomVariable)) {
-      throw new Error('Logs format variable not found');
-    }
-    return variable;
-  }
-
   private updateFields() {
-    const variable = this.getLogsFormatVariable();
     const disabledFields = [
       '__time',
       'timestamp',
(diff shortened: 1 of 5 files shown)


src/services/variableGetters.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomVariable, sceneGraph, SceneObject } from '@grafana/scenes';
+import { VAR_LOGS_FORMAT } from './variables';
+
+export function getLogsFormatVariable(sceneRef: SceneObject) {
```
gtk-grafana, line 4: we've got a bunch of variable getters spread around, should we collect them all somewhere?

xenartro: Yeah, we could use a clean up. Today I ran this quick search https://github.com/search?q=repo%3Agrafana%2Fexplore-logs%20sceneGraph.lookupVariable&type=code and, honestly, expected it to be worse.

matyax: Yeah, we could use a clean up. Today I ran this quick search https://github.com/search?q=repo%3Agrafana%2Fexplore-logs%20sceneGraph.lookupVariable&type=code and, honestly, expected it to be worse.

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -225,18 +214,15 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     if (newState.data?.state === LoadingState.Done) {
       const frame = newState.data?.series[0];
       if (frame) {
-        const res = extractParserAndFieldsFromDataFrame(frame);
+        console.log('serviceScene');
```
matyax, line 217:
```suggestion
```

matyax (commented):
Haven't tried these changes, but left a few suggestions.

src/Components/ServiceSelectionScene/SelectServiceButton.tsx
```diff
@@ -3,24 +3,54 @@ import React from 'react';
...
 
+function setParserIfFrameExistsForService(service: string, sceneRef: SceneObject) {
+  const serviceSelectionScene = sceneGraph.getAncestor(sceneRef, ServiceSelectionScene);
+
+  const theChildWeWant = serviceSelectionScene.state.body.state.children.find((child) => {
+    if (child instanceof SceneCSSGridItem) {
+      const body = child.state.body;
+
+      // The query runner is only defined for the logs panel
+      const queryRunner = body?.state.$data as SceneQueryRunner | undefined;
```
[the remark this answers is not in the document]
gtk-grafana: No, but I added some more runtime checks and removed the type assertions though

gtk-grafana:
Thanks for the feedback @matyax, should all be cleaned up, and I noticed that we aren't setting the parser correctly for legacy logs dataplane format, so I went ahead and fixed that too

src/services/variableGetters.ts
```diff
@@ -0,0 +1,10 @@
...
+  const variable = sceneGraph.lookupVariable(VAR_LOGS_FORMAT, sceneRef);
+  if (!(variable instanceof CustomVariable)) {
+    throw new Error('Logs format variable not found');
+  }
+  return variable;
+}
```
matyax, line 10: I think we could use variables.ts for these functions.

matyax (approved):
This is great! I would only ask if you can move the variable getter to variables.ts.

src/services/variableGetters.ts
matyax: https://github.com/grafana/explore-logs/pull/632

gtk-grafana:
@matyax see my comment on https://github.com/grafana/explore-logs/pull/632, adding imported functions to variables.ts breaks the build.

PR #631, opened 2024-07-24
Labels: Infinite loading state when detected_labels are empty on load

gtk-grafana:
So this fix might want to wait until after we have updated the detected_labels API to return single cardinality values, right now we show single-cardinality fields in the UI if the user has selected a label that reduces the cardinality to 1, because we don't update the detected_labels state when the array is empty.


gtk-grafana:
Closing until refactoring of resource calls to runtime datasource query calls is complete

PR #632, opened 2024-07-25
Variables: move variable lookup to getter functions

matyax:
A significant portion of the codebase reads or writes to variables. Being a common task, we're now moving that code to specific functions.

This aims to decrease unnecessary duplication and leakage of implementation details between scenes (e.g. having to know the exact class when you're only interested in the value).

### DIFF
--- a/src/Components/ServiceScene/LineFilterScene.tsx
+++ b/src/Components/ServiceScene/LineFilterScene.tsx
@@ -21,7 +21,7 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
   }
 
   private onActivate = () => {
-    const lineFilterValue = this.getVariable().getValue();
+    const lineFilterValue = getLineFilterVariable(this).getValue();
     if (!lineFilterValue) {
       return;
     }
@@ -33,14 +33,6 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
       lineFilter: matches[1].replace(/\\(.)/g, '$1'),
     });
   };
-
-  private getVariable() {
-    const variable = getLineFilterVariable(this);
-    if (!variable) {
-      throw new Error('Logs format variable not found');
-    }
-    return variable;
-  }
 
   updateFilter(lineFilter: string) {
     this.setState({
@@ -54,7 +46,7 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
   };
 
   updateVariable = debounce((search: string) => {
-    const variable = this.getVariable();
+    const variable = getLineFilterVariable(this);
     variable.changeValueTo(`|~ \`(?i)${escapeRegExp(search)}\``);
     reportAppInteraction(
       USER_EVENTS_PAGES.service_details,
--- a/src/Components/ServiceScene/LineFilterScene.tsx
+++ b/src/Components/ServiceScene/LineFilterScene.tsx
@@ -1,9 +1,9 @@
 import { css } from '@emotion/css';
-import { CustomVariable, SceneComponentProps, SceneObjectBase, SceneObjectState, sceneGraph } from '@grafana/scenes';
+import { SceneComponentProps, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
 import { Field } from '@grafana/ui';
 import { debounce, escapeRegExp } from 'lodash';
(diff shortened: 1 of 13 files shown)


src/services/variables.ts
```diff
@@ -17,3 +21,51 @@ export const EXPLORATION_DS = { uid: VAR_DATASOURCE_EXPR };
...
+export function getPatternsVariable(scene: SceneObject) {
+  const variable = sceneGraph.lookupVariable(VAR_PATTERNS, scene);
+  if (variable instanceof CustomVariable) {
+    return variable;
+  }
+  return null;
+}
+
+export function getLabelsVariable(scene: SceneObject) {
+  return getAdHocFiltersVariable(VAR_LABELS, scene);
```
svennergr, line 34: Guess we should move `getAdHocFiltersVariable` here?

svennergr (approved):
Nice!

gtk-grafana (changes requested):
@matyax Putting these in variables.ts breaks the build, I forget why, something to do with imports and PLUGIN_ID, which is why I moved the getters to a new file in https://github.com/grafana/explore-logs/pull/629

image

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -281,7 +289,13 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     const timeRange = sceneGraph.getTimeRange(this);
 
     const timeRangeValue = timeRange.state.value;
-    const filters = sceneGraph.lookupVariable(VAR_LABELS, this)! as AdHocFiltersVariable;
+    const filters = getLabelsVariable(this);
+
+    if (!filters) {
```
gtk-grafana, line 294: do we want to move these console.errors into the helpers?

src/Components/ServiceScene/LineFilterScene.tsx
```diff
@@ -35,8 +35,8 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
   };
 
   private getVariable() {
-    const variable = sceneGraph.lookupVariable(VAR_LINE_FILTER, this);
-    if (!(variable instanceof CustomVariable)) {
+    const variable = getLineFilterVariable(this);
+    if (!variable) {
       throw new Error('Logs format variable not found');
```
gtk-grafana, line 40: same question here, do we want the helper function to throw an error? Or just log a warning/error

src/Components/ServiceScene/ServiceScene.tsx
matyax: I've been thinking about it but I'm still unsure. There are places where we throw an Error if the variable is not found, so that means that we expect them to be defined all the time, or else the app explodes. If that is really the expectation, I would maybe move that to these functions, but we need to throw, or else we still need to return `null` and the consumer of the function needs to conditionally check for that.

matyax: To summarize. If the variable is not found, should we: 
- Assume it should be defined and returned by the function, or else throw.
- Assume it can potentially be undefined and let the Scene/Component handle it.

gtk-grafana: Assuming it should be defined and failing loudly is the best pattern IMO, but if we're worried about introducing runtime errors in some situation we're not currently aware of, maybe we just log the error for now, and come back and harden these up when we're confident this isn't happening in prod?

gtk-grafana: I guess it's still in public preview, so maybe being bold is the right answer right now

matyax: Yeah, silently failing is definitely the worst option. Let's fail loudly and see where it takes us. Thanks!

matyax: Done! This simplified the code quite a bit, which is very nice.

src/Components/ServiceScene/LineFilterScene.tsx
matyax: THROW!

matyax:
Should be good to merge. Let me know otherwise and I'll follow it up. Thanks everyone!

PR #633, opened 2024-07-25
Order agnostic state comparison

gtk-grafana:
We have some state comparisons causing unwanted re-rendering when the compared objects are the same, but the order is different.
Using lodash isEqual on its own will return false when the order changes, so we simply convert to sets, and compare the sets with lodash isEqual. Converting to a set should be more performant then sorting arrays and then comparing.

This also fixes a bug on activation of the LabelBreakdownScene which currently is triggering a re-query of the labels, causing a bunch of network requests to be executed which are immediately cancelled.


gtk-grafana:
Found a bug. Putting back in draft

src/services/comparison.ts
```diff
@@ -0,0 +1,23 @@
+import _ from 'lodash';
+
+/**
+ * Order doesn't matter, converts to sets, deep compare via lodash isEqual
+ * @param arr1
+ * @param arr2
+ */
+
+export const areArraysEqual = (arr1: any[] | undefined, arr2: any[] | undefined) => {
```
matyax, line 9: In general I would suggest to not accept undefined arguments, but since this is a very specific implementation to compare our state arrays, and simplifies the code on the other end, 👍 👍 👍

matyax (approved):
👏 👏

PR #639, opened 2024-07-26
feat(entrypoints): add entrypoints from panel and Explore

svennergr:
More of a hack day project. This PR adds entrypoints from panels and Explore:

https://github.com/user-attachments/assets/e6a4f207-faed-419f-9494-e7db4c8d69eb


src/services/extensions/links.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export const linkConfigs: PluginExtensionLinkConfig[] = [
+  {
+    extensionPointId: PluginExtensionPoints.DashboardPanelMenu,
+    title,
+    description,
+    icon,
+    path: createAppUrl(),
+    configure: (context?: PluginExtensionPanelContext) => {
+      if (!context) {
```
matyax, line 20: Lines 20:40 and 66:86 seem to be doing the same thing. Maybe extract to a common place?

src/services/routing.ts
```diff
@@ -138,3 +138,33 @@ export function buildServicesRoute(extraQueryParams?: UrlQueryMap): UrlQueryMap
     ...extraQueryParams,
   };
 }
+
+export function createAppUrl(path = '/explore', urlParams?: URLSearchParams): string {
+  return `/a/${pluginJson.id}${path}${urlParams ? `?${urlParams.toString()}` : ''}`;
+}
+
+export enum UrlParameterType {
+  DatasourceId = 'var-ds',
```
matyax, line 147: Can be taken from variables.ts, right?

matyax:
```suggestion
  DatasourceId = `var-${VAR_NAME}`,
```

Not sure if this is possible with TS.

matyax (commented):
Looks great but couldn't get it to work locally.

src/services/routing.ts
svennergr: Yea, it's not, but I switched to using a `const`. Good recommendation.

src/services/extensions/links.ts
svennergr: Sounds good!

svennergr:
@matyax in order to test, since this needs preloading of the plugin, make sure to build the plugin first and then start Grafana. Or restart Grafana after the plugin was built.

matyax (approved):
:rocket:

PR #642, opened 2024-07-26
Hack day: outlier detection

matyax:
POC / WIP

Experiment. Not meant to be merged.


sd2k (commented):
Nice, this is really cool! From an ML point of view I think this is great.

The annotations do seem to get a bit noisy 😅 I wonder if it'd be better to annotate a range between first and last outlying point or something. Or, omit the annotations and colour the series differently depending if they're outlying or not - I think that's what the scenes-ml outlier detector does (e.g. in the [docs screenshot](https://grafana-dev.com/developers/scenes/scenes-ml/outlier-detection)).

One other thing: the latest version of augurs includes an alternative outlier detection implementation (MAD) which behaves slightly differently - the two are documented [here](https://grafana.com/docs/grafana-cloud/alerting-and-irm/machine-learning/forecasts/outlier-detection/#select-the-detection-algorithm-and-sensitivity), perhaps it's worth discussing which is better for this use case (although I detect it might vary...)

PR #644, opened 2024-07-29
Logs Volume: track filtering interactions

matyax:
Last week we added filtering support from Logs Volume, but we have no visibility on the usage of this feature.


gtk-grafana (approved):
Good catch

PR #645, opened 2024-07-30
Fields breakdown: Executing aggregation query twice

gtk-grafana:
This is running an extra query we don't want to run, and it causes the whole page to flicker.

To reproduce: You must navigate to the field label view, and then select any field value for the first time (the second query is not run on subsequent navigations when the scene is already cached).

Sometimes the first duplicate query is cancelled, sometimes not, but we always fire the query twice with the exact same query args on the first view of any field value breakdown.

The fix:
Don't call updateBody when the variable loading state changes, just set the loading state of the scene (or do nothing TODO)


gtk-grafana:
Ugh, this only helps sometimes, need a better solution for this race condition

PR #647, opened 2024-07-31
feat(feedback): move feedback button

svennergr:
Moves the feedback button:
https://raintank-corp.slack.com/archives/C05FWM2DN07/p1722364395732199

image


matyax:
Looks great there, but I would at least duplicate the gap between "Feedback" and the time picker.

test

matyax:
E.g.
Gap 8px

imagen


Gap 4px

imagen

matyax (approved):

matyax:
Maybe we need to position the button without changing the gap, because it's currently misaligned, and will be even more with a bigger gap.

gap

svennergr:
@matyax not the biggest fan, but what do you think about

```
      position: 'relative',
      top: theme.spacing(-1),
```
image

matyax:
Yeah, that should do it 👍 . Thanks!

matyax (approved):
:stamp:

PR #651, opened 2024-07-31
WIP: Replacing datasource resource calls with query runners

gtk-grafana:
WIP - interpolation is not working, and some functionality in the service selection scene is still broken


src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -102,128 +104,37 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
...
-          const ds = getDataSourceVariable(this).getValue()?.toString();
-          let servicesToQuery: string[] = [];
-          if (ds && this.state.servicesByVolume) {
-            servicesToQuery = createListOfServicesToQuery(
-              this.state.servicesByVolume,
-              ds,
-              newState.searchServicesString
-            );
-          }
+    this.state.$data.subscribeToState((newState) => {
```
gtk-grafana, line 111: add subscription to subs

src/Components/ServiceScene/Breakdowns/FieldSelector.tsx
```diff
@@ -50,6 +50,7 @@ export function ServiceFieldSelector({
   selectOption,
   isLoading,
 }: AsyncFieldSelectorProps) {
+  //@todo why is the custom option broken?
```
gtk-grafana, line 53: clean up

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -440,41 +363,6 @@ function createListOfServicesToQuery(services: string[], ds: string, searchStrin
   return Array.from(new Set([...favoriteServicesToQuery, ...servicesToQuery]));
 }
 
-function shouldUpdateServicesByVolume(newTime: TimeRange, oldTime: TimeRange) {
```
gtk-grafana, line 443: is this something we care to keep?

yarn.lock
```diff
@@ -2,15 +2,10 @@
 # yarn lockfile v1
 
 
-"@aashutoshrathi/word-wrap@^1.2.3":
```
gtk-grafana, line 5: revert?

src/services/datasource.ts
```diff
@@ -25,26 +45,85 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+      'index/volume',
+      {
+        expr: targetsInterpolated[0].expr.replace(VAR_SERVICE_EXPR_HACK, ''),
+        query: targetsInterpolated[0].expr.replace(VAR_SERVICE_EXPR_HACK, ''),
+        from: request.range.from.utc().toISOString(),
+        to: request.range.to.utc().toISOString(),
+        limit: 1000,
+      },
+      {
+        headers: {
```
svennergr, line 106:
```suggestion
        requestId: request.requestId ?? 'volume',
        headers: {
```

This gets us automatic request cancellation. I noticed, when switching datasource, those inflight requests were not cancelled.

svennergr:
@gtk-grafana I don't think you need the `VAR_SERVICE_EXPR_HACK`. What I did in `ServiceSelectionScene` to make this work:

1. Add:
```ts
    variableNames: [VAR_DATASOURCE],
    onReferencedVariableValueChanged: this.onReferencedVariableValueChanged.bind(this),
  });
```

2. `onReferencedVariableValueChanged`
```ts
  private onReferencedVariableValueChanged(variable: SceneVariable) {
    if (variable.state.name === VAR_DATASOURCE) {
      if (this.state.$data instanceof SceneQueryRunner) {
        this.state.$data.runQueries();
      }
      return;
    }
  }
```

src/services/datasource.ts
gtk-grafana: Great catch, can confirm, thank you Sven!

gtk-grafana:
Thanks Sven, should be updated now, and I also added an e2e test to ensure that changing the datasource re-triggers the volume query

src/services/datasource.ts
```diff
@@ -25,26 +44,82 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+              const dsResponse = ds.query(request);
+              dsResponse.subscribe(subscriber);
+              break;
+            }
           }
         });
     });
   }
 
+  private transformVolumeResponse(
```
svennergr, line 80: Maybe rename this function into something like `requestVolume` / `getVolume`? I think `transformResponse` is just a "minor" part of this method.

src/services/datasource.ts
```diff
@@ -25,26 +44,82 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+    if (request.targets.length !== 1) {
+      throw new Error('Volume query can only have a single target!');
+    }
+
+    const targetsInterpolated = ds.interpolateVariablesInQueries(request.targets, request.scopedVars);
+
+    const dsResponse = ds.getResource(
+      'index/volume',
+      {
+        expr: targetsInterpolated[0].expr,
```
svennergr, line 94:
```suggestion
```

I didn't find this in https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-log-volume

src/services/datasource.ts
```diff
@@ -25,26 +44,82 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+        requestId: request.requestId ?? 'volume',
+        headers: {
+          'X-Query-Tags': `Source=${PLUGIN_ID}`,
+        },
+      }
+    );
+    dsResponse.then((response: IndexVolumeResponse | undefined) => {
+      response?.data.result.sort((lhs, rhs) => {
+        return Number(rhs.value[1]) - Number(lhs.value[1]);
+      });
```
svennergr, line 110: Can we get some comment that `value[0]` is the timestamp and `value[1]` is the value?

tests/exploreServices.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { test, expect } from '@grafana/plugin-e2e';
 import { ExplorePage } from './fixtures/explore';
 import {testIds} from "../src/services/testIds";
+import {mockVolumeApiResponse} from "./mocks/mockVolumeApiResponse";
```
svennergr, line 4: I know it's tests, but those imports look different in formatting.

svennergr (approved):
Awesome! Love the tests! Left a few nits, but great stuff.

src/services/datasource.ts
gtk-grafana: yeah this was a leftover from me playing around, good catch

src/services/datasource.ts
gtk-grafana: we should have a type somewhere that describes this, I'll add a type assertion or something

tests/exploreServices.spec.ts
gtk-grafana: oh yeah, the linter doesn't run in the test files, I'll fix!

src/services/query.ts
```diff
@@ -13,6 +14,25 @@ export type LokiQuery = {
...
+
+/**
+ * Builds the resource query
+ * @param expr string to be interpolated and executed in the resource request
+ * @param resource
+ * @param queryParamsOverrides
+ */
+export const buildResourceQuery = (
+  expr: string,
+  resource: 'volume' | 'patterns' | 'detected_labels',
```
matyax, line 26: We're not using this for logs? 0 context, just asking.

svennergr: Logs use `buildLokiQuery` already now - so they won't be resource calls but "usual" queries. @gtk-grafana maybe worth renaming `buildLokiQuery` to `buildDataQuery` to match the "names" we use in the datasource backend?

gtk-grafana:
Thanks for the feedback, everything should be updated now!

Quick bug fix: I noticed that when running queries against loki-ops I was still getting the cancelled queries I worked so hard to fix, found out it was the loading state was gating the render of the body, which was cleaning up the existing query runners. Dunno why it wasn't happening with the gdev-loki datasource, but I just changed the render method to show the body when we have services, so we don't clean up all those query runners for panels for a specific service that will probably be in the next request as well.
If the service isn't in the new volume request, it will trigger a query that gets cancelled when the volume comes back, but I think we're stuck with that unless we can set a dependency on a query runner that says don't update your query on time range change until the volume call completes.

src/services/query.ts
gtk-grafana: Done!

PR #654, opened 2024-08-01
Revert "chore: upgrade augurs"

svennergr:
Reverts grafana/explore-logs#650


matyax:
We shoulnd't need this one 🤞

PR #658, opened 2024-08-02
Sorting: add outlier detection as a sorting option

matyax:
This PR adds outlier detection to sort value breakdowns. It works by using [DBSCAN](https://grafana.com/docs/grafana-cloud/alerting-and-irm/machine-learning/forecasts/outlier-detection/#select-the-detection-algorithm-and-sensitivity) to detect outlying intervals, and sorting by the detected quantity.

Addtionally, we're wrapping all the WASM invocations, logging the errors and falling back to StdDev when it happens.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/SortByScene.tsx
@@ -31,7 +31,7 @@ export class SortByScene extends SceneObjectBase<SortBySceneState> {
         {
           value: 'outliers',
           label: 'Detected outliers',
-          description: 'Order by detected outliers in the data',
+          description: 'Order by the amount of detected outliers in the data',
         },
         {
           value: ReducerID.stdDev,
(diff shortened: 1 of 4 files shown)


src/Components/ServiceScene/Breakdowns/SortByScene.tsx
```diff
@@ -28,6 +28,11 @@ export class SortByScene extends SceneObjectBase<SortBySceneState> {
           label: 'Most relevant',
           description: 'Smart ordering of graphs based on the most significant spikes in the data',
         },
+        {
+          value: 'outliers',
+          label: 'Detected outliers',
+          description: 'Order by detected outliers in the data',
```
svennergr, line 34: What does that mean, btw? Is it more "Order by the count/amount of outliers"?

svennergr:
I think we need to also gate https://github.com/grafana/explore-logs/blob/85085a9643def5e8ecb103aa43faa2ea8abcd971/src/module.ts#L8 to only init when wasm is supported.

src/Components/ServiceScene/Breakdowns/SortByScene.tsx
matyax: It's a good question. Let's add more details to the description.

matyax:
Right! Great idea, done.

gtk-grafana (approved):
LGTM and working well locally

PR #659, opened 2024-08-02
ServiceScene: Fix invalid queries being called when navigating to service selection

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/657

### DIFF
--- a/src/Components/ServiceScene/ServiceScene.tsx
+++ b/src/Components/ServiceScene/ServiceScene.tsx
@@ -13,7 +13,7 @@ import {
   SceneVariable,
   VariableDependencyConfig,
 } from '@grafana/scenes';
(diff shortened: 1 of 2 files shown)


src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -309,6 +332,10 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     const breakdownView = getDrilldownSlug();
     const breakdownViewDef = breakdownViewsDefinitions.find((v) => v.value === breakdownView);
 
+    if (!body) {
```
gtk-grafana, line 335: the body is always added first in the activation, and setState is synchronous, so it should never be undefined at runtime. Figured a hard error is best way to sniff out if that changes in the future?

matyax (approved):

PR #666, opened 2024-08-02
Service selection scene: get service volume by partial or exact match

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/663

Note for reviewers: Needed to use `.*` instead of `.+`. `.+` requires additional chars of context to match. To get around the empty query being `.*.*` I added a replace in the runtime datasource after interpolation to replace `.*.*` with `.+`.

Also added some more e2e test coverage for these cases. 

Also updated the port of the grafana e2e instance because recently I started seeing my local grafana instance instead of the one in the docker container.


matyax (approved):
Tested the search and works as expected 👌

PR #668, opened 2024-08-06
Service selection: Results shown from wrong datasource after switching

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/662

### DIFF
--- a/src/services/query.ts
+++ b/src/services/query.ts
@@ -21,7 +21,6 @@ export type LokiQuery = {
  * @param expr string to be interpolated and executed in the resource request
  * @param resource
  * @param queryParamsOverrides
- * @param datasource
  */
 export const buildResourceQuery = (
   expr: string,
--- a/src/services/query.ts
+++ b/src/services/query.ts
@@ -1,4 +1,5 @@
-import { AdHocVariableFilter, DataSourceApi } from '@grafana/data';
+import { AdHocVariableFilter } from '@grafana/data';
+import { DataSourceRef } from '@grafana/schema';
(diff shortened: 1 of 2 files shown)


src/services/query.ts
```diff
@@ -30,6 +31,7 @@ export const buildResourceQuery = (
     ...defaultQueryParams,
     resource,
     ...queryParamsOverrides,
+    datasource: { uid: VAR_DATASOURCE_EXPR } as DataSourceApi,
```
svennergr, line 34: Interesting, I think the type in `LokiQuery` should have been:

```
datasource?: DataSourceRef;
```

Mind changing that so you don't need the type cast here.

src/services/query.ts
```diff
@@ -30,6 +31,7 @@ export const buildResourceQuery = (
     ...defaultQueryParams,
     resource,
     ...queryParamsOverrides,
+    datasource: { uid: VAR_DATASOURCE_EXPR },
```
svennergr, line 34: I think we would need the same in `buildDataQuery`, right?

svennergr (approved):
LGTM

src/services/query.ts
gtk-grafana: Good question, I think since it always redirects back to the service selection scene whenever the datasource is changed that we won't see the impact yet?

Although when I make this change locally I do see some queries that get run after the datasource change, but before the navigation that get cancelled immediately. 

Let's not change it in this PR, and I'll keep this in mind as I continue the refactoring into the runtime datasource.

PR #672, opened 2024-08-07
Patterns: Use runtime datasource for queries

gtk-grafana:
Refactoring the patterns call to use runtime datasource.

Fixes: https://github.com/grafana/explore-logs/issues/670

This is a pretty big refactor, but the end goal should be that we don't re-query the patterns on every route change, just when:
1. You change the time range
1. You change a label (but we get 2 queries one before the route change and one after, the first one gets cancelled, this is acceptable for now, but we should do https://github.com/grafana/explore-logs/issues/640 next, and then follow up with adding the detected_labels to runtime datasource and clean up any remaining hacks to stop duplicate queries on route changes)
1. You view the patterns route (for the first time, or 10% of time range has elapsed)
1. The first time you view a route with a particular service selected

Ok, I've tried a few different approaches with this, and while this "requires" (not really required, but it would be good to avoid the [hack to call a private method](https://github.com/grafana/explore-logs/blob/c3a6182bb9b0b8dafe89acad7b83355a2caf18b3/src/services/LogsSceneQueryRunner.ts)) a scenes core change (https://github.com/grafana/scenes/pull/866), I think this is the cleanest approach so far.

Some history: Initially I saw that the special $data provider didn't allow multiple query runners, but did allow multiple queries, so I tried stuffing the patterns resource query in with the typical data api call. But this broke a bunch of assumptions that scenes has about the data being returned by a query runner, primarily we had multiple queries that called different api endpoints, so the results would come in at different times. I could work around this by holding the "done" state until all queries had resolved, but then if you had one query taking longer then the other, you'd either need to change the assumption that you wait for a query state to be "done" before doing stuff with the response, but this would cause unnecessary re-rendering in panelBuilders that directly consumed data providers, as sometimes the dataframes would be undefined when it was actually loading.
Anyway, that was a terrible idea.

So instead of trying to use $data, in this current attempt we're just adding more custom data providers to the state of the ServiceScene `$patternsData`. This works much better, but if I activated the query provider in the serviceScene we'd still get unnecessary executions of the patterns call, for example on every route change. But we just want to run the patterns call on specific route activations, or when the time range changes, or the labels variable has been updated, or the user is entering the application for the first time. But we can't move the data provider down the scene tree as we expose the count of the patterns in the tabs which are visible across all routes.

So currently we're never activating the patternsData query runner, but we call runQueries in the specific cases where we want it to update. This works fine except we need to extend the `SceneQueryRunner` class and prevent `runQueries` from calling `subscribeToTimeRangeChanges`, as this will make the exact same request twice (but only the first time a scene is activated).

This should give us the desired functionality to only request the patterns when we need it, instead of on every route change that has ServiceScene as a parent (which is the entire application, except the service selection scene), and still take advantage of query caching caching within scenes.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -34,9 +34,9 @@ export default defineConfig<PluginOptions>({
     trace: 'on-first-retry',
 
     // Turn on when debugging local tests
-    // video: {
-    //   mode: 'on',
-    // }
+    video: {
+      mode: 'on-first-retry',
+    }
   },
   expect: { timeout: 15000 },
 
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -35,7 +35,7 @@ export default defineConfig<PluginOptions>({
 
     // Turn on when debugging local tests
     video: {
-      mode: 'on-first-retry',
+      mode: 'on',
     }
   },
   expect: { timeout: 15000 },
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -34,9 +34,9 @@ export default defineConfig<PluginOptions>({
     trace: 'on-first-retry',
 
     // Turn on when debugging local tests
-    video: {
-      mode: 'on',
-    }
+    // video: {
+    //   mode: 'on',
+    // }
   },
   expect: { timeout: 15000 },
 
--- a/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
@@ -218,9 +218,9 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
           body: PanelBuilders.timeseries()
             .setTitle(optionValue)
             .setData(
-              getQueryRunner(
-                buildDataQuery(getTimeSeriesExpr(this, optionValue), { legendFormat: `{{${optionValue}}}` })
-              )
+              getQueryRunner([
+                buildDataQuery(getTimeSeriesExpr(this, optionValue), { legendFormat: `{{${optionValue}}}` }),
+              ])
             )
             .setHeaderActions(new SelectLabelAction({ labelName: optionValue }))
             .setCustomFieldConfig('stacking', { mode: StackingMode.Normal })
@@ -276,7 +276,7 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
     const getFilter = () => this.state.search.state.filter ?? '';
 
     return new LayoutSwitcher({
-      $data: getQueryRunner(query),
+      $data: getQueryRunner([query]),
       options: [
         { value: 'single', label: 'Single' },
         { value: 'grid', label: 'Grid' },
--- a/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
@@ -295,6 +295,7 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
         }),
         new ByFrameRepeater({
           body: new SceneCSSGridLayout({
+            isLazy: true,
             templateColumns: GRID_TEMPLATE_COLUMNS,
             autoRows: '200px',
             children: [
--- a/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
@@ -424,16 +424,6 @@ function getStyles(theme: GrafanaTheme2) {
 
 const GRID_TEMPLATE_COLUMNS = 'repeat(auto-fit, minmax(400px, 1fr))';
 
-export function buildLabelBreakdownActionScene() {
-  return new SceneFlexLayout({
-    children: [
-      new SceneFlexItem({
-        body: new LabelBreakdownScene({}),
-      }),
-    ],
-  });
-}
-
 export function buildLabelValuesBreakdownActionScene(value: string) {
   return new SceneFlexLayout({
     children: [
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
@@ -138,7 +138,6 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
 
   private updatePatternFrames(lokiPatterns?: DataFrame[]) {
     if (!lokiPatterns) {
-      console.warn('failed to update pattern frames');
       return;
     }
 
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
@@ -181,7 +181,7 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
       const sum: number = dataFrame.meta?.custom?.sum;
       const patternFrame: PatternFrame = {
         dataFrame,
-        pattern: dataFrame.name as string,
+        pattern: dataFrame.name ?? '',
         sum,
         status: existingPattern?.type,
       };
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
@@ -22,7 +22,6 @@ import { PatternsFrameScene } from './PatternsFrameScene';
 import { PatternsViewTextSearch } from './PatternsViewTextSearch';
 import { PatternsNotDetected, PatternsTooOld } from './PatternsNotDetected';
 import { areArraysEqual } from '../../../../services/comparison';
-import { Unsubscribable } from 'rxjs';
 
 export interface PatternsBreakdownSceneState extends SceneObjectState {
   body?: SceneFlexLayout;
@@ -36,7 +35,6 @@ export interface PatternsBreakdownSceneState extends SceneObjectState {
   // Subset of patternFrames, undefined if empty, empty array if search results returned nothing (no data)
   filteredPatterns?: PatternFrame[];
   patternFilter: string;
-  dataSub?: Unsubscribable;
 }
 
 export type PatternFrame = {
@@ -69,8 +67,8 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
     const { body, loading, blockingMessage } = model.useState();
     const { value: timeRange } = sceneGraph.getTimeRange(model).useState();
     const logsByServiceScene = sceneGraph.getAncestor(model, ServiceScene);
-    const { $data } = logsByServiceScene.useState();
-    const patterns = getPatternsFrames($data?.state.data);
+    const { $patternsData } = logsByServiceScene.useState();
+    const patterns = getPatternsFrames($patternsData?.state.data);
     const styles = useStyles2(getStyles);
     const timeRangeTooOld = dateTime().diff(timeRange.to, 'hours') >= PATTERNS_MAX_AGE_HOURS;
 
@@ -105,7 +103,7 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
     const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
     this.setBody();
 
-    const patterns = getPatternsFrames(serviceScene.state.$data?.state.data);
+    const patterns = getPatternsFrames(serviceScene.state.$patternsData?.state.data);
 
     // If the patterns exist already, update the dataframe
     if (patterns) {
@@ -113,23 +111,7 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
     }
 
     // Subscribe to changes from pattern API call
-    const dataSub = serviceScene.state.$data.subscribeToState(this.onDataProviderChange);
-    this.setState({
-      dataSub,
-    });
-    this._subs.add(dataSub);
-
-    // Subscribe to changes to the query provider
-    serviceScene.subscribeToState((newState, prevState) => {
-      if (newState.$data.state.key !== prevState.$data.state.key) {
-        const dataSub = serviceScene.state.$data.subscribeToState(this.onDataProviderChange);
-        this.state.dataSub?.unsubscribe();
-        this.setState({
-          dataSub,
-          loading: true,
-        });
-      }
-    });
+    this._subs.add(serviceScene.state.$patternsData.subscribeToState(this.onDataProviderChange));
   }
 
   private onDataProviderChange = (newState: SceneDataState, prevState: SceneDataState) => {
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
@@ -111,10 +111,10 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
     }
 
     // Subscribe to changes from pattern API call
-    this._subs.add(serviceScene.state.$patternsData.subscribeToState(this.onDataProviderChange));
+    this._subs.add(serviceScene.state.$patternsData.subscribeToState(this.onDataChange));
   }
 
-  private onDataProviderChange = (newState: SceneDataState, prevState: SceneDataState) => {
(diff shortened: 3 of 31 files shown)


src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -218,9 +218,9 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
           body: PanelBuilders.timeseries()
             .setTitle(optionValue)
             .setData(
-              getQueryRunner(
-                buildDataQuery(getTimeSeriesExpr(this, optionValue), { legendFormat: `{{${optionValue}}}` })
-              )
+              getQueryRunner([
```
gtk-grafana, line 221: We don't need multiple queries anymore, and I would highly suggest we don't mix queries that call different API endpoints, but if we wanted to include multiple queries that call the same API endpoint this should work.

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -238,6 +160,18 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
 
   private getVizPanel() {
     this.lineFilterScene = new LineFilterScene();
+
+    // If this is called before the query has executed, we need to return an empty data node to init the loading state, instead of no-data
+    const data =
```
gtk-grafana, line 165: I think this can be cleaned up now, this was a workaround to stop the flickering caused by having multiple queries to different endpoints in the same query provider.

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  public onActivate() {
+    if (!this.state.body) {
+      this.setState({
+        body: this.getLogsPanel(),
+      });
+    }
+
+    this._subs.add(
```
gtk-grafana, line 46: clean this up

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -1,22 +1,59 @@
...
+   * We can't subscribe to the state of the data provider anymore, because there are multiple queries running in each data provider
+   * So we need to manually update the data state to prevent unnecessary re-renders that cause flickering and break loading states
+   */
+  public onActivate() {
+    this._subs.add(
+      sceneGraph.getData(this).subscribeToState((newState, prevState) => {
+        const dataFrame = getLogsPanelFrame(newState.data);
+
+        // Just this query is done
+        if (dataFrame) {
```
gtk-grafana, line 35: clean this up

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -43,24 +80,24 @@ export class LogsTableScene extends SceneObjectBase {
       }
     };
 
-    const styles = getStyles();
+    const dataFrame = getLogsPanelFrame(data?.data);
```
gtk-grafana, line 83: clean this up

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -371,153 +351,35 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
+// @todo move to datasource or queries?
+export function getPatternsQueryRunner() {
+  return getResourceQueryRunner([buildResourceQuery(VAR_LABELS_EXPR, 'patterns', { refId: PATTERNS_QUERY_REFID })]);
+}
+
+function getServiceSceneQueryRunner() {
+  return getQueryRunner([buildDataQuery(LOG_STREAM_SELECTOR_EXPR, { refId: LOGS_PANEL_QUERY_REFID })]);
+}
+//
+// function getQueryRunnerFromProvider(queryRunner: SceneDataProvider): SceneQueryRunner {
```
gtk-grafana, line 375: delete

src/services/datasource.ts
```diff
@@ -32,13 +39,29 @@ type IndexVolumeResponse = {
...
   constructor(pluginId: string, uid: string) {
     super(pluginId, uid);
   }
 
   query(request: SceneDataQueryRequest): Promise<DataQueryResponse> | Observable<DataQueryResponse> {
-    return new Observable<DataQueryResponse>((subscriber) => {
+    const numberOfQueries = request.targets.length;
+    let numberOfQueriesThatResolved = 0;
+
+    const observable = new Observable<DataQueryResponse>((subscriber) => {
```
gtk-grafana, line 64: clean this up too, be good to throw an error when a resource call makes it to the data request

playwright.config.ts
```diff
@@ -34,9 +34,9 @@ export default defineConfig<PluginOptions>({
     trace: 'on-first-retry',
 
     // Turn on when debugging local tests
-    // video: {
-    //   mode: 'on',
-    // }
+    video: {
```
gtk-grafana, line 37: revert?

src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
```diff
@@ -101,21 +103,25 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
...
-      serviceScene.subscribeToState((newState, prevState) => {
-        if (!areArraysEqual(newState.patterns, prevState.patterns)) {
-          this.updatePatternFrames(newState.patterns);
-        }
-      })
-    );
+    this._subs.add(serviceScene.state.$patternsData.subscribeToState(this.onDataProviderChange));
   }
 
+  private onDataProviderChange = (newState: SceneDataState, prevState: SceneDataState) => {
```
gtk-grafana, line 117: this isn't needed anymore either

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -8,15 +8,28 @@ import { LogsPanelHeaderActions } from '../Table/LogsHeaderActions';
...
+    super(state);
+
+    this.addActivationHandler(this.onActivate.bind(this));
+  }
+
+  /**
+   * We can't subscribe to the state of the data provider anymore, because there are multiple queries running in each data provider
+   * So we need to manually update the data state to prevent unnecessary re-renders that cause flickering and break loading states
+   */
+  public onActivate() {}
```
gtk-grafana, line 26: not needed, remove the comment too

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -227,76 +252,31 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
-        const res = updateParserFromDataFrame(frame, this);
-        const fields = res.fields.filter((f) => !disabledFields.includes(f)).sort((a, b) => a.localeCompare(b));
-        if (!areArraysEqual(fields, this.state.fields)) {
-          this.setState({
-            fields: fields,
-            loading: false,
-          });
-        }
-      } else {
+    const frame = getLogsPanelFrame(newState.data);
```
gtk-grafana, line 255: add the loading state back in

src/Components/Table/TableProvider.tsx
```diff
@@ -7,7 +7,7 @@ import { parseLogsFrame } from '../../services/logsFrame';
 import { SelectedTableRow } from './LogLineCellComponent';
 
 interface TableProviderProps {
-  dataFrame: DataFrame;
+  dataFrame?: DataFrame;
```
gtk-grafana, line 10: y tho

src/services/LogsSceneQueryRunner.ts
```diff
@@ -0,0 +1,18 @@
+import { QueryRunnerState, sceneGraph, SceneQueryRunner } from '@grafana/scenes';
+
+export class LogsSceneQueryRunner extends SceneQueryRunner {
+  constructor(initialState: QueryRunnerState) {
+    super(initialState);
+  }
+
+  public runQueries() {
```
gtk-grafana, line 8: update after we hear back on https://github.com/grafana/scenes/pull/866

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -1,84 +1,69 @@
...
 }
 
 export interface ServiceSceneState extends SceneObjectState, ServiceSceneCustomState {
   body: SceneFlexLayout | undefined;
   drillDownLabel?: string;
+  $data: SceneDataProvider;
+  $patternsData: SceneQueryRunner;
+}
+
+export function getLogsPanelFrame(data: PanelData | undefined) {
```
gtk-grafana, line 61: These functions aren't doing anything anymore. We should remove?

matyax:
I haven't tested this yet, so I plan to do a more in-depth review tomorrow, but took a good look at the changes and didn't find anything that stands out, aka, looks pretty good 👌

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
matyax: At first glance, I like this ability of running more than one query. What was the original purpose of this change?

gtk-grafana: Initially I tried to smoosh all the queries into $data, which was a bad idea as queries would end at different times, but if we have multiple data queries, that should work fine as they are all returned from the backend at the same time

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -149,24 +139,55 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
 
     this.setBreakdownView();
     this.setEmptyFiltersRedirection();
+    const slug = getDrilldownSlug();
+
+    // If we don't have a patterns count in the tabs, or we are activating the patterns scene, run the pattern query
+    if (this.state.patternsCount === undefined || slug === 'patterns') {
+      this.state.$patternsData.runQueries();
```
gtk-grafana, line 146: This will run uncessarily when the scene has already been activated, check for data first!

src/Components/ServiceScene/Breakdowns/PatternsLogsSampleScene.tsx
svennergr, line 1: Unrelated to this PR, but this file should live in the `Patterns` directory, no?

src/Components/ServiceScene/LogsActionBarScene.tsx
svennergr, line 1: Maybe just `ActionBarScene` instead of `LogsActionBarScene`?

src/Components/ServiceScene/LogsActionBarScene.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+        <TabsBar>
+          {breakdownViewsDefinitions.map((tab, index) => {
+            return (
+              <Tab
+                data-testid={tab.testId}
+                key={index}
+                label={tab.displayName}
+                active={currentBreakdownViewSlug === tab.value}
+                counter={!loading ? getCounter(tab, { ...state, $data }) : undefined}
```
svennergr, line 55:
```suggestion
                counter={loading ? undefined : getCounter(tab, { ...state, $data })}
```

just a nit, but imho a bit easier to read, especially with the `icon` line below.

src/Components/ServiceScene/LogsActionBarScene.tsx
```diff
@@ -0,0 +1,111 @@
...
+                icon={loading ? 'spinner' : undefined}
+                onChangeTab={() => {
+                  if (tab.value !== currentBreakdownViewSlug || allowNavToParent) {
```
svennergr, line 58:
```suggestion
                  if (tab.vaue && (tab.value !== currentBreakdownViewSlug || allowNavToParent)) {
```

To get rid of the nested `if` in line 67, maybe?

src/services/datasource.ts
```diff
@@ -50,35 +70,151 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+  private getData(
+    request: SceneDataQueryRequest,
+    ds: DataSourceWithBackend<DataQuery>,
+    subscriber: Subscriber<DataQueryResponse>
+  ) {
+    const dataQueryRequest = { ...request };
+    // override the target datasource to Loki
+    dataQueryRequest.targets = request.targets.filter((target) => {
+      return !target.resource;
+    });
```
svennergr, line 111: AFAIK this is already done in `query`?

```suggestion
```

src/services/datasource.ts
```diff
@@ -50,35 +70,151 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+
+    if (!dataQueryRequest.targets.length) {
+      throw new Error('No valid queries!');
+    }
```
svennergr, line 115:
```suggestion
```

Also, this can't really happen as we call it when we loop through `dataQueryRequest.targets`.

src/services/datasource.ts
```diff
@@ -50,35 +70,151 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+              case 'volume': {
+                this.getVolume(dataQueryRequest, ds, subscriber);
+                break;
+              }
+              case 'patterns': {
+                this.getPatterns(dataQueryRequest, ds, subscriber);
+                break;
+              }
+              default: {
+                this.getData(dataQueryRequest, ds, subscriber);
```
svennergr, line 93: After reviewing `getData`, aren't we sending multiple `targets` when we do this in the loop through `dataQueryRequest.targets`? I guess it should happen outside, maybe? Otherwise, not sure I follow.

svennergr (commented):
Looking great, works very good. Left a couple of comments, but nothing too big. Good job!

matyax:
So far it works very well, but I found this case where the query is not re-run after removing the last filter. Can consistently reproduce with any field.

https://github.com/user-attachments/assets/11a96aa3-cbd7-4148-867f-113fe3f6199d

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
gtk-grafana: Good point, that's probably worth opening in another issue. Might be good for the lazy-loaded panels as well to at least batch things up a bit.

src/Components/ServiceScene/LogsActionBarScene.tsx
matyax: +1 ActionBarScene.

src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
```diff
@@ -66,7 +67,8 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
     const { body, loading, blockingMessage } = model.useState();
     const { value: timeRange } = sceneGraph.getTimeRange(model).useState();
     const logsByServiceScene = sceneGraph.getAncestor(model, ServiceScene);
-    const { patterns } = logsByServiceScene.useState();
+    const { $patternsData } = logsByServiceScene.useState();
+    const patterns = getPatternsFrames($patternsData?.state.data);
```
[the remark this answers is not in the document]
gtk-grafana: No we do not, and it would introduce extra renders. Good catch!

src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
```diff
@@ -133,74 +139,37 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
     });
   }
 
-  private updatePatternFrames(lokiPatterns?: LokiPattern[]) {
+  private updatePatternFrames(lokiPatterns?: DataFrame[]) {
     if (!lokiPatterns) {
       console.warn('failed to update pattern frames');
```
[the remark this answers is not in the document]
gtk-grafana: No, but it would be nice to have more verbose logs in dev. Removed!

src/Components/ServiceScene/Breakdowns/PatternsLogsSampleScene.tsx
gtk-grafana: Yeah probably, worth moving in this PR or should we follow up?

svennergr: Your call. I think it's fine moving here.

src/services/datasource.ts
gtk-grafana: Yeah good catch, I added an assertion that we only have a single resource per request

gtk-grafana: Aaaand removed the unnecessary iteration

src/Components/ServiceScene/Breakdowns/PatternsLogsSampleScene.tsx
gtk-grafana: Done!

matyax:
I can no longer reproduce ☝️

matyax (approved):
To the moon 🚀

svennergr (approved):
LGTM

PR #673, opened 2024-08-09
add play link to demonstrate product

moxious:
Via docs shortcode, we can give users the opportunity to directly play with and inspect the feature on Grafana Play

### DIFF
--- a/docs/sources/_index.md
+++ b/docs/sources/_index.md
@@ -52,6 +52,9 @@ Using Explore Logs you can:
 - Browse automatic visualizations of your log data based on its characteristics.
 - Do all of this without writing LogQL queries.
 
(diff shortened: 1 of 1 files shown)


moxious:
If this PR and doing things like this is desirable, it should be straightforward to add links to other docs pages that illustrate the various features we're trying to show off.  These would generally be placed above but near the "Try it out" instructions on subpages.

The idea is that many such pages have this sorta structure:

- Here's the cool thing ("you can do log patterns")
- Here's a clickable "see right now" example on Play of the thing doing its thing.
- Here's a walkthrough of where to do the pieces of this in your own instance (after you setup)

matyax (approved):

PR #674, opened 2024-08-12
chore(deps): resolve `fast-loops` to `1.1.4` to prevent `CVE-2024-39008` notifications

svennergr:
The `fast-loops` dependency has `CVE-2024-39008`. It is nested in @grafana/data -> react-use, etc. 
For now, we will just manually resolve to the fixed `1.1.4` version.

Also, the `jackspeak` resolution did not seem needed anymore? I could execute tests without a problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "tslib": "2.5.3"
   },
   "resolutions": {
-    "jackspeak": "2.1.1"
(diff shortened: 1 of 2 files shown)


matyax:
Just out of curiosity, what's the potential impact of this frontend CVE?

matyax (approved):

PR #682, opened 2024-08-12
Detected labels: Runtime datasource refactor

gtk-grafana:
Refactoring the detected_labels api call to use the run time datasource in order to prevent calling detected_labels on every route, and instead only when the query labels have changed, or when the labels route is viewed, or when we don't have a count for labels.

Also in this PR we remove the auto-navigation on change of any label or filter, this was dealt with on this PR as otherwise we get queries that are executed and immediately cancelled, only to be fired up again after the auto-route change.

Fixes https://github.com/grafana/explore-logs/issues/679
Fixes https://github.com/grafana/explore-logs/issues/692
Fixes https://github.com/grafana/explore-logs/issues/680
Fixes: https://github.com/grafana/explore-logs/issues/564

Notes for reviewer:
In this PR, we refactor out the labels and fields into 3 scenes, one parent, one for the labels, one of the label values.
Loading state was also broken in the value scenes due to creating a scene data node to select the individual frame, I've added a transformation so we retain the loading state of the query in the viz.

TODO: e2e tests, manual testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,8 @@
     "lint": "eslint --cache --ignore-path ./.gitignore --ext .js,.jsx,.ts,.tsx .",
     "lint:fix": "npm run lint -- --fix",
     "e2e": "playwright test -x --trace on",
+    "e2e:10x": "playwright test -x --trace on --repeat-each=10 --workers 1",
+    "e2e:flake": "yarn e2e:10x",
     "e2e:debug": "playwright test --debug -x --trace on /test",
     "e2e:codegen": "playwright install && playwright codegen",
     "generate-logs": "pushd generator; go mod download; go run . -url http://localhost:3100/loki/api/v1/push; popd",
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "tslib": "2.5.3"
   },
   "resolutions": {
-    "jackspeak": "2.1.1"
+    "**/fast-loops": "^1.1.4"
   },
   "packageManager": "yarn@1.22.21"
 }
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -101,7 +101,9 @@ function validateVariableNameForField(field: string, variableName: string) {
 function resolveVariableNameForField(field: string, scene: SceneObject) {
   const serviceScene = sceneGraph.getAncestor(scene, ServiceScene);
   // const indexedLabel = serviceScene.state.labels?.find((label) => label.label === field);
-  const indexedLabel = serviceScene.state.$detectedLabelsData.state.data?.series[0].fields?.find((label) => label.name === field);
+  const indexedLabel = serviceScene.state.$detectedLabelsData?.state.data?.series[0].fields?.find(
+    (label) => label.name === field
+  );
   return indexedLabel ? VAR_LABELS : VAR_FIELDS;
 }
 
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -84,9 +84,6 @@ export function replaceFilter(
     validateVariableNameForField(key, resolveVariableNameForField(key, scene)),
     scene
   );
-  if (!variable) {
-    return;
-  }
 
   variable.setState({
     filters: [
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -65,7 +65,6 @@ export function addToFilters(
     ];
   }
 
-  // Emit event
   scene.publishEvent(new AddFilterEvent(operator, key, value), true);
 
   variable.setState({
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { AdHocVariableFilter, DataFrame } from '@grafana/data';
+import { AdHocVariableFilter, BusEventBase, DataFrame } from '@grafana/data';
 import { SceneObjectState, SceneObjectBase, SceneComponentProps, SceneObject, sceneGraph } from '@grafana/scenes';
 import { VariableHide } from '@grafana/schema';
 import { USER_EVENTS_ACTIONS, USER_EVENTS_PAGES, reportAppInteraction } from 'services/analytics';
@@ -12,6 +12,13 @@ import { ServiceScene } from '../ServiceScene';
 export interface AddToFiltersButtonState extends SceneObjectState {
   frame: DataFrame;
   variableName: string;
+}
+
+export class AddFilterEvent extends BusEventBase {
+  constructor(public operator: FilterType, public key: string, public value: string) {
+    super();
+  }
+  public static type = 'add-filter';
 }
 
 /**
@@ -58,6 +65,9 @@ export function addToFilters(
       },
     ];
   }
+
+  // Emit event
+  scene.publishEvent(new AddFilterEvent(operator, key, value), true);
 
   variable.setState({
     filters,
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -1,13 +1,13 @@
 import React from 'react';
 
 import { AdHocVariableFilter, BusEventBase, DataFrame } from '@grafana/data';
-import { SceneObjectState, SceneObjectBase, SceneComponentProps, SceneObject, sceneGraph } from '@grafana/scenes';
+import { SceneComponentProps, SceneObject, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
 import { VariableHide } from '@grafana/schema';
-import { USER_EVENTS_ACTIONS, USER_EVENTS_PAGES, reportAppInteraction } from 'services/analytics';
+import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from 'services/analytics';
 import { getAdHocFiltersVariable, LEVEL_VARIABLE_VALUE, VAR_FIELDS, VAR_LABELS, VAR_LEVELS } from 'services/variables';
 import { FilterButton } from 'Components/FilterButton';
 import { FilterOp } from 'services/filters';
-import { ServiceScene } from '../ServiceScene';
+import { getDetectedLabelsFrame } from '../ServiceScene';
 
 export interface AddToFiltersButtonState extends SceneObjectState {
   frame: DataFrame;
@@ -34,7 +34,6 @@ export function addAdHocFilter(filter: AdHocVariableFilter, scene: SceneObject, 
   addToFilters(filter.key, filter.value, type, scene, variableName);
 }
 
-// @todo bubble up event
 export function addToFilters(
   key: string,
   value: string,
@@ -110,11 +109,7 @@ function validateVariableNameForField(field: string, variableName: string) {
 }
 
 function resolveVariableNameForField(field: string, scene: SceneObject) {
-  const serviceScene = sceneGraph.getAncestor(scene, ServiceScene);
-  // const indexedLabel = serviceScene.state.labels?.find((label) => label.label === field);
-  const indexedLabel = serviceScene.state.$detectedLabelsData?.state.data?.series[0].fields?.find(
-    (label) => label.name === field
-  );
+  const indexedLabel = getDetectedLabelsFrame(scene)?.fields?.find((label) => label.name === field);
   return indexedLabel ? VAR_LABELS : VAR_FIELDS;
 }
 
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -27,6 +27,7 @@ export function addAdHocFilter(filter: AdHocVariableFilter, scene: SceneObject, 
   addToFilters(filter.key, filter.value, type, scene, variableName);
 }
 
+// @todo bubble up event
 export function addToFilters(
   key: string,
   value: string,
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -100,7 +100,8 @@ function validateVariableNameForField(field: string, variableName: string) {
 
 function resolveVariableNameForField(field: string, scene: SceneObject) {
   const serviceScene = sceneGraph.getAncestor(scene, ServiceScene);
-  const indexedLabel = serviceScene.state.labels?.find((label) => label.label === field);
+  // const indexedLabel = serviceScene.state.labels?.find((label) => label.label === field);
+  const indexedLabel = serviceScene.state.$detectedLabelsData.state.data?.series[0].fields?.find((label) => label.name === field);
   return indexedLabel ? VAR_LABELS : VAR_FIELDS;
 }
 
--- a/src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
@@ -12,7 +12,7 @@ import {
 } from '@grafana/scenes';
 import { buildDataQuery, LokiQuery } from '../../../services/query';
 import { getSortByPreference } from '../../../services/store';
-import { DataQueryError, LoadingState, ReducerID } from '@grafana/data';
+import { DataQueryError, LoadingState } from '@grafana/data';
 import { LayoutSwitcher } from './LayoutSwitcher';
 import { getQueryRunner } from '../../../services/panel';
 import { ByFrameRepeater } from './ByFrameRepeater';
@@ -30,6 +30,7 @@ import { AddFilterEvent } from './AddToFiltersButton';
 import { navigateToDrilldownPage } from '../../../services/navigate';
 import { PageSlugs } from '../../../services/routing';
 import { ServiceScene } from '../ServiceScene';
+import { DEFAULT_SORT_BY } from '../../../services/sorting';
 
 export interface FieldValuesBreakdownSceneState extends SceneObjectState {
   body?: LayoutSwitcher | SceneReactObject;
@@ -134,7 +135,7 @@ export class FieldValuesBreakdownScene extends SceneObjectBase<FieldValuesBreakd
     const groupByVariable = getFieldGroupByVariable(this);
     const tagKey = String(groupByVariable.state.value);
 
-    const { sortBy, direction } = getSortByPreference('fields', ReducerID.stdDev, 'desc');
+    const { sortBy, direction } = getSortByPreference('fields', DEFAULT_SORT_BY, 'desc');
 
     const fieldsBreakdownScene = sceneGraph.getAncestor(this, FieldsBreakdownScene);
     const getFilter = () => fieldsBreakdownScene.state.search.state.filter ?? '';
--- a/src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
@@ -1,0 +1,137 @@
+import {
+  PanelBuilders,
+  SceneComponentProps,
+  SceneCSSGridLayout,
+  SceneFlexItem,
+  SceneFlexLayout,
+  sceneGraph,
+  SceneObjectBase,
+  SceneObjectState,
+  SceneReactObject,
+} from '@grafana/scenes';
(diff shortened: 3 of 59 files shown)


src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -100,7 +100,10 @@ function validateVariableNameForField(field: string, variableName: string) {
 
 function resolveVariableNameForField(field: string, scene: SceneObject) {
   const serviceScene = sceneGraph.getAncestor(scene, ServiceScene);
-  const indexedLabel = serviceScene.state.labels?.find((label) => label.label === field);
+  // const indexedLabel = serviceScene.state.labels?.find((label) => label.label === field);
+  const indexedLabel = serviceScene.state.$detectedLabelsData?.state.data?.series[0].fields?.find(
```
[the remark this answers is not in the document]
gtk-grafana: Fixed, added helper function `getDetectedLabelsFrame`

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -84,69 +84,81 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
     );
     this._subs.add(this.subscribeToEvent(SortCriteriaChanged, this.handleSortByChange));
 
-    const variable = this.getVariable();
-    const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
-
     // Need to update labels with current state
-    if (serviceScene.state.labels) {
-      this.updateLabels(serviceScene.state.labels);
+    if (serviceScene.state.$detectedLabelsData?.state.data?.series?.[0]) {
```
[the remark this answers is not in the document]
gtk-grafana: Added `getDetectedLabelsFrame`

src/services/datasource.ts
```diff
@@ -209,6 +217,59 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+        start: request.range.from.utc().toISOString(),
+        end: request.range.to.utc().toISOString(),
+      },
+      {
+        headers: {
+          'X-Query-Tags': `Source=${PLUGIN_ID}`,
+        },
+      }
+    );
+    detectedLabels.then((response) => {
```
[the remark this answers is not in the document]
gtk-grafana: Updated to use async/await, and added some try-catch blocks which return an error state in the panel data.

src/services/metadata.ts
```diff
@@ -18,19 +18,26 @@ export class MetadataService {
...
+
+    this.serviceSceneState.patternsCount = count;
+  }
+
+  public setLabelsCount(count: number) {
+    if(!this.serviceSceneState){
+      this.serviceSceneState = {}
+    }
+
+    this.serviceSceneState.labelsCount = count
```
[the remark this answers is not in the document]
gtk-grafana: Still?

matyax:
Just double checking. When I exclude, sometimes the value stays excluded, other times it's just gone. Is it because it might not be in the updated data?

https://github.com/user-attachments/assets/64bf89d6-2fdb-405f-a8fa-94f193778eca

matyax:
I excluded everything from a breakdown. Labels show empty panels, fields show no data.

https://github.com/user-attachments/assets/c41cbd0c-65f2-49f6-ab89-bb4de09de92a

Additionally I found this label with "no data".

imagen

package.json
```diff
@@ -11,6 +11,8 @@
     "lint": "eslint --cache --ignore-path ./.gitignore --ext .js,.jsx,.ts,.tsx .",
     "lint:fix": "npm run lint -- --fix",
     "e2e": "playwright test -x --trace on",
+    "e2e:10x": "playwright test -x --trace on --repeat-each=10 --workers 1",
```
matyax, line 14: 🆒

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -58,6 +65,9 @@ export function addToFilters(
     ];
   }
 
+  // Emit event
```
matyax, line 68:
```suggestion
```

src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
```diff
@@ -0,0 +1,231 @@
...
+      this.state.$data?.subscribeToState((newState) => {
+        this.onValuesDataQueryChange(newState, query);
+      })
+    );
+  }
+
+  private onValuesDataQueryChange(newState: SceneDataState, query: LokiQuery) {
+    if (newState.data?.state === LoadingState.Done) {
+      // No panels for the user to select, presumably because everything has been excluded
+      const event = this.state.lastFilterEvent;
```
matyax, line 75: I like the concept, but not the fact that we're storing event references and using that to infer behaviors. Wouldn't it be almost the same if we just redirect when `newState.data.series.length < 1`?

matyax: Like:

https://github.com/user-attachments/assets/a30a5406-3ca6-4c6f-b757-ec14b9d9398f

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -49,16 +39,26 @@ import { getFieldOptions } from '../../../services/filters';
...
-  body?: SceneObject;
+  body?: SceneReactObject | FieldsAggregatedBreakdownScene | FieldValuesBreakdownScene | SceneFlexLayout;
   search: BreakdownSearchScene;
   sort: SortByScene;
   value?: string;
   loading?: boolean;
   error?: string;
   blockingMessage?: string;
   changeFields?: (n: string[]) => void;
+  lastFilterEvent?: AddFilterEvent;
```
matyax, line 61: This is unused in this scene, right?

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -196,16 +213,27 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
...
-          : this.buildValuesLayout(newState);
+      // Otherwise update the body, but don't re-instantiate if it's already the right class
+      if (newState.value === ALL_VARIABLE_VALUE && this.state.body instanceof FieldValuesBreakdownScene) {
+        stateUpdate.body = new FieldsAggregatedBreakdownScene({});
+      } else if (newState.value !== ALL_VARIABLE_VALUE && this.state.body instanceof FieldsAggregatedBreakdownScene) {
+        stateUpdate.body = new FieldValuesBreakdownScene({});
+      } else if (
+        this.state.body === undefined ||
+        this.state.body instanceof SceneFlexLayout ||
+        this.state.body instanceof SceneReactObject
```
matyax, line 224:
```
        this.state.body instanceof SceneFlexLayout ||
        this.state.body instanceof SceneReactObject
```

I'm not sure I understand what this means.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -463,7 +353,14 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
               />
             )}
           </div>
-          <div className={styles.content}>{body && <body.Component model={body} />}</div>
+
+          {/* @todo why are the types like this? */}
```
matyax, line 357: We should know 😆

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -84,69 +85,82 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
+    const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
+    const groupByVariable = getLabelGroupByVariable(this);
+
     this.setState({
-      loading: true,
+      loading: serviceScene.state.$detectedLabelsData?.state.data?.state !== LoadingState.Done,
+      error: serviceScene.state.$detectedLabelsData?.state.data?.state === LoadingState.Error,
     });
 
+    //@todo audit
```
matyax, line 96:
```suggestion
```

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -195,160 +205,28 @@ export class LabelBreakdownScene extends SceneObjectBase<LabelBreakdownSceneStat
...
   public onChange = (value?: string) => {
     if (!value) {
       return;
     }
 
-    const variable = this.getVariable();
+    const variable = getLabelGroupByVariable(this);
     variable.changeValueTo(value);
 
-    const { sortBy, direction } = getSortByPreference('labels', ReducerID.stdDev, 'desc');
```
matyax, line 351: Great catch here.

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -0,0 +1,219 @@
...
+    });
+  }
+
+  private onValuesDataQueryChange(newState: SceneDataState) {
+    if (newState.data?.state === LoadingState.Done) {
+      // No panels for the user to select, presumably because everything has been excluded
+      const event = this.state.lastFilterEvent;
+
+      // @todo discuss: Do we want to let users exclude all labels? Or should we redirect when excluding the penultimate panel?
+      if (newState.data?.state === LoadingState.Done && event) {
```
matyax, line 89: Ditto comments in the sibling scene.

matyax (commented):
Made it halfway through. Will continue later.

I tested the changes and works pretty well! I commented the issues I found, namely an empty panel and sorting not working.

src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
gtk-grafana: Almost the same yes, but then if a user changes the variables up top, or happens to refresh the page and get no parsed results, they'll get an unexpected redirect. I wanted redirection to only happen on click of the last include or exclude button. Otherwise users should be presented with the "no fields, clear" UI instead of being moved around for reasons unknown to the user.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
gtk-grafana: I added a bit of explanation in a comment here

gtk-grafana:
> Just double checking. When I exclude, sometimes the value stays excluded, other times it's just gone. Is it because it might not be in the updated data?

Yup, parsed fields might not exist on the new dataframe.

> Sorting appears to be partially working. The panel names change, but the data doesn't. The sorting implementation relies on dataframe object references, so maybe there's something there. I can help if needed.

Oofdah. I'll take a look.

> I excluded everything from a breakdown. Labels show empty panels, fields show no data. Additionally I found this label with "no data".

This is annoying, but expected, the detected_labels like the patterns and volume APIs only take stream selectors, and we don't know if they all return no-data since the view is lazyloaded. Also is the current functionality on main

gtk-grafana:
@matyax the issue with sorting is the `SceneDataTransformer` in fields.ts
``` .setData(new SceneDataNode({ data: { ...data, series: [frame] } }))``` works as expected, but strips out all loading states. Would love to find some time tomorrow to debug

File: src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx (hunk 10 above)
Comment (gtk-grafana) [reply src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx]: This helped: https://www.executeprogram.com/courses/advanced-typescript/lessons/type-algebra, basically some of these objects had different definitions (at least one of the methods had differing access control modifiers) of sceneObject methods, so when attempting to define a type that is the union of the various body types, typescript reduced the definition to never. To get around this I made the definition of body a bit more verbose, and defined each union as the intersection of the object and SceneObject. Which appears to work as expected, and is probably better then a type assertion!

Might be nice to come back to this when we have time and update scenes core to sync the access control modifiers

**Comment by gtk-grafana** [comment]
> Something is broken when changing the service.

This is also a problem on main, I created a new issue to track: https://github.com/grafana/explore-logs/issues/704.

**Comment by gtk-grafana** [comment]
@svennergr the infinite loading state has been fixed

**Comment by matyax** [comment]
Resumed testing these changes. All the issues mentioned before are gone 👌 

One thing caught my attention, and I can't understand if somehow this makes sense. At first glance looks like it doesnt.

https://github.com/user-attachments/assets/4630be8a-1dcc-4d0e-a1f7-185767927be3

In this test I'm just applying a filter with `include`, which takes me back to the values view. But for a brief moment it shows fields which start displaying metric data, only to disappear around 5-8 seconds later, showing a completely different list.

**Comment by gtk-grafana** [comment]
@matyax, yeah that's because the fields are parsed from your last result of the logsPanelQuery, while the result of the logsPanelQuery query is loading with your new filter the panels are also loading. I think the detected_fields API is likely our only way around this, without forcing the user into a loading state and removing all the panels whenever a filter is added. With many datasets adding a filter doesn't have a big impact on the resulting parsed fields, but since it looks like you are picking a sparse field, in a HUGE dataset, sometimes there aren't many fields in common between the two queries.

I can repro in main on ops, you just have to click into the fields tab fast enough after the redirect, so I'm gonna suggest we open an issue for this and address in the detected_fields refactor?

File: tests/exploreServices.spec.ts [hunk 14]
```diff
@@ -15,10 +15,12 @@ test.describe('explore services page', () => {
...
     await explorePage.servicesSearch.click();
     await explorePage.servicesSearch.pressSequentially('mimir');
+
+    await page.getByTestId('data-testid Panel header mimir-ingester').first().scrollIntoViewIfNeeded()
     // service name should be in time series panel
-    await expect(page.getByTestId('data-testid Panel header mimir-ingester').nth(0)).toBeVisible();
+    expect(await page.getByTestId('data-testid Panel header mimir-ingester').nth(0)).toBeDefined()
     // service name should also be in logs panel, just not visible to the user
-    await expect(page.getByTestId('data-testid Panel header mimir-ingester').nth(1)).toBeVisible();
+    expect(await page.getByTestId('data-testid Panel header mimir-ingester').nth(1)).toBeDefined();
```
matyax, line 23: I guess defined is less flaky than visible?

src/services/variables.ts
```diff
@@ -19,15 +20,22 @@ export const VAR_LOGS_FORMAT = 'logsFormat';
...
 
+export function getLogsStreamSelector(options: {
+  labelExpressionToAdd?: string;
+  structuredMetadataToAdd?: string;
+  fieldExpressionToAdd?: string;
+}) {
+  const { labelExpressionToAdd = '', structuredMetadataToAdd = '', fieldExpressionToAdd = '' } = options;
+  return `{${VAR_LABELS_EXPR}${labelExpressionToAdd}} ${structuredMetadataToAdd} ${VAR_PATTERNS_EXPR} ${VAR_LOGS_FORMAT_EXPR} ${fieldExpressionToAdd} ${VAR_LEVELS_EXPR} ${VAR_FIELDS_EXPR} ${VAR_LINE_FILTER_EXPR}`;
+}
+
```
matyax, line 38: Nit.
```suggestion
export function getLogsStreamSelector(labelExpressionToAdd = '', structuredMetadataToAdd = '', fieldExpressionToAdd = '') {
  return `{${VAR_LABELS_EXPR}${labelExpressionToAdd}} ${structuredMetadataToAdd} ${VAR_PATTERNS_EXPR} ${VAR_LOGS_FORMAT_EXPR} ${fieldExpressionToAdd} ${VAR_LEVELS_EXPR} ${VAR_FIELDS_EXPR} ${VAR_LINE_FILTER_EXPR}`;
}

```

src/services/sorting.ts
```diff
@@ -4,6 +4,8 @@ import { getLabelValueFromDataFrame } from './levels';
 import { memoize } from 'lodash';
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from './analytics';
 
+export const DEFAULT_SORT_BY = 'changepoint';
```
matyax, line 7: <3

src/services/query.ts
```diff
@@ -58,7 +58,7 @@ const defaultQueryParams = {
 };
 
 export function renderLogQLStreamSelector(filters: AdHocVariableFilter[]) {
```
matyax, line 60: Nit, but this function no longer renders the stream selector, but a sequence of labels.

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -58,24 +57,33 @@ export interface ServiceSceneState extends SceneObjectState, ServiceSceneCustomS
...
+export function getDetectedLabelsFrame(sceneRef: SceneObject) {
+  const serviceScene = sceneGraph.getAncestor(sceneRef, ServiceScene);
+  return serviceScene.state.$detectedLabelsData?.state.data?.series?.[0];
+}
+
 export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
   protected _variableDependency = new VariableDependencyConfig(this, {
     variableNames: [VAR_DATASOURCE, VAR_LABELS, VAR_FIELDS, VAR_PATTERNS, VAR_LEVELS],
-    onReferencedVariableValueChanged: this.onReferencedVariableValueChanged.bind(this),
+    // onReferencedVariableValueChanged: this.onReferencedVariableValueChanged.bind(this),
```
matyax, line 75:
```suggestion
```

matyax (approved):
Fantastic work Galen 👏

tests/exploreServices.spec.ts
gtk-grafana: I saw some executions in ci where the volume was different then local and it wasn't rendered until scrolling to it, but it should be in the viewport now... I'll double check that this is needed

src/services/variables.ts
gtk-grafana: I dunno about this, there's no order to the props, I might have a fieldExpressionToAdd but no labelExpressionToAdd, I don't want to have to call the function with a bunch of empty strings in the function, i.e. getLogsStreamSelector('','',fieldExpression)

src/services/query.ts
gtk-grafana: How's `renderLogQLStreamLabels`?

tests/exploreServices.spec.ts
gtk-grafana: I changed it back to is visible, passed 10x locally, let's see aboot ci

src/services/query.ts
gtk-grafana: I couldn't bring myself to name it renderLogQLLabels, the `LL` hurt me inside

tests/exploreServices.spec.ts
gtk-grafana: I changed it back, seems to be passing? Could be that I changed it before adding the scrollTo and never changed it back. I'll re-run in CI a few times and if it flakes we can take another look

src/services/variables.ts
matyax: I see your point, but very quickly you can use the same argument to make every function receive and object and not have multiple arguments.

src/services/query.ts
gtk-grafana: Love it. Changed!

src/services/variables.ts
gtk-grafana: IMO if you have more then 2 optional params it should be an object

src/services/datasource.ts
```diff
@@ -204,13 +214,71 @@ class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+    const targets = request.targets.filter((target) => {
+      return target.resource === 'detected_labels';
     });
 
+    if (targets.length !== 1) {
+      throw new Error('Detected labels query can only have a single target!');
+    }
+
+    const targetsInterpolated = ds.interpolateVariablesInQueries(targets, request.scopedVars);
+    const interpolatedTarget = targetsInterpolated[0];
```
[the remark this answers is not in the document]
gtk-grafana: If targets is empty, it could also be empty, so the assertion above that targets.length must be 1, should be fine, until the implementation of `interpolateVariablesInQueries` changes in the loki datasource.

Is it worth adding another assertion that `targetsInterpolated` has length and throwing an error if it doesn't? otherwise we'll likely get an error in the network request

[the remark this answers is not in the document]
gtk-grafana: Added new error if the result is empty

src/services/variables.ts
matyax: I couldn't articulate it yesterday, but I was actually thinking about this:

```suggestion
export function getLogsStreamSelector({ labelExpressionToAdd = '', structuredMetadataToAdd = '', fieldExpressionToAdd = '' }) {
  return `{${VAR_LABELS_EXPR}${labelExpressionToAdd}} ${structuredMetadataToAdd} ${VAR_PATTERNS_EXPR} ${VAR_LOGS_FORMAT_EXPR} ${fieldExpressionToAdd} ${VAR_LEVELS_EXPR} ${VAR_FIELDS_EXPR} ${VAR_LINE_FILTER_EXPR}`;
}
```

Single deconstructed argument. `options` as an object was the source of discomfort.

Irrelevant, but a fun conversation to have IMO.

gtk-grafana: 🤦 yeah this is what we want, I'll update it next time I'm in here

PR #684, opened 2024-08-13
ServiceScenes: Detected level variable not clearing after changing service

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/683

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -25,6 +25,7 @@ import {
   getFieldsVariable,
   getLevelsVariable,
   getPatternsVariable,
+  getUrlParamNameForVariable,
   VAR_DATASOURCE,
   VAR_FIELDS,
   VAR_LABELS,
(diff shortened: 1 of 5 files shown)


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -104,7 +105,8 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     this.setState(stateUpdate);
 
     this.updatePatterns(this.state, getPatternsVariable(this));
-    this.syncFieldsWithUrl(getFieldsVariable(this));
+    this.resetVariablesIfNotInUrl(getFieldsVariable(this), 'var-fields');
+    this.resetVariablesIfNotInUrl(getLevelsVariable(this), 'var-levels');
```
svennergr, line 109: We should get these values from `variables.ts`

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -118,18 +120,19 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
   }
 
   /**
-   * @todo why do we need to manually sync fields, but nothing else?
-   * @param fieldsVariable
+   * @todo why do we need to manually sync fields and levels, but not other ad hoc variables?
```
svennergr, line 123: That is a very good question. Shouldn't we use something like `SceneObjectUrlSyncConfig` for this?

gtk-grafana: You'd think, but it doesn't work, the variable is still cached in the scene despite not being in the url params

gtk-grafana: Well actually let me double check, I forgot you can't just add keys, but implementation needs to be added in `getUrlState`, and `updateFromUrl`

gtk-grafana: Yeah, the url sync seems like it would only help if we were storing some state of the IndexScene in the url, but we're having trouble with the state of the $variable filters. I tried adding levels, fields, and even $variables to the dep array and was never able to get anything useful out of the url functions. 🤷

matyax (approved):
Tested and fixes the issue. Great work with the e2e tests!

PR #685, opened 2024-08-13
Breakdowns: Level, detected level, level_extracted, oh my!

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/676

The same level panel shows up in several different places in the UI. Let's remove them all except the one that's pushed to the top of the label breakdown.


matyax (approved):
🙌

svennergr:
For posterity: theoretically `detected_level`, `level` and `level_extracted` can all have different values. Still, I guess it's a good change going forward.

matyax:
Good point @svennergr . I think the main objective right now is to be consistent and minimize possible sources of confusion by, for example, having more than one `level` identifier.

We can come back to this in the future with something smarter if, for example, there's a chance of having `level` but not `detected_level` or something similar.

PR #703, opened 2024-08-20
Fields: include and exclude empty values

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/590

Updates the buttons on the field names view to allow users to include/exclude all logs that contain that field.

Base branch: https://github.com/grafana/explore-logs/pull/682
Should not be merged until parent branch has been merged!

Fields:
image

Labels:
image

Tooltips:
image

image

image
image
image


gtk-grafana:
Opened up for review after implementing the above, and reviewing UI with @zizzpudding.

matyax:
I'm seeing the following issue:

https://github.com/user-attachments/assets/84fd512e-c8c4-49fd-84b6-d7e42cb3afa9

src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
```diff
@@ -1,24 +1,137 @@
...
+    });
+
     return (
-      <Button variant="secondary" size="sm" onClick={model.onClick} aria-label={`Select ${model.useState().labelName}`}>
-        Select
-      </Button>
+      <>
+        {showFilterField === true && (
+          <FilterButton
+            isExcluded={existingFilter?.operator === '='}
```
matyax, line 36: May want to use `FilterOp`.

src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
```diff
@@ -1,24 +1,137 @@
...
+      if (logLinesWithLabelCount < logsPanelData.length) {
+        this.setState({
+          showFilterField: true,
+        });
+      } else {
+        this.setState({
+          showFilterField: false,
+        });
+      }
+    }
```
matyax, line 135: Nit:

```suggestion
    if (!labels || !logsPanelData) {
      this.setState({
        showFilterField: false,
      });
      return;
    }
    // iterate through all the labels on the log panel query result and count how many times this exists
    const logLinesWithLabelCount = labels.values.reduce((acc, labels) => {
      if (labels?.[this.state.labelName]) {
        acc++;
      }
      return acc;
    }, 0);

    const panel = sceneGraph.getAncestor(this, VizPanel);
    const percentage = ((logLinesWithLabelCount / logsPanelData.length) * 100).toLocaleString();
    const description = `${this.state.labelName} exists on ${percentage}% of ${logsPanelData.length} sampled log lines`;

    // Update the desc
    panel.setState({
      description,
    });

    if (logLinesWithLabelCount < logsPanelData.length) {
      this.setState({
        showFilterField: true,
      });
    }
```

src/services/query.ts
```diff
@@ -66,6 +66,9 @@ export function renderLogQLFieldFilters(filters: AdHocVariableFilter[]) {
 }
 
 function renderFilter(filter: AdHocVariableFilter) {
+  if (filter.value === '""') {
```
matyax, line 69: Would it break everything if the value is literally an empty string? I dislike seeing the "" in the field.

matyax (commented):
Few comments, but looks great.

src/services/query.ts
gtk-grafana: Probably, i'll take a look tho

src/services/variables.ts
```diff
@@ -27,6 +27,8 @@ export const ALL_VARIABLE_VALUE = '$__all';
 export const LEVEL_VARIABLE_VALUE = 'detected_level';
 export const PATTERNS_TEXT_FILTER = 'patternsFilter';
 
+export const EmptyVariableValue = '""';
```
matyax, line 30:
```suggestion
export const EMPTY_VARIABLE_VALUE = '""';
```

👉 👈

gtk-grafana: Done!

src/services/query.ts
gtk-grafana: So yeah, empty string means no filter in the variable, and is evaluated falsey in a bunch of places. This is the correct value for the query, I think we'd need more control over adhoc variables to get this to work

matyax: I imagined that was a possibility. Thanks for checking!

src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
```diff
@@ -1,24 +1,177 @@
...
+  getLabelsVariable,
+  getLevelsVariable,
+  LEVEL_VARIABLE_VALUE,
+  SERVICE_NAME,
+} from '../../../services/variables';
+import { AdHocVariableFilter, Field, Labels, LoadingState } from '@grafana/data';
+import { FilterOp } from '../../../services/filters';
 
-interface SelectLabelActionSceneState extends SceneObjectState {
+interface SelectFieldActionSceneState extends SceneObjectState {
```
matyax, line 27:
```suggestion
interface SelectLabelActionSceneState extends SceneObjectState {
```

src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
```diff
@@ -1,24 +1,177 @@
...
+    const variable = model.getVariable();
+    const variableName = variable.useState().name;
+    const existingFilter = model.getExistingFilter(variable);
+
     return (
-      <Button variant="secondary" size="sm" onClick={model.onClick} aria-label={`Select ${model.useState().labelName}`}>
-        Select
-      </Button>
+      <>
+        {variable && showFilterField === true && (
```
matyax, line 48: If variable could not be defined, then `const variableName = variable.useState().name;` (line 43) should throw. So either one or the other.

src/services/query.ts
```diff
@@ -95,6 +95,9 @@ export function renderLogQLFieldFilters(filters: AdHocVariableFilter[]) {
 }
 
 function renderFilter(filter: AdHocVariableFilter) {
+  if (filter.value === EMPTY_VARIABLE_VALUE) {
```
matyax, line 98: Silly, but took me a minute to understand this. Maybe adding a comment would be nice.

matyax (commented):
Left a few minor comments but looks ready do go 👌

matyax (approved):
Feel free to follow up on the comments and merge 🫡

src/services/query.ts
gtk-grafana: Done!

PR #712, opened 2024-08-21
chore: upgrade compose loki to enable aggregated metrics

trevorwhitney:
Upgrades the loki and loki-config used in both compose files to be one that has support for aggregated metrics.


gtk-grafana (approved):
Thank you!

PR #713, opened 2024-08-21
ServiceSelection: add support for aggregated metric

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/711

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "@grafana/data": "^11.1.1",
     "@grafana/lezer-logql": "^0.2.6",
     "@grafana/runtime": "^11.1.1",
-    "@grafana/scenes": "5.7.2",
+    "@grafana/scenes": "5.9.0",
     "@grafana/ui": "^11.1.1",
     "@lezer/common": "^1.2.1",
     "@lezer/lr": "^1.4.1",
--- a/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
+++ b/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
@@ -5,13 +5,14 @@ import { DashboardCursorSync, DataFrame, GrafanaTheme2, LoadingState, TimeRange,
 import {
   behaviors,
   PanelBuilders,
+  QueryRunnerState,
   SceneComponentProps,
   SceneCSSGridItem,
   SceneCSSGridLayout,
-  SceneDataProvider,
   sceneGraph,
   SceneObjectBase,
   SceneObjectState,
+  SceneQueryRunner,
   SceneVariableSet,
   VizPanel,
 } from '@grafana/scenes';
@@ -37,7 +38,7 @@ import {
 import { selectService, SelectServiceButton } from './SelectServiceButton';
 import { buildDataQuery, buildResourceQuery } from 'services/query';
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from 'services/analytics';
-import { getQueryRunner, setLevelColorOverrides } from 'services/panel';
+import { getQueryRunner, getSceneQueryRunner, setLevelColorOverrides } from 'services/panel';
 import { ConfigureVolumeError } from './ConfigureVolumeError';
 import { NoVolumeError } from './NoVolumeError';
 import { getLabelsFromSeries, toggleLevelVisibility } from 'services/levels';
@@ -46,6 +47,8 @@ import { CustomConstantVariable } from '../../services/CustomConstantVariable';
 import { areArraysEqual } from '../../services/comparison';
 
 export const SERVICE_NAME = 'service_name';
+// Don't export AGGREGATED_SERVICE_NAME, we want to rename things so the rest of the application is agnostic to how we got the services
+const AGGREGATED_SERVICE_NAME = '__aggregated_metric__';
 
 interface ServiceSelectionSceneState extends SceneObjectState {
   // The body of the component
@@ -53,15 +56,13 @@ interface ServiceSelectionSceneState extends SceneObjectState {
   // Show logs of a certain level for a given service
   serviceLevel: Map<string, string[]>;
   // Logs volume API response as dataframe with SceneQueryRunner
-  $data: SceneDataProvider;
+  $data: SceneQueryRunner;
+  $labels: SceneQueryRunner;
+  serviceName: typeof SERVICE_NAME | typeof AGGREGATED_SERVICE_NAME;
 }
 
 function getMetricExpression(service: string) {
   return `sum by (${LEVEL_VARIABLE_VALUE}) (count_over_time({${SERVICE_NAME}=\`${service}\`} | drop __error__ [$__auto]))`;
-}
-
-function getLogExpression(service: string, levelFilter: string) {
-  return `{${SERVICE_NAME}=\`${service}\`}${levelFilter}`;
 }
 
 export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionSceneState> {
@@ -79,11 +80,17 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
           }),
         ],
       }),
-      $data: getQueryRunner([buildResourceQuery(`{${SERVICE_NAME}=~\`.*${VAR_SERVICE_EXPR}.*\`}`, 'volume')]),
+      $data: getSceneQueryRunner({
+        queries: [buildResourceQuery(`{${SERVICE_NAME}=~\`.*${VAR_SERVICE_EXPR}.*\`}`, 'volume')],
+        runQueriesMode: 'manual',
+      }),
+      $labels: getSceneQueryRunner({ queries: [buildResourceQuery('', 'labels')], runQueriesMode: 'manual' }),
       serviceLevel: new Map<string, string[]>(),
+      serviceName: SERVICE_NAME,
       ...state,
     });
 
+    this.state.$labels.activate();
     this.addActivationHandler(this.onActivate.bind(this));
   }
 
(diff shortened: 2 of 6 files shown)


gtk-grafana:
Concerns/questions:

What if I query the last 30 days, but only have aggregated metric for past 15? In the current implementation we'll just check that the label exists in the time range, and the user will only see the most recent 15 days in the logs/time series. 

Is it worth feature flagging this feature for now? I can envision a more robust solution where loki makes the determination which label to use automatically if it's available for the full duration of a query, which would enable split queries over a long duration to use the aggregated metric for the splits that have it, and the regular query for those that don't, right now the frontend doesn't have the information to properly determine what time-range should use what selector

trevorwhitney:
@gtk-grafana what do we split by, a day right? maybe we need to check each split to know if there are aggregated metrics or not? I think it's probably safe to assume it is or isn't there for a whole split (which may lose some data), but you're right we probably want to guard against losing a week.

gtk-grafana:
We split by the default duration defined in Loki (1 day), unless the query is greater then 4 hours and less then 26 hours, then we split by 2 hours.

@matyax @svennergr I'd be interested to hear your thoughts, I guess it's possible to run a bunch of labels queries and manually determine the split duration for all queries, but that seems gross.

matyax:
I fear that if we have to run "probing" queries before sending the actual metric query to display we will lose the advantage of the aggregated metrics because of RTT + query execution times.

trevorwhitney:
> I fear that if we have to run "probing" queries before sending the actual metric query to display we will lose the advantage of the aggregated metrics because of RTT + query execution times.

I'm wondering if a `/services` endpoint in Loki could replace the `/volume` call, and include information on how to best construct the queries, to avoid the probing queries. I'm going to experiment and see if I can come up with something.

trevorwhitney:
So, I changed my mind. I think a better way to handle probing queries, and it's to put it behind a feature flag that could maybe eventually be a config (agg_metrics: on/off). If we limit the look back range to 2 weeks for GA it will work for our cloud customers, and if we have this toggleable behind a feature flag/config an OSS user can just wait a few weeks after deploying Loki 3.2 and turning agg metrics on to enable it

package.json
```diff
@@ -85,7 +85,7 @@
     "@grafana/data": "^11.1.1",
     "@grafana/lezer-logql": "^0.2.6",
     "@grafana/runtime": "^11.1.1",
-    "@grafana/scenes": "5.7.2",
+    "@grafana/scenes": "5.9.0",
```
trevorwhitney, line 88: did we bump this to get the ability to run scene queries manually? I don't think we need that anymore since we don't need to wait on labels anymore, but maybe we still want to upgrade?

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -29,23 +37,35 @@ import { getFavoriteServicesFromStorage } from 'services/store';
...
-
+import { config, getAppEvents } from '@grafana/runtime';
+import { VariableHide } from '@grafana/schema';
+import { PLUGIN_ID } from '../../services/routing';
+
+// @ts-expect-error
+const aggregatedMetricsEnabled: boolean | undefined = config.featureToggles.exploreLogsAggregatedMetrics;
+// Don't export AGGREGATED_SERVICE_NAME, we want to rename things so the rest of the application is agnostic to how we got the services
+const AGGREGATED_SERVICE_NAME = '__aggregated_metric__';
+const AGGREGATED_METRIC_START_DATE = dateTime('2024-08-30', 'YYYY-MM-DD');
```
trevorwhitney, line 67: eventually we may want this to be configurable (ie. OSS users), but I'm totally fine with this for now

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -71,20 +85,49 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
...
+                label: AGGREGATED_SERVICE_NAME,
+              },
+            ],
+          }),
         ],
       }),
-      $data: getQueryRunner([buildResourceQuery(`{${SERVICE_NAME}=~\`.*${VAR_SERVICE_EXPR}.*\`}`, 'volume')]),
+      $data: getSceneQueryRunner({
+        queries: [buildResourceQuery(`{${SERVICE_NAME_EXPR}=~\`.*${VAR_SERVICE_EXPR}.*\`}`, 'volume')],
+        runQueriesMode: 'manual',
```
trevorwhitney, line 118: are we still waiting on the labels call here? I don't think we need to now that we a) have the feature flag, and b) aren't using `__aggregated_metrics__` in the volume call.

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -71,20 +85,49 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
...
+      }),
+      $labels: aggregatedMetricsEnabled
```
trevorwhitney, line 120: what are we using labels for?

trevorwhitney (commented):
looking good! curious what the `labels` call is for now that we've gone back to the regular `/volume` call using `service_name=~".+"`? I'm fine leaving the resource type and logic to run a labels query (as we need it for `__stream_shard__` and I'm sure for other things in the future), but I don't think we need to add the state to the ServiceSelection scene unless I'm missing something?

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
gtk-grafana: To let us know if the `__aggregated_metrics__` label exists for that time range, just because the feature flag is enabled doesn't mean we have aggregated metrics for every service for every loki datasource? Or does it?

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
gtk-grafana: Agreed, @svennergr is there a way we can configure this in deployment tools with an env variable or something so OSS users can set their own date?

package.json
gtk-grafana: I think we still need to run the service queries manually so we can update the `SERVICE_NAME_VAR` before the query runs if the time range is too long, otherwise we'd change the variable after the initial query was already run (and then immediately cancel the request)

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
gtk-grafana: Cleaned up now in the latest commit, no more labels!

src/Components/IndexScene/ToolbarScene.tsx
```diff
@@ -0,0 +1,144 @@
...
+      disabled: boolean;
+    };
+  };
+}
+export class ToolbarScene extends SceneObjectBase<ToolbarSceneState> {
+  constructor(state: Partial<ToolbarSceneState>) {
+    const overrideFalsy =
+      localStorage.getItem(AGGREGATED_METRICS_USER_OVERRIDE_LOCALSTORAGE_KEY) === 'false' ? false : undefined;
+    const userOverride =
+      localStorage.getItem(AGGREGATED_METRICS_USER_OVERRIDE_LOCALSTORAGE_KEY) === 'true' ? true : overrideFalsy;
```
svennergr, line 27: Not sure I understand this. What stops us from using 
```
const userOverride = localStorage.getItem(AGGREGATED_METRICS_USER_OVERRIDE_LOCALSTORAGE_KEY) === 'true' ? true : false;
```

src/Components/IndexScene/ToolbarScene.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    if (options.aggregatedMetrics) {
+      return (
+        <Dropdown overlay={renderPopover} placement="bottom" onVisibleChange={model.onToggleOpen}>
+          <ToolbarButton icon="cog" variant="canvas" isOpen={isOpen} />
+        </Dropdown>
+      );
+    }
+
+    // Empty
```
svennergr, line 114:
```suggestion
```

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -108,6 +136,96 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
...
+      },
+    });
+  }
+
+  private runServiceQueries() {
+    const toolbar = this.getQueryOptionsToolbar();
+    const toolbarEnabled =
+      !toolbar?.state.options.aggregatedMetrics.disabled && toolbar?.state.options.aggregatedMetrics.active;
+
+    if ((!this.isTimeRangeTooEarlyForAggMetrics() || !aggregatedMetricsEnabled) && toolbarEnabled) {
```
svennergr, line 221: I know naming this is not easy, but maybe something like `aggregatedMetricsActive` instead of `toolbarEnabled`?

```suggestion
    const aggregatedMetricsActive =
      !toolbar?.state.options.aggregatedMetrics.disabled && toolbar?.state.options.aggregatedMetrics.active;

    if ((!this.isTimeRangeTooEarlyForAggMetrics() || !aggregatedMetricsEnabled) && aggregatedMetricsActive) {
```

tests/exploreServices.spec.ts
```diff
@@ -144,9 +146,15 @@ test.describe('explore services page', () => {
         await route.fulfill({ json: {} });
       });
 
+      await page.route('**/resources/labels*', async (route) => {
+        labelsQueryCount++;
```
svennergr, line 150: Is this something we should assert on?

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -170,6 +288,14 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
     }
   }
 
+  private getLogExpression(service: string, levelFilter: string) {
+    return `{${SERVICE_NAME_EXPR}=\`${service}\`}${levelFilter}`;
+  }
```
svennergr, line 293: The logs query must not use `__aggregated_metric__`. It's really just for metrics, not for logs.

tests/exploreServices.spec.ts
gtk-grafana: Not anymore, we removed this call. Fixed

src/Components/IndexScene/ToolbarScene.tsx
gtk-grafana: in local storage we can have 'false' || 'true' || null, but I agree this was written confusedly. I cleaned it up a bit and everything appears to still be working. Make sure to test queries that start before aug 30, and with empty local storage state.

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
gtk-grafana: its so fast tho

gtk-grafana: Fixed

src/Components/IndexScene/ToolbarScene.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+function getStyles(theme: GrafanaTheme2) {
+  return {
+    popover: css({
+      display: 'flex',
+      padding: theme.spacing(2),
+      flexDirection: 'column',
+      background: theme.colors.background.primary,
+      boxShadow: theme.shadows.z3,
+      borderRadius: theme.shape.borderRadius(),
```
matyax, line 123: Small deprecation here.

gtk-grafana: good catch, fixed!

matyax (approved):
Let's go!

PR #715, opened 2024-08-22
Shard query splitting: Split queries by stream shards

### LINKED ISSUE #690: Query splitting: split by shard in Explore Logs

matyax:
# Query splitting by stream shards.

Query splitting was introduced in Loki to optimize querying for long intervals and high volume of data, dividing a big request into smaller sub-requests, combining and displaying the results as they arrive.

This approach, inspired by the time-based query splitting, takes advantage of the __stream_shard__ internal label, representing how data is spread into different sources that can be queried individually.

The main entry point of this module is runShardSplitQuery(), which prepares the query for execution and passes it to splitQueriesByStreamShard() to begin the querying loop.

splitQueriesByStreamShard() has the following structure:
- Creates and returns an Observable to which the UI will subscribe
- Requests the __stream_shard__ values of the selected service:
  - If there are no shard values, it falls back to the standard querying approach of the data source in runNonSplitRequest()
  - If there are shards:
    - It groups the shard requests in an array of arrays of shard numbers in groupShardRequests()
    - It begins the querying loop with runNextRequest()
- runNextRequest() will send a query using the nth (cycle) shard group, and has the following internal structure:
  - adjustTargetsFromResponseState() will filter log queries targets that already received the requested maxLines
  - interpolateShardingSelector() will update the stream selector with the current shard numbers
  - After query execution:
    - If the response is successful:
      - It will add new data to the response with combineResponses()
      - nextRequest() will use the current cycle and the total groups to determine the next request or complete execution with done()
    - If the response is unsuccessful:
      - If there are retry attempts, it will retry the current cycle, or else continue with the next cycle
      - If the returned error is Maximum series reached, it will not retry
- Once all request groups have been executed, it will be done()

## Key features
- Shards retrieved by a `/values` call including the selected service.
- Starts with low volume shards, intertwined with higher volume shards.
- Groups shards based on the current time interval and available shards. 
- Sends queries grouping shards in a selector `__stream_shard__=~"shardn1|shardn2|...|shardn"`.
- For short intervals (< 24 hours) groups shards in queries. For longer intervals, queries individual shards.
- Recombines responses without asuming any particular order, including overlapping values.
- Retries failed requests up to 3 times, with exponential backoff (1500 ms * 2^failed attempts of waiting time) Does not retry if the response is "maximum series reached".

> [!IMPORTANT]  
> Requires the `exploreLogsShardSplitting` feature flag enabled

Closes #690

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,12 +49,6 @@ jobs:
             node_modules
           key: ${{ steps.restore-npm-cache.outputs.cache-primary-key }}
 
-      - name: Check types
-        run: yarn typecheck
-      - name: Lint
-        run: yarn lint
-      - name: Unit tests
-        run: yarn test:ci
       - name: Build frontend
         run: yarn build
       - name: Start grafana docker
@@ -63,9 +57,6 @@ jobs:
         run: npx playwright install chromium --with-deps
       - name: Wait for docker
         run: sleep 60
-      - name: Run e2e tests
-        id: run-tests
-        run: yarn e2e
       - name: Upload Playwright artifacts
         uses: actions/upload-artifact@v4
         if: ${{ (always() && steps.run-tests.outcome == 'success') || (failure() && steps.run-tests.outcome == 'failure') }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,12 @@ jobs:
             node_modules
           key: ${{ steps.restore-npm-cache.outputs.cache-primary-key }}
 
+      - name: Check types
+        run: yarn typecheck
+      - name: Lint
+        run: yarn lint
+      - name: Unit tests
+        run: yarn test:ci
       - name: Build frontend
         run: yarn build
       - name: Start grafana docker
@@ -57,6 +63,9 @@ jobs:
         run: npx playwright install chromium --with-deps
       - name: Wait for docker
         run: sleep 60
+      - name: Run e2e tests
+        id: run-tests
+        run: yarn e2e
       - name: Upload Playwright artifacts
         uses: actions/upload-artifact@v4
         if: ${{ (always() && steps.run-tests.outcome == 'success') || (failure() && steps.run-tests.outcome == 'failure') }}
--- a/project-words.txt
+++ b/project-words.txt
@@ -482,6 +482,6 @@ frametype
 Nanos
 nanos
 subquery
-Subsciption
+Subscription
 uuidv
-wihtout
+without
--- a/project-words.txt
+++ b/project-words.txt
@@ -476,3 +476,12 @@ yass
 noparser
 whoopsie
 myproject
+becasue
+decbytes
+frametype
+Nanos
+nanos
+subquery
+Subsciption
+uuidv
+wihtout
--- a/project-words.txt
+++ b/project-words.txt
@@ -476,7 +476,6 @@ yass
 noparser
 whoopsie
 myproject
-becasue
 decbytes
 frametype
 Nanos
--- a/src/services/combineResponses.test.ts
+++ b/src/services/combineResponses.test.ts
@@ -1035,6 +1035,20 @@ describe('mergeFrames', () => {
       ],
     });
   });
+
+  it('merging exactly the same data produces the same data', () => {
+    const { logFrameA } = getMockFrames();
+
+    const responseA: DataQueryResponse = {
+      data: [logFrameA],
+    };
+    const responseB: DataQueryResponse = {
+      data: [logFrameA],
+    };
+    expect(combineResponses(responseA, responseB)).toEqual({
+      data: [logFrameA],
+    });
+  });
 });
 
 export function getMockFrames() {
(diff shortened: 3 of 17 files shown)


gtk-grafana:
should this be in draft for now?

matyax:
100%. Just wanted the build for publishing.

matyax:
@grafana/observability-logs If you want to start the review while I work on the coverage, that'd be great. I don't plan to add any significant change to the code for now.

matyax:
Coverage added 😤

.github/workflows/ci.yml
```diff
@@ -63,9 +57,6 @@ jobs:
         run: npx playwright install chromium --with-deps
       - name: Wait for docker
         run: sleep 60
-      - name: Run e2e tests
```
gtk-grafana, line 66: why are we removing the e2e tests in ci?

project-words.txt
```diff
@@ -473,3 +473,12 @@ REFID
 unroute
 myown
 yass
+becasue
```
gtk-grafana, line 476: Can we fix the spelling errors like `becasue: because`, `Subsciption: subscription`, and `wihtout: without`? The rest are all fine

.github/workflows/ci.yml
matyax: We removed this originally to publish WIP. Added back.

project-words.txt
matyax: Fixed.

src/services/combineResponses.test.ts
```diff
@@ -0,0 +1,1403 @@
+import { DataFrame, DataFrameType, DataQueryResponse, Field, FieldType, QueryResultMetaStat } from '@grafana/data';
```
matyax, line 1: See https://github.com/grafana/grafana/blob/main/packages/grafana-o11y-ds-frontend/src/combineResponses.test.ts

src/services/combineResponses.ts
```diff
@@ -0,0 +1,281 @@
+import {
```
matyax, line 1: See https://github.com/grafana/grafana/blob/main/packages/grafana-o11y-ds-frontend/src/combineResponses.ts

src/services/shardQuerySplitting.test.ts
```diff
@@ -0,0 +1,194 @@
+import { of } from 'rxjs';
```
matyax, line 1: See https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/querySplitting.test.ts

src/services/shardQuerySplitting.ts
```diff
@@ -0,0 +1,264 @@
+import { Observable, Subscriber, Subscription } from 'rxjs';
```
matyax, line 1: See: https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/querySplitting.ts

gtk-grafana:
Can we prevent "no data" from showing up (and keep the panel in the loading state) when it's streaming but the shards we've queried don't have data?
image

matyax:
Definitely. What I wanted to try was to create the initial frame with fields with empty values, and see if that gets rid of that behavior. I did a quick test, but it will require an update to the merging code and some extra complexity, so I'd like to follow that up independently. And/or check the viz to see what is it that triggers the display of No Data.

gtk-grafana:
From what I remember the streaming (LoadingState) state expects dataframes, I think setting the initial state as LoadingState.Loading fixes 90% of this without any other refactor

matyax:
Fantastic, I'll give it a try. The reason why we used streaming with splitting was so that the panel updates the viz as new data appears, but we can start with loading and change to streaming easily.

gtk-grafana:
Yeah I think it should be "loading" state, until we get a non-empty response. We shouldn't update the "series" prop until we get a non-empty response, or the request is done.

But easier said then done :P if it's too much work we can punt on it as this is under a ff, but this is the only real issue I'm seeing right now in my testing

gtk-grafana:
@matyax I remember you mentioning that we wanted to exclude avg_over_time queries from being sharded, has that been addressed yet in this PR?

src/services/datasource.ts
```diff
@@ -128,11 +130,14 @@ export class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+    ds: DataSourceWithBackend<LokiQuery>,
     subscriber: Subscriber<DataQueryResponse>
   ) {
-    // query the datasource and return either observable or promise
-    const dsResponse = ds.query(request);
+    // @ts-expect-error
+    const shardingEnabled = config.featureToggles.exploreLogsShardSplitting;
+
+    // Query the datasource and return either observable or promise
+    const dsResponse = isLogsRequest(request) || !shardingEnabled ? ds.query(request) : runShardSplitQuery(ds, request);
```
gtk-grafana, line 140: Exclude requests that contain `avg_over_time` expressions?

matyax: Oh yes, definitely!

matyax: Done.

PR #717, opened 2024-08-23
Extract parser: update heuristic for mixed logs

matyax:
The first log of the sample can be anything, and the fact that we're only checking one felt awkward.

Updated the implementation to:
- Check all the lines
- Use JSON.parse() instead of checking the first character, like we do in [LogRowMessage](https://github.com/grafana/grafana/blob/main/public/app/features/logs/components/LogRowMessage.tsx#L72).
- Add both parsers when the type is mixed and not only json

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -471,3 +471,4 @@ svgs
 Dont
 REFID
 unroute
+myown
--- a/src/services/fields.ts
+++ b/src/services/fields.ts
@@ -18,8 +18,10 @@ export type DetectedLabelsResponse = {
   detectedLabels: DetectedLabel[];
 };
 
+type ExtractedFieldsType = 'logfmt' | 'json' | 'mixed';
+
 interface ExtractedFields {
-  type: 'logfmt' | 'json';
+  type: ExtractedFieldsType;
   fields: string[];
 }
 
@@ -30,7 +32,7 @@ export function updateParserFromDataFrame(frame: DataFrame, sceneRef: SceneObjec
   let newType;
   if (!res.type) {
     newType = '';
-  } else if (res.type === 'json') {
+  } else if (res.type === 'mixed') {
     newType = `| json  | logfmt | drop __error__, __error_details__`;
   } else {
     newType = ` | ${res.type}`;
@@ -55,8 +57,23 @@ export function extractParserAndFieldsFromDataFrame(data: DataFrame) {
     }, {}) ?? {}
   );
 
+  const types: ExtractedFieldsType[] = [];
+
   const linesField = data.fields.find((f) => f.name === 'Line' || f.name === 'body');
-  result.type = linesField?.values[0]?.[0] === '{' ? 'json' : 'logfmt';
+  linesField?.values.forEach((value: string) => {
+    if (types.length === 2) {
+      return;
+    }
+    try {
+      if (JSON.parse(value) && !types.includes('json')) {
+        types.push('json');
+      }
+    } catch (e) {
+      types.push('logfmt');
+    }
+  });
+
+  result.type = types.length === 1 ? types[0] : 'mixed';
 
   return result;
 }
--- a/src/services/fields.ts
+++ b/src/services/fields.ts
@@ -64,10 +64,7 @@ export function extractParserAndFieldsFromDataFrame(data: DataFrame) {
     return result;
   }
 
-  for (let i = 0; i < linesField.values.length; i++) {
-    if (types.length === 2) {
-      break;
-    }
+  for (let i = 0; i < linesField.values.length && types.length < 2; i++) {
     const line = linesField.values[i].trim();
     if (line.startsWith('{') && line.endsWith('}')) {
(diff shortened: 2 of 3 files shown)


matyax:
Follow up from the chat we had yesterday. BTW, I don't mind going back to checking the first character (after trimming) like we were doing, and we can also apply both parsers and drop error on JSON (I don't know the original reason for that).

What felt non-negotiable was only checking the first line.

src/services/fields.ts
```diff
@@ -53,8 +55,23 @@ export function extractParserAndFieldsFromDataFrame(data: DataFrame) {
...
+  const types: ExtractedFieldsType[] = [];
+
   const linesField = data.fields.find((f) => f.name === 'Line' || f.name === 'body');
-  result.type = linesField?.values[0]?.[0] === '{' ? 'json' : 'logfmt';
+  linesField?.values.forEach((value: string) => {
+    if (types.length === 2) {
+      return;
+    }
+    try {
+      if (JSON.parse(value) && !types.includes('json')) {
```
matyax, line 66: 1000 logs:

- JSON.parse 8ms
- trim().startsWith({) 0.3ms

gtk-grafana:
To recap our discussion in slack: This is a harder problem then it looks, if the result is mostly one format or the other, but occasionally mixed, we risk firing extra requests, or request looping despite any checking we do on the client.

We can always choose to use both parsers, which reduces the number of requests, but increases the cost of every query.

I know @cyriltovena gave us the ok in the previous version of the app to use both parsers, but maybe it's worth having the conversation again? 
CC @trevorwhitney 

My gut says that the cost of running unnecessary queries is higher then using two parsers all the time, but that has performance implications for users that have good data that is 100% one format.

Straw person: We use both parsers, and if users need to squeeze out more performance we add a config to the loki datasource to assert if the datasource is json, logfmt, or mixed?

Crazy idea: Can we have a new "auto" parser?

src/services/fields.ts
gtk-grafana: We can check the first (and last) char before attempting the parse, but that only helps for mixed mode, or users with logfmt. 100% json would see the worse performance, and huge JSON lines will take longer to parse

matyax: I don't think parsing is required, we can just check the start and end chars (great idea). Main goal: not just check [0][0]

matyax:
Auto parser would be amazing.

gtk-grafana:
Should be a win for logQL users as well

src/services/fields.ts
matyax: Changed the approach and removed parsing. Tested it with actual mixed logs and works as expected.

Perf: 1000 logs 0.10000002384185791 ms

matyax:
What do you think? Should we go with this improvement and resume the conversation of smarter alternatives later?

src/services/fields.ts
gtk-grafana: and that won't change (enough to matter) as the size of the json increases, opposed to parsing

project-words.txt
```diff
@@ -471,3 +471,4 @@ svgs
 Dont
 REFID
 unroute
+myown
```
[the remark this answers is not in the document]
gtk-grafana: values: ['{"jsonLabel": "jsonValue"}', 'level=error myown=summer'],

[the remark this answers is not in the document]
gtk-grafana: you could tell the spellcheck to ignore test files 🤷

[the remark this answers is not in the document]
matyax: Yeah, just wandering what good serves that string from the test file. We should skip tests.

src/services/fields.ts
```diff
@@ -53,8 +55,24 @@ export function extractParserAndFieldsFromDataFrame(data: DataFrame) {
     }, {}) ?? {}
   );
 
+  const types: ExtractedFieldsType[] = [];
+
   const linesField = data.fields.find((f) => f.name === 'Line' || f.name === 'body');
-  result.type = linesField?.values[0]?.[0] === '{' ? 'json' : 'logfmt';
+  linesField?.values.forEach((value: string) => {
```
gtk-grafana, line 61: lets use a normal for loop or some or something that will stop executing when we find a non-json log line

gtk-grafana: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#description

> There is no way to stop or break a forEach() loop other than by throwing an exception. If you need such behavior, the forEach() method is the wrong tool.

gtk-grafana:
This is a good change, but I'll want some more time to test on Monday, worried about re-introducing the request looping bug

src/services/fields.ts
matyax: Sure. No need to be suboptimal.

gtk-grafana (approved):
Looks good, and is working well locally. I like how this only executes another query for mixed json results! Great work Matias

PR #718, opened 2024-08-23
Filters: add support for filters with multiple values

matyax:
Aimed to fix https://github.com/grafana/explore-logs/issues/691 , this PR adds support for all fields and labels with multiple values.

Fields demo:

https://github.com/user-attachments/assets/984f4d25-aa11-409b-9e4e-19a708bdf9f1

Labels demo:

https://github.com/user-attachments/assets/cfd97a70-a94e-4901-ac1f-15269ec0fc86

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -43,7 +43,7 @@ import { getDrilldownSlug, PageSlugs } from '../../services/routing';
 import { ServiceSelectionScene } from '../ServiceSelectionScene/ServiceSelectionScene';
 import { LoadingPlaceholder } from '@grafana/ui';
 import { locationService } from '@grafana/runtime';
-import { renderLogQLFieldFilters, joinFilters, renderPatternFilters } from 'services/query';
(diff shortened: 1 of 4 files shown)


matyax:
I did this all with fields, but didn't check labels. Looking now 🧐

matyax:
Added support for labels too.

gtk-grafana:
This works, but the value breakdown UIs don't support including multiple yet.

matyax:
Oh absolutely. But you can get there by starting with exclusions.

gtk-grafana:
It's a good way to keep the size of the PR down. I'm fine with merging this as is and following up in the UI (which I think will be a bigger PR)

matyax:
Agree. If you would like to follow that up, since you recently refactored it, go for it. Otherwise I'll follow this up. Thanks!

gtk-grafana:
Might be nice to use a multi select variable instead of taking up all this space for variables? Can you create 2 two follow up issues (you can assign the changing of the queries in the breakdowns to allow for multiple includes) so we don't lose track over the weekend?

gtk-grafana (approved):
LGTM, works well locally

matyax:
Created #723 and #724

PR #719, opened 2024-08-23
LogsVolumePanel: re-run query on filters change

matyax:
With the current implementation, there was the possibility of being locked with a logs volume panel with no data, that will not be updated by changes in the top filters.

https://github.com/user-attachments/assets/f685304d-d311-49f8-b133-776baaac3741


gtk-grafana (approved):
LGTM, works well locally

PR #722, opened 2024-08-23
Value breakdowns: Add volume sort

gtk-grafana:
Quick win!
Fixes: https://github.com/grafana/explore-logs/issues/721


matyax (approved):
🏆

matyax:
Maybe also add Max? 🙈  // @svennergr

PR #725, opened 2024-08-23
Value breakdowns: Add calculated sort

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/499

When sorting by a calc value, we add that to the legend:
image

If using a non-calc value, the panels show up as they currently do:
image

Note: this doesn't add the calcs to the single view, 
Also since we renamed the calc values, the names don't line up with what is presented in the legend.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
@@ -11,38 +11,60 @@ import {
   SceneObjectState,
   SceneReactObject,
 } from '@grafana/scenes';
-import {buildDataQuery, LokiQuery} from '../../../services/query';
-import {getSortByPreference} from '../../../services/store';
-import {DataQueryError, LoadingState} from '@grafana/data';
-import {LayoutSwitcher} from './LayoutSwitcher';
-import {getQueryRunner} from '../../../services/panel';
-import {ByFrameRepeater} from './ByFrameRepeater';
-import {Alert, DrawStyle, LoadingPlaceholder} from '@grafana/ui';
-import {getFilterBreakdownValueScene} from '../../../services/fields';
-import {getLabelValue} from './SortByScene';
-import {getFieldGroupByVariable, VAR_FIELDS} from '../../../services/variables';
+import { buildDataQuery, LokiQuery } from '../../../services/query';
+import { getSortByPreference } from '../../../services/store';
+import { DataQueryError, LoadingState } from '@grafana/data';
+import { LayoutSwitcher } from './LayoutSwitcher';
+import { getQueryRunner } from '../../../services/panel';
+import { ByFrameRepeater } from './ByFrameRepeater';
+import { Alert, DrawStyle, LoadingPlaceholder } from '@grafana/ui';
+import { getFilterBreakdownValueScene } from '../../../services/fields';
+import { getLabelValue } from './SortByScene';
+import { getFieldGroupByVariable, VAR_FIELDS } from '../../../services/variables';
 import React from 'react';
 import {
   FIELDS_BREAKDOWN_GRID_TEMPLATE_COLUMNS,
   FieldsBreakdownScene,
   getFieldBreakdownExpr,
 } from './FieldsBreakdownScene';
-import {AddFilterEvent} from './AddToFiltersButton';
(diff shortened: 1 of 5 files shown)


src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
```diff
@@ -11,27 +11,27 @@ import {
...
-import { getSortByPreference } from '../../../services/store';
-import { DataQueryError, LoadingState } from '@grafana/data';
-import { LayoutSwitcher } from './LayoutSwitcher';
-import { getQueryRunner } from '../../../services/panel';
-import { ByFrameRepeater } from './ByFrameRepeater';
-import { Alert, DrawStyle, LoadingPlaceholder } from '@grafana/ui';
-import { getFilterBreakdownValueScene } from '../../../services/fields';
-import { getLabelValue } from './SortByScene';
-import { getFieldGroupByVariable, VAR_FIELDS } from '../../../services/variables';
+import {buildDataQuery, LokiQuery} from '../../../services/query';
```
gtk-grafana, line 14: uhhhh fix

src/Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene.tsx
```diff
@@ -38,12 +38,34 @@ export interface FieldValuesBreakdownSceneState extends SceneObjectState {
...
   lastFilterEvent?: AddFilterEvent;
 }
+
 export class FieldValuesBreakdownScene extends SceneObjectBase<FieldValuesBreakdownSceneState> {
   constructor(state: Partial<FieldValuesBreakdownSceneState>) {
     super(state);
     this.addActivationHandler(this.onActivate.bind(this));
   }
 
+  public static Selector({ model }: SceneComponentProps<FieldValuesBreakdownScene>) {
```
gtk-grafana, line 48: auto-format moved the static methods up top, I can revert this if we want to keep lines changed down

src/services/store.ts
```diff
@@ -59,13 +60,19 @@ export function addLastUsedDataSourceToStorage(dsKey: string) {
...
+  defaultSortBy: SortBy,
+  defaultDirection: SortDirection
+): { sortBy: SortBy | ''; direction: SortDirection } {
   const preference = localStorage.getItem(`${SORT_BY_LOCALSTORAGE_KEY}.${target}.by`) ?? '';
   const parts = preference.split('.');
   if (!parts[0] || !parts[1]) {
     return { sortBy: defaultSortBy, direction: defaultDirection };
   }
-  return { sortBy: parts[0], direction: parts[1] };
+  const sortBy = parts[0] as SortBy;
```
matyax, line 73: Yeah, that's why I just left those as just strings.

src/services/fields.ts
```diff
@@ -59,12 +61,26 @@ export function extractParserAndFieldsFromDataFrame(data: DataFrame) {
...
+  }
+  return reducerID;
+});
+
 export function getFilterBreakdownValueScene(
   getTitle: (df: DataFrame) => string,
   style: DrawStyle,
-  variableName: typeof VAR_FIELDS | typeof VAR_LABELS
+  variableName: typeof VAR_FIELDS | typeof VAR_LABELS,
+  sort: SortByScene
```
matyax, line 80:
```suggestion
  sortByScene: SortByScene
```

src/services/fields.ts
```diff
@@ -59,12 +61,26 @@ export function extractParserAndFieldsFromDataFrame(data: DataFrame) {
...
 ) {
   return (frame: DataFrame, frameIndex: number) => {
+    const reducerID = getReducerId(sort.state.sortBy);
```
matyax, line 83: Would it be too crazy to pass the state and not the Scene? Having scenes here makes me a bit itchy.

matyax (approved):
Very cool! 👏 👏

src/services/fields.ts
gtk-grafana: This actually breaks the option from being updated on sort change :'(

src/services/fields.ts
gtk-grafana: Agreed. Fixed!

src/services/store.ts
gtk-grafana: Yeah, but asserting after pulling from localstorage is expected 🤷

Issue #729, opened 2024-08-26
Persisted displayed fields: add e2e


matyax:
e2e requires an updated and released Grafana version that currently is not available.

gtk-grafana:
stale

PR #730, opened 2024-08-26
Displayed fields: persist selection in local storage and URL

### LINKED ISSUE #706: Stretch goal: persist displayed fields selections
Use case:
- User makes a query
- User opens log details
- User selects a few of the fields to be displayed
- Result: we persist this selection and reuse it in the future

matyax:
- Persisting in local storage allows users to customize displayed fields and use them in repeated sessions with the same data source and service.
- Persisting in URL allows to share this state with others

Fixes #706 

e2e in #729


matyax:
Currently, the only way to clear displayed fields is by going to log details and disable them one by one. I can think of 2 options:

- Add a clear button

imagen

- Use a multi-select drop down to externally modify these fields.

@zizzpudding what do you think?

zizzpudding:
In the interest of time, I think the "clear all" sounds like a reasonable approach. The wording, however seems to be confusing or a bit of a disconnect from the action. "clear selected fields"  sounds more like clearing some filter criteria. Is there a reason why you wouldn't use "Show original log line", like we do in Explore? 

If you have the luxury of time and want to go the multi-select list approach, can you provide a list like we do with the table view?  You could hide it in a menu behind a button that says something like "manage view" 

Here's an image of the table view field selector for reference. 
image

matyax:
> In the interest of time, I think the "clear all" sounds like a reasonable approach. The wording, however seems to be confusing or a bit of a disconnect from the action. "clear selected fields" sounds more like clearing some filter criteria. Is there a reason why you wouldn't use "Show original log line", like we do in Explore?

Agree. Love "Show original log line". I added it with a tooltip:

imagen

> If you have the luxury of time and want to go the multi-select list approach, can you provide a list like we do with the table view? You could hide it in a menu behind a button that says something like "manage view"

Sounds like a great follow up task.

Thank you @zizzpudding !

gtk-grafana:
@matyax overall looks good, but it's not working for me locally until I refresh the page?

https://github.com/user-attachments/assets/8200abca-f531-4c0d-a352-8b4af825a64b

matyax:
@gtk-grafana Ah, sorry. Forgot to mention that you need a Grafana with https://github.com/grafana/grafana/pull/92418

PR #733, opened 2024-08-27
Displayed fields: persist selection in local storage and URL

### LINKED ISSUE #706: Stretch goal: persist displayed fields selections

matyax:
Reopening https://github.com/grafana/explore-logs/pull/730 for after the release.

- Persisting in local storage allows users to customize displayed fields and use them in repeated sessions with the same data source and service.
- Persisting in URL allows to share this state with others.
- Added a button to clear the selection so a) users don't need to go and toggle each field off and b) they have a scape hatch to clear fields that are currently not visible.

Other minor changes:
- Moved and unified SERVICE_NAME. Couldn't import from the current location because it will cause an import error related with the plugin id. Added some missing usages of the constant.
- In LogListScene removed some duplication between `updateFromUrl` and `setStateFromUrl`.
- Wrap lines now updates the viz instead of re-building it.

Demo:

https://github.com/user-attachments/assets/ea3fb7f3-4d1b-4440-815e-39af3bf55707

Fixes #706 

e2e in #729 

> [!IMPORTANT]  
> Requires an underlying Grafana with this change: https://github.com/grafana/grafana/pull/92418


matyax:
🥺🥹

gtk-grafana (approved):
This looks familiar :P. LGTM

PR #735, opened 2024-08-27
Levels variable: populate values

gtk-grafana:
Populate the values of the levels variable!

image


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -220,7 +249,6 @@ function getVariableSet(initialDatasourceUid: string, initialFilters?: AdHocVari
     applyMode: 'manual',
     layout: 'vertical',
     getTagKeysProvider: () => Promise.resolve({ replace: true, values: [] }),
-    getTagValuesProvider: () => Promise.resolve({ replace: true, values: [] }),
```
matyax, line 223: OHHHHHH. I'm am very certain that it would have been way easier to spot if this wasn't called "tags".

matyax:
I added a level filter from Logs Volume and I'm not getting values:

imagen

src/services/variables.ts
```diff
@@ -16,8 +16,8 @@ export const VAR_FIELDS = 'fields';
 export const VAR_FIELDS_EXPR = '${fields}';
 export const VAR_PATTERNS = 'patterns';
 export const VAR_PATTERNS_EXPR = '${patterns}';
-export const VAR_LEVELS = 'levels';
```
matyax, line 19: We support levels plural, what was the reason behind this change?

gtk-grafana: because the actual name of the label is `level`, when calling `label/level/values`

gtk-grafana:
> I added a level filter from Logs Volume and I'm not getting values:

I can't seem to repro, does that service have any label values?

src/services/variables.ts
gtk-grafana: We could revert the change and hardcode the 'level' string in the `getTagValuesProvider` though

gtk-grafana:
I can repro in loki-dev, gdev was throwing me off because it has an indexed label called `level`, since we can't call the label/.../values endpoint for structured metadata we might be SOL.

I guess we could fallback to parse the logs query, but that's kinda hacky?

gtk-grafana:
Well that sucks, I guess we need a way to get structured metadata values from loki before we can do this. Closing out, I'll add the e2e fix to another branch.

PR #736, opened 2024-08-28
feat(detected-fields): use `/detected_fields` API

svennergr:
Adds support for structured metadata using the detected_fields API endpoint.

Notes for your reviewer:
Firstly, I'm sorry for the size of this PR.

This PR changes a bunch of how the application works.
* The parser variable is no longer used
* * except as a hack to get the log context query working  (todo open issue in grafana/grafana)
* We no longer have a concept of "this response is json or logfmt or mixed"
* * Instead each field is checked against the response in detected_fields, and we use the parser from there, if multiple parsers are defined in the response, we'll use mixed
* If a field is included in a query, we will use the appropriate parser
* If multiple fields, with multiple parsers are included in a query, we will use mixed
* * Unless structured metadata and single other parser is used, then we should just use one
* Queries should be overall more optimal, line-filters and patterns are now coming before the parsers
* * We should only ever use the minimum required set of parsers
* * * The exception is the patterns and log panel queries, which will always use both parsers so the list of labels is consistent between detected_fields (also getting both parsers)
* Bug fixes
* * memoization in sorting was causing some time series queries to run, but the panels would never update
* * Aggregated field breakdown was re-rendering all panels, causing queries to get cancelled and then re-inited, this should be fixed now as well (but still a problem in the labels, todo create follow up issue)
* * When deleting content out of the line-filter, the empty line filter would still be included in the query unless the user clicked the "x". Now just deleting the content removes the line-filter from all queries
* New features
* * Since the detected_fields is sorted by cardinality desc, we pushed up all the non-performant time-series to the top, which was eating up browser resources
* * * Pulled in the UI from explore to only show 20 series by default, and allow the user to manually show more
* e2e rehaul:
* * Tracked down flaking tests, and updated how we block queries to help speed things up and reduce strain on the testing loki instance
* * Added a bunch of tests, some json coverage, new helper methods, and more!

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -22,7 +22,7 @@ export default defineConfig<PluginOptions>({
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : 2,
+  workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: 'html',
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -38,9 +38,7 @@ export default defineConfig<PluginOptions>({
     //   mode: 'on',
     // }
   },
-
   expect: { timeout: 15000 },
-
 
   /* Configure projects for major browsers */
   projects: [
@@ -53,7 +51,7 @@ export default defineConfig<PluginOptions>({
     // 2. Run tests in Google Chrome. Every test will start authenticated as admin user.
     {
       name: 'chromium',
-      use: { ...devices['Desktop Chrome'], storageState: 'playwright/.auth/admin.json'},
+      use: { ...devices['Desktop Chrome'], storageState: 'playwright/.auth/admin.json' },
       dependencies: ['auth'],
     },
   ],
--- a/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
@@ -6,8 +6,6 @@ import {
   AdHocFiltersVariable,
   QueryRunnerState,
   SceneComponentProps,
-  SceneFlexItem,
-  SceneFlexLayout,
   sceneGraph,
   SceneObject,
   SceneObjectBase,
@@ -24,7 +22,7 @@ import { getSortByPreference } from 'services/store';
 import {
   ALL_VARIABLE_VALUE,
   getFieldGroupByVariable,
-  getFieldsVariable,
+  getLabelsVariable,
   getLogsStreamSelector,
   LOG_STREAM_SELECTOR_EXPR,
   LogsQueryOptions,
@@ -53,6 +51,7 @@ import {
   extractParserFieldFromParserArray,
   extractParserFromDetectedFieldParserFieldValue,
 } from '../../../services/fields';
+import { EmptyLayoutScene } from './EmptyLayoutScene';
 
 export const averageFields = ['duration', 'count', 'total', 'bytes'];
 export const FIELDS_BREAKDOWN_GRID_TEMPLATE_COLUMNS = 'repeat(auto-fit, minmax(400px, 1fr))';
@@ -62,7 +61,7 @@ export interface FieldsBreakdownSceneState extends SceneObjectState {
     | (SceneReactObject & SceneObject)
     | (FieldsAggregatedBreakdownScene & SceneObject)
     | (FieldValuesBreakdownScene & SceneObject)
-    | (SceneFlexLayout & SceneObject);
+    | (EmptyLayoutScene & SceneObject);
   search: BreakdownSearchScene;
   sort: SortByScene;
   value?: string;
@@ -120,7 +119,7 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
     this._subs.add(groupByVariable.subscribeToState(this.variableChanged));
 
     this._subs.add(
-      getFieldsVariable(this).subscribeToState((newState, prevState) => {
+      getLabelsVariable(this).subscribeToState((newState, prevState) => {
         const variable = getFieldGroupByVariable(this);
         const newService = newState.filters.find((filter) => filter.key === SERVICE_NAME);
         const prevService = prevState.filters.find((filter) => filter.key === SERVICE_NAME);
@@ -138,8 +137,10 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
     this._subs.add(
       serviceScene.state.$detectedFieldsData?.subscribeToState(
         (newState: QueryRunnerState, oldState: QueryRunnerState) => {
-          if (newState.data?.state === LoadingState.Done && newState.data.series?.[0]) {
-            this.updateOptions(newState.data.series?.[0]);
+          if (newState.data?.state === LoadingState.Done) {
+            if (newState.data.series?.[0]) {
+              this.updateOptions(newState.data.series?.[0]);
+            }
           }
         }
       )
@@ -156,7 +157,8 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
     if (
       newState.value !== oldState.value ||
       !areArraysEqual(newState.options, oldState.options) ||
-      this.state.body === undefined
+      this.state.body === undefined ||
+      this.state.body instanceof EmptyLayoutScene
     ) {
       this.updateBody(newState);
     }
@@ -164,6 +166,10 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
 
   private updateOptions(dataFrame: DataFrame) {
     if (!dataFrame || !dataFrame.length) {
+      this.setState({
+        loading: false,
+        body: new EmptyLayoutScene({ type: 'fields' }),
+      });
       return;
     }
 
@@ -233,7 +239,7 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
       if (variablesToClear.length > 1) {
         stateUpdate.body = this.buildClearFiltersLayout(() => this.clearVariables(variablesToClear));
       } else {
-        stateUpdate.body = this.buildEmptyLayout();
+        stateUpdate.body = new EmptyLayoutScene({ type: 'fields' });
       }
     } else {
       // Otherwise update the body, but don't re-instantiate if it's already the right class
@@ -244,7 +250,7 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
       } else if (
         // If the body hasn't been created, or the no-data views are active, we want to replace and render the correct scene
         this.state.body === undefined ||
-        this.state.body instanceof SceneFlexLayout ||
+        this.state.body instanceof EmptyLayoutScene ||
         this.state.body instanceof SceneReactObject
       ) {
         stateUpdate.body =
@@ -281,34 +287,6 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
       }
     });
   };
-
-  private buildEmptyLayout() {
-    return new SceneFlexLayout({
-      direction: 'column',
-      children: [
-        new SceneFlexItem({
-          body: new SceneReactObject({
-            reactNode: (
-              <GrotError>
-                <Alert title="" severity="warning">
-                  We did not find any fields for the given timerange. Please{' '}
-                  <a
-                    className={emptyStateStyles.link}
-                    href="https://forms.gle/1sYWCTPvD72T1dPH9"
-                    target="_blank"
-                    rel="noopener noreferrer"
-                  >
-                    let us know
-                  </a>{' '}
-                  if you think this is a mistake.
-                </Alert>
-              </GrotError>
-            ),
-          }),
-        }),
-      ],
-    });
-  }
 
   private buildClearFiltersLayout(clearCallback: () => void) {
     return new SceneReactObject({
--- a/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
@@ -158,7 +158,8 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
       newState.value !== oldState.value ||
       !areArraysEqual(newState.options, oldState.options) ||
       this.state.body === undefined ||
-      this.state.body instanceof EmptyLayoutScene
+      this.state.body instanceof EmptyLayoutScene ||
+      this.state.body instanceof SceneReactObject
     ) {
       this.updateBody(newState);
     }
--- a/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
@@ -61,7 +61,7 @@ export interface FieldsBreakdownSceneState extends SceneObjectState {
   loading?: boolean;
   error?: string;
   blockingMessage?: string;
-  changeFields?: (n: string[]) => void;
+  changeFieldCount?: (n: number) => void;
 }
 
 export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneState> {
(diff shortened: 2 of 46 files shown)


src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.tsx
```diff
@@ -329,3 +351,32 @@ export function buildLabelValuesBreakdownActionScene(value: string) {
     ],
   });
 }
+
+export function buildLabelsQuery(sceneRef: SceneObject, optionValue: string, optionName: string) {
```
gtk-grafana, line 355: We should probably have some unit tests for these functions

src/Components/ServiceScene/Breakdowns/LabelsAggregatedBreakdownScene.tsx
```diff
@@ -76,6 +108,13 @@ export class LabelsAggregatedBreakdownScene extends SceneObjectBase<LabelsAggreg
       );
     }
 
+    const childrenClones = children.map((child) => child.clone());
+
+    // We must subscribe to the data providers for all children after the clone or we'll see bugs in the row layout
```
gtk-grafana, line 113: Should add e2e test switching between layouts

src/Components/ServiceScene/Breakdowns/TimeSeriesLimitSeriesTitleItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+export const MAX_NUMBER_OF_TIME_SERIES = 20;
+
+export interface TimeSeriesLimitSeriesTitleItemSceneState extends SceneObjectState {
+  toggleShowAllSeries: (model: TimeSeriesLimitSeriesTitleItemScene) => void;
+  showAllSeries: boolean;
+  currentSeriesCount?: number;
+}
+
+export class TimeSeriesLimitSeriesTitleItemScene extends SceneObjectBase<TimeSeriesLimitSeriesTitleItemSceneState> {
```
gtk-grafana, line 23: needs e2e coverage

src/Components/ServiceScene/Breakdowns/TimeSeriesLimitSeriesTitleItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+      !($data instanceof SceneDataTransformer) ||
+      showAllSeries ||
+      data?.state !== LoadingState.Done ||
+      !currentSeriesCount ||
+      data.series.length < MAX_NUMBER_OF_TIME_SERIES
+    ) {
+      return null;
+    }
+
+    //@todo is there a better way to get the total number of series before transforming then accessing a private prop?
```
gtk-grafana, line 59: This should be "safe enough" for now as we will always run the transformation on dataframe update, so the _prevDataFromSource should always be the untransformed frame. Needs to be tested though if we can't find a better solution

src/Components/ServiceScene/LineFilterScene.tsx
```diff
@@ -47,7 +47,12 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
 
   updateVariable = debounce((search: string) => {
     const variable = getLineFilterVariable(this);
-    variable.changeValueTo(`|~ \`(?i)${escapeRegExp(search)}\``);
+    if (search === '') {
```
gtk-grafana, line 50: be nice to update the e2e test to ensure the line filter is removed from the queries after the user deletes the content in the input

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -69,41 +78,76 @@ export function getDetectedLabelsFrame(sceneRef: SceneObject) {
...
+        const prevServiceName = getServiceNameFromVariableState(prevState);
         if (newState.filters.length === 0) {
           this.redirectToStart();
         }
         // If we remove the service name filter, we should redirect to the start
         if (!newState.filters.some((f) => f.key === SERVICE_NAME)) {
           this.redirectToStart();
         }
+
+        // Clear filters if changing service, they might not exist, or might have a different parser
```
gtk-grafana, line 144: This is a slight degradation in UX, ideally we would check the detected_fields response against the new service and update parsers. Should create an issue for this

src/services/filters.test.ts
```diff
@@ -116,17 +116,6 @@ describe('getLabelOptions', () => {
 });
 
 describe('getFieldOptions', () => {
-  it('should remove level_extracted if it is present in the list', () => {
```
gtk-grafana, line 119: removed in the datasource

gtk-grafana: updated test to include level_extracted

playwright.config.ts
```diff
@@ -22,7 +22,7 @@ export default defineConfig<PluginOptions>({
   /* Retry on CI only */
```
matyax, line 22: These changes seems to require to be reversed (opt out, deleted space, etc)

gtk-grafana: Fixed

src/services/fields.ts
```diff
@@ -18,79 +25,88 @@ export type DetectedLabelsResponse = {
...
-  const res = extractParserAndFieldsFromDataFrame(frame);
-
-  let newType;
-  if (!res.type) {
-    newType = '';
-  } else if (res.type === 'mixed') {
-    newType = `| json  | logfmt | drop __error__, __error_details__`;
-  } else {
-    newType = ` | ${res.type}`;
+const getReducerId = memoize((sortBy: SortBy) => {
```
matyax, line 39: I would not memoize this.

src/services/fields.ts
```diff
@@ -18,79 +25,88 @@ export type DetectedLabelsResponse = {
...
-  const types: ExtractedFieldsType[] = [];
-  const linesField = data.fields.find((f) => f.name === 'Line' || f.name === 'body');
-
-  if (!linesField) {
-    return result;
+export function extractParserFieldFromParserArray(parsers?: string[]) {
+  const parsersSet = new Set(parsers?.map((v) => v.toString()) ?? []);
+
+  // Structured metadata doesn't change the parser we use, so remove it
+  parsersSet.delete('');
```
matyax, line 68: If we delete `''`, do we need a `case '':` in `extractParserFromDetectedFieldParserFieldValue`? If not, then extractParserFromDetectedFieldParserFieldValue is not needed.

src/services/fields.ts
```diff
@@ -171,3 +187,29 @@ export function getLabelTypeFromFrame(labelKey: string, frame: DataFrame, index
...
+
+export function getFilterTypeFromLabelType(type: LabelType | null, key: string, value: string): VariableFilterType {
+  switch (type) {
+    case LabelType.Indexed: {
+      return VAR_LABELS;
+    }
+    case LabelType.Parsed: {
+      return VAR_FIELDS;
+    }
+    case LabelType.StructuredMetadata: {
```
matyax, line 199: I don't understand this. LEVELS for structured metadata?

src/services/filters.ts
```diff
@@ -45,15 +45,11 @@ export function getLabelOptions(labels: string[]) {
 
   return [{ label: 'All', value: ALL_VARIABLE_VALUE }, ...labelOptions];
 }
+// @todo remove "level" as well?
```
matyax, line 48: Do we need this todo?

src/services/variables.ts
```diff
@@ -16,26 +25,49 @@ export const VAR_SERVICE = 'service';
...
 export const PATTERNS_SAMPLE_SELECTOR_EXPR = `{${VAR_LABELS_EXPR}} ${VAR_PATTERNS_EXPR} ${VAR_LOGS_FORMAT_EXPR}`;
 export const EXPLORATION_DS = { uid: VAR_DATASOURCE_EXPR };
 export const ALL_VARIABLE_VALUE = '$__all';
 export const LEVEL_VARIABLE_VALUE = 'detected_level';
-export const PATTERNS_TEXT_FILTER = 'patternsFilter';
 export const SERVICE_NAME = 'service_name';
 export const EMPTY_VARIABLE_VALUE = '""';
 
-export function getLogsStreamSelector(options: {
+export type ParserType = 'logfmt' | 'json' | 'mixed' | '';
```
matyax, line 44: I'm not sure I understand the empty parser. Should maybe be `none`?

tests/fixtures/explore.ts
```diff
@@ -1,8 +1,13 @@
-import type { Page, Locator } from '@playwright/test';
```
matyax, line 1: We should prettier --write this file.

tests/fixtures/explore.ts
```diff
@@ -20,6 +25,17 @@ export class ExplorePage {
     this.refreshPicker = this.page.getByTestId(testIds.header.refreshPicker)
   }
 
+  async setDefaultViewportSize(){
+    await this.page.setViewportSize({ width: 1280, height: 720 });
+  }
+
+  async setLimoViewportSize(){
```
matyax, line 32: Limo?

matyax (commented):
Looks great! Amazing work both of you. Left a few comments, and will comment on a few things that I see while testing.

matyax:
Wrong count?

imagen

imagen

Not sure if we agreed this is acceptable after hiding. It doesn't look great in any case.

Empty page after excluding all:

imagen

matyax:
☝️  missing field in count seems to be level

imagen

matyax:
I don't get Logs context:

imagen

But I do in Explore:

imagen

Query: `{service_name=`tns-app`} |> `<_> stdout F level=info msg="HTTP client success" status=<_> url=<_> duration=<_> traceID=<_>` | json | logfmt | drop __error__, __error_details__ | status=`208``

matyax:
That's all I could find for now. Given the size of this PR, I wouldn't mind merging and following up.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -403,5 +418,62 @@ export function getFieldBreakdownExpr(field: string) {
...
+    return (
+      `avg_over_time(${getLogsStreamSelector(options)} | unwrap ` +
+      (optionValue === 'duration' ? `duration` : optionValue === 'bytes' ? `bytes` : ``) +
+      `(${optionValue}) [$__auto]) by ()`
+    );
+  } else {
+    return `sum by (${optionValue}) (count_over_time(${getLogsStreamSelector(options)} [$__auto]))`;
+  }
+}
+
```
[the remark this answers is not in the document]
gtk-grafana: Done!

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -403,5 +418,62 @@ export function getFieldBreakdownExpr(field: string) {
...
+  fieldsVariable: AdHocFiltersVariable,
+  detectedFieldsFrame?: DataFrame
+) {
+  const parserField: Field<string> | undefined = detectedFieldsFrame?.fields[2];
+  const namesField: Field<string> | undefined = detectedFieldsFrame?.fields[0];
+  const index = namesField?.values.indexOf(optionValue);
+
+  const parserForThisField =
+    index !== undefined && index !== -1
+      ? extractParserFromDetectedFieldParserFieldValue(parserField?.values?.[index] ?? 'mixed')
```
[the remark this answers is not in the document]
gtk-grafana: Functions have been renamed, they were a bit verbose.

src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
```diff
@@ -146,13 +150,19 @@ export class SelectLabelActionScene extends SceneObjectBase<SelectLabelActionSce
     }, 0);
 
     const panel = sceneGraph.getAncestor(this, VizPanel);
-    const percentage = ((logLinesWithLabelCount / logsPanelData.length) * 100).toLocaleString();
-    const description = `${this.state.labelName} exists on ${percentage}% of ${logsPanelData.length} sampled log lines`;
+    if (logLinesWithLabelCount !== undefined && logsPanelData.length > 0) {
+      const percentage = ((logLinesWithLabelCount / logsPanelData.length) * 100).toLocaleString();
```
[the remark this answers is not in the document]
gtk-grafana: Browser's current is fine IMO, but I didn't put a lot of thought into it

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -69,41 +78,76 @@ export function getDetectedLabelsFrame(sceneRef: SceneObject) {
   return serviceScene.state.$detectedLabelsData?.state.data?.series?.[0];
 }
 
+export function getDetectedFieldsFrame(sceneRef: SceneObject) {
+  const serviceScene = sceneGraph.getAncestor(sceneRef, ServiceScene);
+  return getDetectedFieldsFrameFromQueryRunnerState(serviceScene.state.$detectedFieldsData?.state);
+}
+
+export const getDetectedFieldsFrameFromQueryRunnerState = (state?: QueryRunnerState) => {
```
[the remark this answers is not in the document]
gtk-grafana: it doesn't return an array, it returns a Field

src/services/datasource.test.ts
```diff
@@ -0,0 +1,167 @@
...
+          {
+            label: 'detected_level',
+            type: 'string',
+            cardinality: 4,
+            parsers: ['logfmt'],
+          },
+        ],
+      };
+
+      // @ts-ignore
```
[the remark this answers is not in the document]
gtk-grafana: Good call, fixed

src/services/fields.ts
gtk-grafana: extractParserFromDetectedFieldParserFieldValue, now `extractParserFromString` is called in more places then here, you are correct that case statement isn't needed for this calling context, but `extractParserFromString` is still needed as we could have multiple parsers for each field, i.e. the set value could be "logfmt, json".

This is confusing though, i wonder if we used something instead of an empty string for structured metadata if that would make it easier to read?

src/services/variables.ts
gtk-grafana: swapped out for "structuredMetadata" should be a bit more clear anyway

tests/fixtures/explore.ts
gtk-grafana: haha I was being silly, I'll clean up tomorrow

gtk-grafana:
A bunch is broken with the latest loki release, pulling back to draft, hopefully there's a quick resolution

tests/fixtures/explore.ts
matyax: Haha no worries, it looked like a typo.

matyax: Now I get it, with the 3000 height.

gtk-grafana:
Not sure if it can be tested with the latest loki, but the field count and associated empty layout should be fixed now

src/services/fields.ts
gtk-grafana: Ok, it wasn't as bad as I thought last night, I think this was carrying over the old assumption before this branch that the only structured metadata was `detected_level`. Updated to only add `detected_level` as a level, any other structured metadata must be a field variable.

Might be nice to update the level variable in the future to just hold generic structured metadata (or create a new variable for it), but for now this should suffice.

tests/fixtures/explore.ts
gtk-grafana: Made it a bit more serious :)

src/services/filters.ts
gtk-grafana: Nope, we want to remove level and extracted_level AFAIK

src/services/fields.ts
gtk-grafana: @matyax this has been updated, should be a bit better now?

src/services/fields.ts
matyax: Right. The whole point of VAR_LEVELS is to be able to isolate them from the Logs Volume query and maybe other tiny extra magic with the Volume filters, but we might (in the future) achieve the same by looking for that key somewhere else.

src/services/fields.ts
matyax: Much better! Thank you.

tests/fixtures/explore.ts
gtk-grafana: I'll do you one better, now we automatically run prettier and eslint when test files are updated!

matyax: 💯

gtk-grafana:
e2e tests are broken on grafana/loki:main-aa1ac99 due to https://github.com/grafana/loki/issues/14053, but with a local loki I can confirm e2e tests are working.

This should pass e2e tests after we update the docker compose files with a loki version containing a fix for https://github.com/grafana/loki/issues/14053 and https://github.com/grafana/loki/issues/14063.

matyax:
Log Context is working 👌

matyax (approved):
👏 👏

trevorwhitney:
yay 🎉

PR #743, opened 2024-09-04
fix: grafana image tag

BitDesert:
Fix the Grafana image tag as the `11.2` tag is not existing:

> Error response from daemon: manifest for grafana/grafana:11.2 not found: manifest unknown: manifest unknown


gtk-grafana (approved):
Good catch, can confirm this fixes running docker compose up locally.

PR #744, opened 2024-09-04
Line filter: Case sensitive search

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/739

image
image


matyax:
Not sure if we need to fix this:

imagen

matyax:
I love how this control looks btw.

matyax (approved):
Other than what I mentioned in the comment, this works flawlessly.

PR #750, opened 2024-09-06
feat: run augurs computation in worker thread

sd2k:
This is mostly a PoC PR demonstrating how we might do this, in case it comes up in future. I've used workerize-loader to simplify the creation of the worker but we could probably do it by hand to avoid the dependency if we wanted.

Edit: actually I just profiled this and it results in a considerable amount less jank for the 'most relevant' sorting. Here are two profiles where I switch to the Most Relevant sorting, [before](https://share.firefox.dev/4dXMCWX) and [after](https://share.firefox.dev/4gb0xKJ) this change; note that in the second one, all of the expensive work is run in the DOM worker, and there's a lot less jank reported.

I think this needs a bunch of double-checking before merging because I may have missed something (e.g. does changing the `publicPath` to be absolute matter? ).

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
     "typescript": "4.8.4",
     "webpack": "^5.86.0",
     "webpack-cli": "^5.1.4",
-    "webpack-livereload-plugin": "^3.0.2"
+    "webpack-livereload-plugin": "^3.0.2",
+    "workerize-loader": "^2.0.2"
   },
   "engines": {
     "node": ">=20"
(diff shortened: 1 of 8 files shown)


matyax:
Thank you Ben! We'll be looking at it soon.

gtk-grafana:
Closing as stale

matyax:
I haven't seen any performance complain requiring this change, but this is something that we can come back to if needed.

PR #757, opened 2024-09-11
Labels: Fix labels list not updating when detected_labels loads while user is viewing another tab

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/755


src/services/filters.ts
```diff
@@ -46,6 +46,7 @@ export function getLabelOptions(labels: string[]) {
   return [{ label: 'All', value: ALL_VARIABLE_VALUE }, ...labelOptions];
 }
 export const FIELDS_TO_REMOVE = ['level_extracted', LEVEL_VARIABLE_VALUE, 'level'];
+export const LABELS_TO_REMOVE = [LEVEL_VARIABLE_VALUE, 'level'];
```
svennergr, line 49: I'm not super convinced we should remove `level`. If people have that as an indexed field, we should leave it as it's better performance than `detected_level`, imo.

svennergr: Also, I think we are missing `service_name` here, right? As we can currently exclude that, which throws a bunch of errors.

https://github.com/user-attachments/assets/5e1568ea-9950-408b-b7ac-4d9a3088a107

gtk-grafana: it's a bit odd seeing both detected_level and level though

generator/generator.go
```diff
@@ -206,11 +206,13 @@ var mimirPod = func(ctx context.Context, logger *AppLogger) {
...
+	podSequence := model.LabelValue(randSeq(5))
 	appLogger := NewAppLogger(model.LabelSet{
-		"cluster":      model.LabelValue(clusters[0]),
-		"namespace":    model.LabelValue("mimir"),
-		"service_name": "mimir-ingester",
-		"pod":          "mimir-ingester" + "-" + model.LabelValue(randSeq(5)),
+		"cluster":           model.LabelValue(clusters[0]),
+		"namespace":         model.LabelValue("mimir"),
+		"service_name":      "mimir-ingester",
+		"pod":               "mimir-ingester" + "-" + podSequence,
```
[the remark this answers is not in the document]
gtk-grafana: Pulled

PR #761, opened 2024-09-12
Fields: Fix incorrect field count

gtk-grafana:
So this makes the count better, but since we're only hiding panels when the fields tab is active we're never going to have a 100% "accurate" count until we stop hiding fields that have too much cardinality to display. But this PR should solve the problem of seeing a different number of panels then the field count.

Maybe in the future we can pick a label that will be used to aggregate fields with too much cardinality.

Note: Not using main as base branch to avoid fixing e2e flake in multiple PRs.
Base branch: https://github.com/grafana/explore-logs/pull/757


src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -388,6 +393,7 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
           ...body.state.children.slice(0, 1),
           breakdownViewDef.getScene((length) => {
             if (breakdownViewDef.value === 'fields') {
+              console.log('setting fields count', length);
```
svennergr, line 396:
```suggestion
```

svennergr (approved):

PR #765, opened 2024-09-12
Module: Split it up

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/764


package.json
```diff
@@ -73,6 +73,7 @@
     "tsconfig-paths": "^4.2.0",
     "typescript": "4.8.4",
     "webpack": "^5.86.0",
+    "webpack-bundle-analyzer": "^4.10.2",
```
matyax, line 76: Are we keeping this one?

matyax (approved):
Awesome!

webpack.config.ts
```diff
@@ -1,14 +1,18 @@
 import type { Configuration } from 'webpack';
```
matyax, line 1: Same, just double checking we want to keep these changes.

src/module.tsx
```diff
@@ -0,0 +1,23 @@
+import { lazy } from 'react';
+import { AppPlugin } from '@grafana/data';
+
+const App = lazy(async () => {
```
svennergr, line 4: Hm, I think the `lazy` breaks preloading, thus links from Explore and panels don't work with this PR.

svennergr (changes requested):

gtk-grafana:
Breaks extensions, apparently there's no way to register links async, we can either break stuff up as much as possible to make the module bundle just the links and the lezer/logql package:
https://github.com/grafana/explore-logs/pull/768

Or we'll need to rewrite the parsing to exclude lezer. Either way this will have to be dealt with post GA

PR #766, opened 2024-09-13
feat(generator): log `traceID` as structured metadata

svennergr:
logs `traceID` as structured metadata. There's a 50% chance to reuse a traceId, such as there should be logs across the same trace.


cyriltovena (approved):
LGTM

gtk-grafana:
A few e2e tests have changed behavior with this change:
* should see empty fields UI
* * You have discovered a bug! when we hide all of the fields due to too many series, we don't show the empty state UI
* should only load fields that are in the viewport
* * We just need to add one more to the total panel count after scrolling to the bottom
* should exclude caller, request should contain logfmt
* * Need to add one more to each of the panel count assertions

PR #768, opened 2024-09-13
Module: Split it up + heavy refactor

gtk-grafana:
We apparently cannot register links async in the plugin module, so this PR attempts to split out as many of the imports as possible so we can get the slimmest module.tsx bundle

Fixes: https://github.com/grafana/explore-logs/issues/764

Before:
image

After:
image


src/Components/LogExplorationPage.tsx
```diff
@@ -36,3 +36,5 @@ export function LogExplorationView() {
 
   return <scene.Component model={scene} />;
 }
+
+export default LogExplorationView;
```
matyax, line 40: Are default exports a requirement for code splitting?

matyax (approved):
This is great! 👏

src/Components/LogExplorationPage.tsx
gtk-grafana: I don't think it's required but it makes the imports a bit easier to write

PR #780, opened 2024-09-18
fix(patterns-breakdown): fix expression to create pattern key breakdown

svennergr:
The query expression used to create pattern-key breakdowns is currently broken and missing the enclosing brackets (`{` and `}`).


gtk-grafana:
LGTM, can confirm this fixes the bug locally.
Could we add an e2e test for basic assertion that the breakdowns are working?

gtk-grafana (approved):

PR #782, opened 2024-09-18
Shard query splitting: send the whole stream selector to fetch shard values

matyax:
Seen in some recent tests in ops, where with filters applied that generated a new stream selector, we were getting all the shard values instead of those that correspond to that particular query.


gtk-grafana:
Looks good, and things appear to work, but I'm not exactly sure how to test?

matyax:
Good call! Could have added repro instructions. Basically, apply a label filter and the values call should use it in its payload. Depending to the data and time range, it would return different shards to query (or not). The important piece is to send the whole stream selector to /values, not just the service.

gtk-grafana (approved):
Looks good, works well. Shame the values endpoint can't take structured metadata tho

PR #783, opened 2024-09-18
label values: fix label values stuck in loading state

gtk-grafana:
Fix bug in the label values where errors are not being surfaced to users and the loading state is left up indefinitely.

If panels are visible, the error will be surfaced within the viz panel, and as a toast message, if panels are not visible the error will be added to the body and remain visible until the next query response.

image

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
+++ b/src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
@@ -49,8 +49,12 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
       const data = sceneGraph.getData(this);
 
       this._subs.add(
-        data.subscribeToState((data) => {
-          if (data.data?.state === LoadingState.Done) {
+        data.subscribeToState((data, prevData) => {
+          if (
+            data.data?.state === LoadingState.Done ||
+            (data.data?.state === LoadingState.Streaming &&
+              data.data.series.length > (prevData.data?.series.length ?? 0))
+          ) {
             this.performRepeat(data.data);
           }
         })
--- a/src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
@@ -124,7 +124,7 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
     // If we're in an error state
     if (newState.data?.state === LoadingState.Error && this.activeLayoutContainsNoPanels()) {
       const activeLayout = this.getActiveLayout();
-      // And the active layout is not showing panels
+      // And the active layout is grid or rows, and doesn't have any panels
       if (activeLayout instanceof ByFrameRepeater) {
         const errorState = this.getErrorStateAlert(newState.data.errors);
         // Replace the loading or error state with new error
--- a/src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
@@ -161,7 +161,7 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
     return new SceneReactObject({
       reactNode: (
         <Alert title={'Something went wrong with your request'} severity={'error'}>
-          {errors?.map((err, key) => this.getErrorJSX(key, err))}
+          {errors?.map((err, key) => this.renderError(key, err))}
         </Alert>
       ),
     });
@@ -279,7 +279,7 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
       if (!this.activeLayoutContainsNoPanels()) {
         appEvents.publish({
           type: AppEvents.alertError.name,
-          payload: errorArray?.map((err, key) => this.getErrorJSX(key, err)),
+          payload: errorArray?.map((err, key) => this.renderError(key, err)),
         });
       }
       this.setState({
@@ -288,7 +288,7 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
     }
   }
 
-  private getErrorJSX(key: number, err: DataQueryError) {
+  private renderError(key: number, err: DataQueryError) {
     return (
       <div key={key}>
         {err.status && (
--- a/src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
@@ -7,13 +7,14 @@ import {
   SceneFlexItem,
   SceneFlexLayout,
   sceneGraph,
+  SceneObject,
   SceneObjectBase,
   SceneObjectState,
   SceneReactObject,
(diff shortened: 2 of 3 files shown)


src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
```diff
@@ -49,8 +49,12 @@ export class ByFrameRepeater extends SceneObjectBase<ByFrameRepeaterState> {
       const data = sceneGraph.getData(this);
 
       this._subs.add(
-        data.subscribeToState((data) => {
-          if (data.data?.state === LoadingState.Done) {
+        data.subscribeToState((data, prevData) => {
+          if (
+            data.data?.state === LoadingState.Done ||
```
matyax, line 54: Maybe just != Error and we account for any change we make? E.g. we go back to Loading for sharding, etc.

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -101,6 +120,51 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
+        // Replace the loading or error state with new error
+        activeLayout.state.body.setState({
+          children: [errorState],
+        });
+      }
+    }
+  }
+
+  private getActiveLayout(): ByFrameRepeater | SceneFlexLayout | undefined {
+    const layoutSwitcher = this.state.body;
```
matyax, line 139: I think you can just check if (!this.state.body) { return undefined; } and simplify a bit.

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -101,6 +120,51 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
+    if (activeLayout instanceof ByFrameRepeater || activeLayout instanceof SceneFlexLayout) {
+      return activeLayout;
+    }
+    return undefined;
+  }
+
+  private activeLayoutContainsNoPanels(): boolean {
+    const activeLayout = this.getActiveLayout();
+
+    if (activeLayout instanceof ByFrameRepeater) {
```
matyax, line 150: No SceneFlexLayout? See line 141.

matyax (approved):
Looks great to me.

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
matyax: I mean, it's either Done, or Loading, or Streaming, or else Error. Not a big deal in any case.

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
gtk-grafana: Yeah the single view doesn't count as panels IMO as it will always render something even without dataframe

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
gtk-grafana: yeah but then I think we need a type assertion

src/Components/ServiceScene/Breakdowns/ByFrameRepeater.tsx
gtk-grafana: Plus with sharding we can have data we want to viz and errors?

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
gtk-grafana: but this should be documented at the very least, updating

gtk-grafana: to clarify, since the single view panel is always rendered, even if no data is received but an error is, the error shows up in the UI and persists until the a request without an error is received. The row/grid views can have an error without any panels, so previously there would be no indication of anything being wrong.

Although now that I think about this while the single view should be fine as it will show a loading state when new queries are fired, we might want to re-add a loading state to the grid/rows when only the error message is visible.

gtk-grafana: A bit of an edge case, created an issue for it tho: https://github.com/grafana/explore-logs/issues/785

matyax:
No further comments 👌

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -101,6 +120,51 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
+    }
+
+    return false;
+  }
+
+  private getErrorStateAlert(errors: DataQueryError[] | undefined) {
+    return new SceneReactObject({
+      reactNode: (
+        <Alert title={'Something went wrong with your request'} severity={'error'}>
+          {errors?.map((err, key) => this.getErrorJSX(key, err))}
```
svennergr, line 164:
```suggestion
          {errors?.map((err, key) => this.renderError(key, err))}
```

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -197,9 +261,58 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
+          payload: errorArray?.map((err, key) => this.getErrorJSX(key, err)),
+        });
+      }
+      this.setState({
+        errors,
+      });
+    }
+  }
+
+  private getErrorJSX(key: number, err: DataQueryError) {
```
svennergr, line 291:
```suggestion
  private renderError(key: number, err: DataQueryError) {
```

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -197,9 +261,58 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
+        displayError.displayed = true;
+      }
+    }
+
+    if (errorArray.length) {
+      // If we don't have any panels the error message will replace the loading state, we want to set it as displayed but not render the toast
+      if (!this.activeLayoutContainsNoPanels()) {
+        appEvents.publish({
+          type: AppEvents.alertError.name,
+          payload: errorArray?.map((err, key) => this.getErrorJSX(key, err)),
```
svennergr, line 282:
```suggestion
          payload: errorArray?.map((err, key) => this.renderError(key, err)),
```

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -197,9 +261,58 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
     });
   }
 
+  private showErrorToast(errors: DisplayErrors) {
```
svennergr, line 264: Probably deserves a more unified place than stuffing this into the Label breakdown, but good for now.

svennergr (approved):
Just a few nits, LGTM!

PR #789, opened 2024-09-19
Aggregated metrics: Use sum_over_time query for aggregated metric queries

gtk-grafana:
Current log counts being displayed are incorrect, the updated sum_over_time query unwrapping by count should return the correct volume counts.

### DIFF
--- a/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
+++ b/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
@@ -176,6 +176,20 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
           }
         })
       );
+
+      // agg metrics need parser and unwrap, have to tear down and rebuild panels when the variable changes
+      this._subs.add(
+        getServiceNameVariable(this).subscribeToState((newState, prevState) => {
+          if (newState.value !== prevState.value) {
+            // Clear the body panels
+            this.setState({
+              body: new SceneCSSGridLayout({ children: [] }),
+            });
+            // And re-init with the new query
+            this.updateBody();
+          }
+        })
+      );
     }
   }
 
@@ -238,6 +252,7 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
       const newChildren: SceneCSSGridItem[] = [];
       const existingChildren: SceneCSSGridItem[] = this.state.body.state.children as SceneCSSGridItem[];
       const timeRange = sceneGraph.getTimeRange(this).state.value;
+      const serviceNameVar = getServiceNameVariable(this);
 
       for (const service of servicesToQuery.slice(0, SERVICES_LIMIT)) {
         const existing = existingChildren.filter((child) => {
@@ -250,7 +265,10 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
           newChildren.push(existing[0], existing[1]);
         } else {
           // for each service, we create a layout with timeseries and logs panel
-          newChildren.push(this.buildServiceLayout(service, timeRange), this.buildServiceLogsLayout(service));
+          newChildren.push(
+            this.buildServiceLayout(service, timeRange, serviceNameVar),
+            this.buildServiceLogsLayout(service)
+          );
         }
       }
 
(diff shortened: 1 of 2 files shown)


src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -292,7 +310,10 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
     return `{${SERVICE_NAME}=\`${service}\`}${levelFilter}`;
   }
 
-  private getMetricExpression(service: string) {
+  private getMetricExpression(service: string, serviceName: CustomConstantVariable) {
+    if (serviceName.state.value === AGGREGATED_SERVICE_NAME) {
+      return `sum by (${LEVEL_VARIABLE_VALUE}) (sum_over_time({${AGGREGATED_SERVICE_NAME}=\`${service}\`} | logfmt | drop __error__ | unwrap count [$__auto]))`;
+    }
     return `sum by (${LEVEL_VARIABLE_VALUE}) (count_over_time({${SERVICE_NAME_EXPR}=\`${service}\`} | drop __error__ [$__auto]))`;
```
gtk-grafana, line 317: @svennergr @matyax does anyone know why we're dropping errors if we're not using a parser? 😕

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -292,8 +310,11 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
     return `{${SERVICE_NAME}=\`${service}\`}${levelFilter}`;
   }
 
-  private getMetricExpression(service: string) {
-    return `sum by (${LEVEL_VARIABLE_VALUE}) (count_over_time({${SERVICE_NAME_EXPR}=\`${service}\`} | drop __error__ [$__auto]))`;
+  private getMetricExpression(service: string, serviceName: CustomConstantVariable) {
+    if (serviceName.state.value === AGGREGATED_SERVICE_NAME) {
+      return `sum by (${LEVEL_VARIABLE_VALUE}) (sum_over_time({${AGGREGATED_SERVICE_NAME}=\`${service}\`} | logfmt | drop __error__ | unwrap count [$__auto]))`;
+    }
+    return `sum by (${LEVEL_VARIABLE_VALUE}) (count_over_time({${SERVICE_NAME}=\`${service}\`} | drop __error__ [$__auto]))`;
```
trevorwhitney, line 317: I don't _think_ we need the `drop __error__` here. If we did need it, it would mean that getting values from structured metadata could produce errors, which I don't think is possible, but will double check.

trevorwhitney (approved):
LGTM, thanks for the quick fix!

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
trevorwhitney: so yeah, for a `count_over_time` using a structured metadata grouping I can't find a way it would produce an error

svennergr: I think we can remove `| drop __error__` from both expressions.

trevorwhitney: yeah, good point, especially since we're doing the grouping by level, they'll effectively get dropped anyway

trevorwhitney: for the `unwrap` we may want a `| __error__=""` instead, which will drop all lines that could not be unwrapped. However, that won't happen since we're creating this stream, so it might be a belt and suspenders situation

gtk-grafana: Hmm, if something changes in the way we create the stream, and we do need to drop errors, that would mean we have a bug in the way we're creating the stream that we'd need to fix on the loki side. If we do drop errors then we still have a bug in loki but we would probably not catch it as quickly, so my argument would be to not drop or exclude errors in either query

trevorwhitney: I agree

svennergr (approved):
LGTM

PR #793, opened 2024-09-20
fix(shardQuerySplitting): do not emit empty data

matyax:
Fixes showing "no data" while streaming.

### DIFF
--- a/src/services/shardQuerySplitting.ts
+++ b/src/services/shardQuerySplitting.ts
@@ -129,13 +129,13 @@ function splitQueriesByStreamShard(
             return;
           }
         }
-        if (mergedResponse.data.length) {
-          mergedResponse.state = LoadingState.Streaming;
-        }
         mergedResponse = combineResponses(mergedResponse, partialResponse);
       },
       complete: () => {
-        subscriber.next(mergedResponse);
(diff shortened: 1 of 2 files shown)


src/services/shardQuerySplitting.ts
```diff
@@ -129,13 +129,13 @@ function splitQueriesByStreamShard(
...
         }
-        if (mergedResponse.data.length) {
-          mergedResponse.state = LoadingState.Streaming;
-        }
         mergedResponse = combineResponses(mergedResponse, partialResponse);
       },
       complete: () => {
-        subscriber.next(mergedResponse);
+        // Prevent flashing "no data"
+        if (mergedResponse.data.length) {
```
svennergr, line 136: What if we only ever get empty frames. I think we need to at least call `subscriber.next(mergedResponse);` when the `state` is `LoadingState.Done` or something like that?

matyax: Good question. The done() function takes care of that. https://github.com/grafana/explore-logs/blob/main/src/services/shardQuerySplitting.ts#L77

svennergr (approved):
🚀

PR #799, opened 2024-09-23
Frontend error handling: Better FetchError handling

gtk-grafana:
FetchError is thrown from backendSrv, right now we log an "unknown error" message, and lose needed context on the error object when sending to Faro.

### DIFF
--- a/src/services/logger.ts
+++ b/src/services/logger.ts
@@ -77,12 +77,8 @@ function populateFetchErrorContext(err: unknown | FetchError, context: LogContex
   }
 }
 
-const attemptFaroErr = (err: Error | FetchError | unknown, context?: LogContext) => {
+const attemptFaroErr = (err: Error | FetchError | unknown, context: LogContext) => {
   try {
-    if (context === undefined) {
-      context = {};
-    }
-
     populateFetchErrorContext(err, context);
 
     if (err instanceof Error) {
--- a/src/services/logger.ts
+++ b/src/services/logger.ts
@@ -42,34 +42,35 @@ const attemptFaroWarn = (msg: string, context?: LogContext) => {
 
 /**
  * Checks unknown error for properties from FetchError and adds them to the context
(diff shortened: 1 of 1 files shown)


src/services/logger.ts
```diff
@@ -40,16 +40,80 @@ const attemptFaroWarn = (msg: string, context?: LogContext) => {
   }
 };
 
-const attemptFaroErr = (err: Error | unknown, context?: LogContext) => {
+/**
+ * Checks unknown error for properties from FetchError and adds them to the context
+ * Property names from the error are prepended with an underscore when added to the error context sent to Faro
+ * Note this renames the "message" to "_errorMessage" in hopes of reducing conflicts of future use of that property name
```
matyax, line 46: Do we really need to care about this?

matyax (approved):
👌

src/services/logger.ts
```diff
@@ -40,16 +40,80 @@ const attemptFaroWarn = (msg: string, context?: LogContext) => {
...
+ * @param err
+ * @param context
+ */
+function populateFetchErrorContext(err: unknown | FetchError, context: LogContext) {
```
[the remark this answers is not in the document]
gtk-grafana: Pushed an update and good catch, no need to run all those runtime checks. Wasn't sure about removing the string checks as those props are still unknown even if the root err is an object.

src/services/logger.ts
gtk-grafana: Probably not no 🤷

gtk-grafana: but if we're really digging into what came from where it might speed things up a bit

src/services/logger.ts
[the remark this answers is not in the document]
matyax: Not really important, but was part of what I wanted to say.

matyax (approved):
👌

svennergr (approved):

PR #801, opened 2024-09-23
Breakdowns: Remove service_name requirement

gtk-grafana:
This PR attempts to remove the requirement that a service_name is selected.
On label/filter update, if the label name currently stored in the URL slug is removed, we will grab the first non-exclusion label variable and use it as the new URL slug, e.g. `a/grafana-lokiexplore-app/explore/cluster/dev-us-central-0/${tabSlug}`


matyax:
At first glance, I feel inclined to keep the behavior of going back to the service selection when, specifically, the service name has been removed, but I'm not 100% sure. On one hand it was a good shortcut for the service selection page, but I don't have enough data to prove that this is a valid workflow that we need to keep; on the other, I feel that this change potentially enriches the capabilities of explore logs, allowing the user to continue filtering and iterating on their search in the drilldown views.

src/Components/Pages.tsx
```diff
@@ -107,33 +113,38 @@ function makeEmptyScene(): (routeMatch: SceneRouteMatch) => EmbeddedScene {
...
-  routeMatch: SceneRouteMatch<{ service: string; label: string }>,
+  routeMatch: RouteMatch,
   parent: SceneAppPageLike,
   slug: ChildDrilldownSlugs
 ): SceneAppPage {
-  const { service } = extractServiceFromRoute(routeMatch);
-  const { label } = extractLabelNameFromRoute(routeMatch);
+  const { labelName, labelValue, breakdownLabel } = extractValuesFromRoute(routeMatch);
+
+  if (!breakdownLabel) {
```
matyax, line 138: Do we gracefully fail here? We can definitely log the error, but hopefully we should take the user back to the service selection page. Basically, anything but a broken page with no explanation.

src/services/expressions.ts
```diff
@@ -21,37 +20,30 @@ import { getFieldsVariable, getLabelsVariable } from './variableGetters';
...
+      return `sum(count_over_time({${VAR_LABELS_EXPR}} ${metadataExpressionToAdd} ${VAR_LINE_FILTER_EXPR} ${VAR_PATTERNS_EXPR} ${JSON_FORMAT_EXPR} ${VAR_FIELDS_EXPR} [$__auto])) by (${streamSelectorName})`;
     }
     if (parser === 'logfmt') {
-      return `sum(count_over_time({${streamSelectors}} ${metadataExpressionToAdd} ${VAR_LINE_FILTER_EXPR} ${VAR_PATTERNS_EXPR} ${LOGS_FORMAT_EXPR} ${VAR_FIELDS_EXPR} [$__auto])) by (${streamSelectorName})`;
+      return `sum(count_over_time({${VAR_LABELS_EXPR}} ${metadataExpressionToAdd} ${VAR_LINE_FILTER_EXPR} ${VAR_PATTERNS_EXPR} ${LOGS_FORMAT_EXPR} ${VAR_FIELDS_EXPR} [$__auto])) by (${streamSelectorName})`;
     }
   }
-  return `sum(count_over_time({${streamSelectors}} ${metadataExpressionToAdd} ${VAR_LINE_FILTER_EXPR} ${VAR_PATTERNS_EXPR} [$__auto])) by (${streamSelectorName})`;
+  return `sum(count_over_time({${VAR_LABELS_EXPR}} ${metadataExpressionToAdd} ${VAR_LINE_FILTER_EXPR} ${VAR_PATTERNS_EXPR} ${VAR_FIELDS_EXPR} [$__auto])) by (${streamSelectorName})`;
```
matyax, line 48: Was this missing? What's the context of this change?

src/Components/ServiceScene/Breakdowns/LabelBreakdownScene.test.ts
```diff
@@ -12,7 +12,7 @@ describe('buildLabelsQuery', () => {
 
     const result = buildLabelsQuery({} as SceneObject, VAR_LABEL_GROUP_BY_EXPR, 'cluster');
     expect(result).toMatchObject({
-      expr: `sum(count_over_time({\${filters} ,cluster != ""}  \${levels} \${patterns} \${lineFilter} [$__auto])) by (${VAR_LABEL_GROUP_BY_EXPR})`,
+      expr: `sum(count_over_time({\${filters} ,cluster != ""}  \${levels} \${patterns} \${lineFilter}  \${fields} [$__auto])) by (${VAR_LABEL_GROUP_BY_EXPR})`,
```
matyax, line 15: This change inspired my question on the expressions.ts change, as I don't understand it in the context of the PR.

src/Components/Pages.tsx
gtk-grafana: I'm not able to repro throwing this error at all, if there is no breakdownLabel in the URL it won't route to this endpoint

src/services/expressions.ts
gtk-grafana: Yeah this was missing, if you had structured metadata added for another field besides the current breakdown it wouldn't add it to the query if we hadn't added a parser yet

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -134,22 +143,45 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
+                ...prevRouteMatch,
+                params: {
+                  ...prevRouteMatch?.params,
+                  labelName: newPrimaryLabel.key,
+                  labelValue: replaceSlash(newPrimaryLabel.value),
+                },
+                url: prevRouteMatch?.url ?? '',
+                path: prevRouteMatch?.path ?? '',
+                isExact: prevRouteMatch?.isExact ?? true,
+              },
```
matyax, line 174: Do we need these other properties url, path, isExact?

src/services/expressions.ts
```diff
@@ -21,37 +20,30 @@ import { getFieldsVariable, getLabelsVariable } from './variableGetters';
...
-    if (streamSelectorName !== LEVEL_VARIABLE_VALUE) {
-      labelExpressionToAdd = { key: streamSelectorName, operator: '!=', value: '' };
-    } else {
+    if (streamSelectorName === LEVEL_VARIABLE_VALUE) {
       metadataExpressionToAdd = `| ${LEVEL_VARIABLE_VALUE} != ""`;
     }
   }
 
-  const labelFilters = [...labelsVariable.state.filters, labelExpressionToAdd].filter(isDefined);
-  const streamSelectors = renderLogQLLabelFilters(labelFilters);
```
matyax, line 39: As far as I can see, renderLogQLLabelFilters() does a lot more than just interpolating label filters. Are we sure about this change?

src/services/routing.ts
```diff
@@ -102,6 +111,26 @@ export function getDrilldownSlug() {
...
+/**
+ * The "primary" label, is the replacement for the service_name paradigm
+ * It must be an indexed label with an include filter
+ * Note: Will return the label as it exists in the url, so "service_name" will be returned as "service", we'll need to adjust for this case if we want to support URLs from before this change
+ */
+export function getPrimaryLabelFromUrl(): RouteProps {
+  const location = locationService.getLocation();
+  const startOfUrl = '/a/grafana-lokiexplore-app/explore';
+  const endOfUrl = location.pathname.slice(location.pathname.indexOf(startOfUrl) + startOfUrl.length + 1);
+  const routeParams = endOfUrl.split('/');
```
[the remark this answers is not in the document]
gtk-grafana: I added an error if the label name/value (formerly known as service name/value) are not in the URL, but this is another case where if the url structure was that mangled it would never route to our existing pages, but it doesn't hurt to fail loudly in case future development changes the routing.

I'm not sure about the array destructuring because the breakdownLabel is not the third element in the array but the 4th, we could define a variable that we don't do anything with but that always hurts :P

matyax:
It works super well! One last thing, can we capture if the user changes the URL and send them back to the service selection? Right now you stay with the previous label.

URL

Only if it doesn't require a lot of extra complexity, otherwise it's not worth it for now.

src/services/expressions.ts
gtk-grafana: It's only required if we're manually building strings instead of interpolating from the VAR_LABELS variable, which already calls renderLogQLLabelFilters automatically when building expressions, see the `expressionBuilder` method on `AdHocFiltersVariable` class

gtk-grafana: Wouldn't hurt to do some more testing to make sure queries are being built correctly though, and if we find any bugs we should add those as test cases

src/services/expressions.ts
matyax: Fantastic, thanks.

src/services/routing.ts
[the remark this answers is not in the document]
gtk-grafana: Actually nvm I need to remove this error I just added as it gets thrown when redirecting back to the service selection when the primary label is removed.

gtk-grafana:
I did find a bug where the "service_name" is showing up in the url when changing services. Looking into a fix now

src/services/routing.ts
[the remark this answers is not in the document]
gtk-grafana: Fixed

src/services/routing.ts
```diff
@@ -102,6 +114,27 @@ export function getDrilldownSlug() {
...
+
+  let labelName = routeParams[0];
+  const labelValue = routeParams[1];
+  const breakdownLabel = routeParams?.[3];
```
matyax, line 130:
```suggestion
  const breakdownLabel = routeParams[3];
```

Not needed.

matyax (approved):
👌

PR #805, opened 2024-09-25
Frontend logging: more generic context

gtk-grafana:
Narrowing each prop individually was cumbersome and not very scalable, this attempts to grab string/numeric/boolean properties from errors that are Records

### DIFF
--- a/src/services/logger.ts
+++ b/src/services/logger.ts
@@ -45,8 +45,6 @@ const attemptFaroWarn = (msg: string, context?: LogContext) => {
 const isRecord = (obj: unknown): obj is Record<string, unknown> => typeof obj === 'object';
 /**
  * Checks unknown error for properties from Records like FetchError and adds them to the context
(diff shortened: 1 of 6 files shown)


src/services/logger.ts
```diff
@@ -42,46 +42,39 @@ const attemptFaroWarn = (msg: string, context?: LogContext) => {
...
-    if (hasCancelled(err) && typeof err.cancelled === 'boolean' && context.cancelled) {
-      context._cancelled = err.cancelled.toString();
-    }
-    if (hasStatusText(err) && typeof err.statusText === 'string') {
-      context._statusText = err.statusText;
-    }
-    if (hasHandled(err) && typeof err.isHandled === 'boolean') {
-      context._isHandled = err.isHandled.toString();
+    if (isRecord(err)) {
+      Object.keys(err).forEach((key: string) => {
```
matyax, line 54: Better 👌

PR #806, opened 2024-09-26
Combine responses: improve merging of duplicated entries

matyax:
Found while working in grafana/grafana. Main fixes are:

- Improved field comparison, where `undefined === undefined` was causing false positives of existing entries.
- Fixed merging of log frames when they have overlapping or duplicated results.
- Added support for the case where `closestIdx()` returns -1 when the passed array is empty.
- Fixed request retrying, moving clearing the timeout to the correct place.

### DIFF
--- a/src/services/combineResponses.ts
+++ b/src/services/combineResponses.ts
@@ -70,19 +70,9 @@ export function mergeFrames(dest: DataFrame, source: DataFrame) {
   const totalFields = Math.max(dest.fields.length, source.fields.length);
 
   for (let i = 0; i < sourceTimeValues.length; i++) {
-    const destTimeValues = destTimeField.values.slice(0) ?? [];
-    const destNanosValues = destTimeField.nanos?.slice(0);
-    const destIdValues = destIdField?.values.slice(0) ?? [];
     const destIdx = resolveIdx(destTimeField, sourceTimeField, i);
 
-    const entryExistsInDest = compareEntries(
-      { ...destTimeField, values: destTimeValues, nanos: destNanosValues },
-      destIdField ? { ...destIdField, values: destIdValues } : destIdField,
-      destIdx,
-      sourceTimeField,
-      sourceIdField,
-      i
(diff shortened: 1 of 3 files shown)


src/services/combineResponses.ts
```diff
@@ -72,11 +72,12 @@ export function mergeFrames(dest: DataFrame, source: DataFrame) {
...
     const destTimeValues = destTimeField.values.slice(0) ?? [];
     const destNanosValues = destTimeField.nanos?.slice(0);
+    const destIdValues = destIdField?.values.slice(0) ?? [];
     const destIdx = resolveIdx(destTimeField, sourceTimeField, i);
 
-    const entryExistsInDest = compareTimestamps(
+    const entryExistsInDest = compareEntries(
       { ...destTimeField, values: destTimeValues, nanos: destNanosValues },
-      destIdField,
+      destIdField ? { ...destIdField, values: destIdValues } : destIdField,
```
svennergr, line 80: Not sure I understand, is this to create a copy of the `Field`?

Maybe still worth doing:
```suggestion
      destIdField ? { ...destIdField, values: destIdValues } : undefined,
```

src/services/shardQuerySplitting.ts
```diff
@@ -191,6 +189,9 @@ function splitQueriesByStreamShard(
       });
     return () => {
       shouldStop = true;
+      if (retryTimer) {
+        clearTimeout(retryTimer);
```
svennergr, line 193:
```suggestion
        clearTimeout(retryTimer);
        retryTimer = null;
```

svennergr (approved):

src/services/combineResponses.ts
matyax: Great question. I've been using it so that the destination fields were not affected by incoming values, but you made me realize that now that's a leftover of the original implementation. I'll see if it still makes sense to keep it.

PR #807, opened 2024-09-30
Line Filter: increase debounce time and add enter support

matyax:
To decrease the chance of firing a lot of requests while a user is typing:

- Increased debounce time to 1 second.
- Added support for running the query immediately with `Enter` if the user needs it.


gtk-grafana (approved):
LGTM

PR #811, opened 2024-09-30
Service selection: Showing incorrect list of services after changing datasource on breakdown views

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/808

Updates service selection functionality to always re-run volume query on activation of service selection scene.
This bug was likely introduced with the [runtime datasource](https://github.com/grafana/explore-logs/pull/651) change.

Ideally there would be a way to know if a dataprovider has changed datasources, and we could only re-run this query when the datasource changed, and otherwise use the cached result if available, but this appears to be a consequence of using a runtime datasource.


src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -142,11 +150,13 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
     if (this.isTimeRangeTooEarlyForAggMetrics()) {
       this.onUnsupportedAggregatedMetricTimeRange();
       if (this.state.$data.state.data?.state !== LoadingState.Done) {
+        console.log('isTimeRangeTooEarlyForAggMetrics', this.state.$data);
```
matyax, line 153: We should probably remove these.

matyax (approved):

PR #824, opened 2024-10-15
chore: upgrade grafana deps to 11.2.x and update extensions to use `addLink`

svennergr:


matyax (approved):

gtk-grafana:
we didn't add any imports so the bundles won't change, but we should still double check that the module.tsx bundle is still properly sized until we have an automatic solution for this

gtk-grafana (commented):
It looks like it's working in explore and dashboards now, but seeing this warning in the console:
`Did not register links from plugin grafana-logs-app due to missing meta information. `

gtk-grafana (approved):

PR #825, opened 2024-10-15
feat(shardQuerySplitting): use rate of change to adjust group size

matyax:
WIP. PR open to get the built plugin.


matyax:
Which would work if the build actually built

PR #826, opened 2024-10-15
Structured metadata: Refactor into new variable

gtk-grafana:
This PR moves structured metadata fields into a dedicated scenes variable.
This enables us to have more control over how metadatafields are interpolated in the query, allowing us to place all metadata before any log parsers, which will result in more efficient/optimal queries.

There should be no changes to the UI, just the queries that are sent to Loki.

We still have a potential race condition where users can filter results in the table/logs panel before the detected_fields response which will give us the parser, in these cases we will fall back to adding fields with "unknown" parsers to the fields variable, and require both logfmt and json parsers for queries with that field. Created https://github.com/grafana/explore-logs/issues/837 to track


src/services/fields.ts
```diff
@@ -179,7 +187,18 @@ export function getLabelTypeFromFrame(labelKey: string, frame: DataFrame, index
   }
 }
 
-export function getFilterTypeFromLabelType(type: LabelType | null, key: string, value: string): VariableFilterType {
+export function getFilterTypeFromLabelType(
+  type: LabelType | null,
+  key: string,
+  value: string,
+  sceneRef: SceneObject
+): VariableFilterType {
```
matyax, line 195: There's something very odd with this function. `type` can be null, or have a value; independent of the type, we check for `getParserForField()`, which supports only `structuredMetadata`, so it can return something else, or nothing, but type can also be null, which goes to the default case.

I think we should refactor this before merging.

svennergr (approved):
LGTM, great job!

Just a nit, might also be interesting for @matyax:
image

When filtering for structured metadata the logs panel filter button state is not updated accordingly. Maybe Matias knows best where to adjust that.

matyax:
Taking a look.

matyax:
Yeah, great catch. We need to update https://github.com/grafana/explore-logs/blob/main/src/Components/ServiceScene/LogsPanelScene.tsx#L125-L176

src/services/fields.ts
gtk-grafana: I'm not sure I fully understood, how does this look? https://github.com/grafana/explore-logs/pull/826/commits/b643f482c1731757ab59979f0a3e0b9c9c841308

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -160,15 +181,15 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
   };
 
   private handleLabelFilter(key: string, value: string, frame: DataFrame | undefined, operator: FilterType) {
-    const labelType = frame ? getLabelTypeFromFrame(key, frame) : LabelType.Parsed;
-    const variableName = getFilterTypeFromLabelType(labelType, key, value);
-    addToFilters(key, value, operator, this, variableName);
+    const variableType = getDesiredVariableForLabel(frame, key, this);
```
matyax, line 184:
```suggestion
    const variableType = getVariableForLabel(frame, key, this);
```
Super minor, but not a big fan of these "desired".

src/services/fields.ts
matyax: Better, thanks.

matyax:
I pulled the latest changes and still don't see the active state:

imagen

matyax (approved):
🚢 🇮🇹

src/services/fields.ts
```diff
@@ -192,12 +232,7 @@ export function getFilterTypeFromLabelType(type: LabelType | null, key: string,
...
       }
-      return VAR_FIELDS;
-    }
-    default: {
-      const err = new Error(`Invalid label type for ${key}`);
-      logger.error(err, { value, msg: `Invalid label type for ${key}` });
-      throw err;
+      return VAR_METADATA;
     }
   }
```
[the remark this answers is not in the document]
gtk-grafana: LabelType can't be null anymore, and we already cover all the cases. I can add the default and throw an error back in, in case someone has a messed up dataframe that includes an unknown LabelType (not `I`, `S` or `P`)?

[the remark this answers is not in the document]
gtk-grafana: Eh, doesn't hurt to fail loud, I added a default case that throws an error.

PR #828, opened 2024-10-16
Add conditional extension point for testing sidecar functionality

aocenas:
This is an experiment to enable sidecar for Logs + other app and test it internally.

Ties in with https://github.com/grafana/explore-traces/pull/102

Needs `appSidecar` feature toggle and optionally `timeRangeProvider` for time picker sync.

https://github.com/user-attachments/assets/856501fb-6149-4b56-b0d1-bb5c6748c689


src/Components/ServiceScene/ActionBarScene.tsx
```diff
@@ -39,6 +41,10 @@ export class ActionBarScene extends SceneObjectBase<ActionBarSceneState> {
       <Box paddingY={0}>
         <div className={styles.actions}>
           <Stack gap={1}>
+            {
+              // @ts-ignore appSidecar not yet in stable runtime
```
aocenas, line 45: Not sure how to deal with runtime updates. Should they be part of this or is it better to wait for a release and do something like this in the meantime?

src/Components/ServiceScene/ActionBarScene.tsx
```diff
@@ -100,3 +106,66 @@ function getStyles(theme: GrafanaTheme2) {
...
+      setFilters(newState.filters);
+    });
+    return () => {
+      sub.unsubscribe();
+    };
+  }, [props.serviceScene]);
+
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const extensions = usePluginLinks({
+    extensionPointId: 'grafana-lokiexplore-app/toolbar',
```
aocenas, line 131: Not sure about the naming. Considering the default text in dropdown states "open related" maybe "intent" should be part of the extension point name.

src/Components/ServiceScene/ActionBarScene.tsx
```diff
@@ -39,6 +41,10 @@ export class ActionBarScene extends SceneObjectBase<ActionBarSceneState> {
       <Box paddingY={0}>
         <div className={styles.actions}>
           <Stack gap={1}>
+            {
+              // @ts-ignore appSidecar not yet in stable runtime
+              config.featureToggles.appSidecar && <ToolbarExtensionsRenderer serviceScene={serviceScene} />
```
aocenas, line 46: Not sure this needs to be behind a featureFlag. The code itself does not have any special tie to sidecar, just the reason for adding it now is to test it.

src/Components/ServiceScene/ActionBarScene.tsx
matyax, line 45:
```suggestion
              // @ts-expect-error appSidecar not yet in stable runtime
```

matyax: If you mean having to add the ts-whatever directive, it's okay. We also do that for other features that haven't been released https://github.com/grafana/explore-logs/blob/main/src/Components/ServiceScene/LogsPanelScene.tsx#L122

matyax (approved):
Looks good, thank you @aocenas !

src/Components/ServiceScene/ActionBarScene.tsx
aocenas: Hmm actually I think the `@ts-expect-error` is wrong here (well also IDE says so :D). I assume the thing is the `config.featureToggles.appSidecar` does not error out in a regular sense, the `appSidecar` will just be `undefined`.
Screenshot 2024-10-30 at 16 36 33

matyax: When we get that it usually means that the feature flag has been released and is present in the type. In any case, it's pretty minor. Thanks for checking!

PR #829, opened 2024-10-16
Shard query splitting: limit group size to be less than the remaining shards

matyax:
After trying different alternative approaches, such as rate of change between the last 2 requests, or rate of change between the last query and the worst performing query, I settled on the current approach which works and is simple, with just an added constrain to the maximum group size, which now cannot be longer than a % of the remaining shards.

Before:

big big group

After:

smaller groups

Additionally:
- Fast responses less aggressively increase the group size (50% or 10%)
- Check for the validity of the selector before passing it to the language provider to prevent flashing errors
  flashing error


gtk-grafana (approved):
Much much better. I'm seeing the majority of requests take <5s, and the ones that do take 5-30 are usually just the final shard, or the empty shard. 

I would say we should get this into main/ops ASAP and figure out if we can get better sharding strategies in Loki so the empty/final shards aren't so heavy to query.

SHIP IT

PR #831, opened 2024-10-16
Patterns: Fix broken data link in pattern viz

gtk-grafana:
A few conflating problems here:
1. Looks like the onClick handler was excluded by https://github.com/grafana/grafana/pull/93950/, 
2. Also the event object changed, so even without the regression this still would have broken in 11.3

Fixes: https://github.com/grafana/explore-logs/issues/827
Requires: https://github.com/grafana/grafana/pull/94815/ or grafana <= 11.2.2

event.origin.labels.name appears before the regression as well, so it should be safe for all supported versions of Grafana


matyax:
> Requires: https://github.com/grafana/grafana/pull/94815 or 11.2.2

The "or" is confusing me, because with the latest Grafana it still doesn't work.

matyax (approved):

matyax:
Ahhhhhhhhh, right. Brainfarted hard.

PR #833, opened 2024-10-17
Patterns: fix flashing no patterns UI when loading

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/832


matyax (approved):
Much better! Thanks for the quick fix.

PR #840, opened 2024-10-18
feat(exploration): add `grafana-lokiexplore-app/metric-exploration/v1` entrypoint

svennergr:
This PR adds the `grafana-lokiexplore-app/metric-exploration/v1` entrypoint that will be used by the Explorations app. Without the app being installed, the entrypoint shouldn't change anything as `usePluginLinks` will return no links.

With [the app](https://github.com/grafana/investigations-app) those links will return add an icon button to the metrics panels for now.


src/Components/ServiceScene/Breakdowns/AddToExplorationButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+        const filter = getFilter(frame);
+        queries?.forEach((query: DataQuery & { legendFormat: string }) => {
+          if (filter) {
+            query.legendFormat = `{{${filter.name}}}`;
+          }
+        });
+      }
+    }, [frame, queries]);
+
+    const extensionPointId = 'grafana-lokiexplore-app/metric-exploration/v1';
```
matyax, line 73: I know this is a WIP, but this is a lot of work happening in the render body of the component. If you can extract some of these to functions and constants (like this extensionPointId) it would be great.

src/Components/ServiceScene/Breakdowns/AddToExplorationButton.tsx
```diff
@@ -0,0 +1,125 @@
...
+              disabled={link.category === 'disabled' || disabledLinks.includes(link.id)}
+              aria-label="extension-link-to-open-exploration"
+              key={link.id}
+              name={link.icon ?? 'panel-add'}
+              onClick={(e) => {
+                if (link.onClick) {
+                  link.onClick(e);
+                }
+                disabledLinks.push(link.id);
+                model.setState({ disabledLinks: disabledLinks });
```
matyax, line 109: This is mutating the current state array instead of creating a new one. This can lead to unexpected behaviors.

```suggestion
                disabledLinks.push(link.id);
                model.setState({ disabledLinks: [...disabledLinks] });
```
disabledLinks.slice() also works.

src/services/fields.ts
```diff
@@ -118,7 +119,8 @@ export function getFilterBreakdownValueScene(
   getTitle: (df: DataFrame) => string,
   style: DrawStyle,
   variableName: typeof VAR_FIELDS | typeof VAR_LABELS | typeof VAR_METADATA,
-  sortByScene: SortByScene
+  sortByScene: SortByScene,
+  labelKey?: string
```
matyax, line 123: Are we setting this anywhere? At least in this PR I don't see it.

matyax (approved):
Looks great!

src/services/fields.ts
svennergr: Should be these changes: https://github.com/grafana/explore-logs/pull/840/files#diff-e7101021dce31dbf3cbbcb5f789bf846d662a1a45562ca278dd01b3703f3ecb5

matyax: Perfect, thanks!

PR #848, opened 2024-10-22
Ad hoc variables: add support for detected_field/.../values

gtk-grafana:
Provides support for new detected_field/.../values endpoint in Loki 3.3.0 and Grafana 11.4 with graceful fallback support for older versions of both. Functionality should remain unchanged if the required versions are not met, so we should be able to release this safely before the dependencies.

Requires Grafana 11.4: 
* https://github.com/grafana/grafana/pull/95204
* https://github.com/grafana/grafana/pull/95608
Requires Loki 3.3.0: 
* https://github.com/grafana/loki/pull/14350


matyax:
I think one way to degrade gracefully will be to catch the error and resolve to an empty array, so we avoid this:

imagen

matyax:
Looking soon.

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -138,6 +148,56 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
...
+  }
+
+  private getFieldsTagValuesExpression(variableType: typeof VAR_FIELDS | typeof VAR_METADATA | typeof VAR_LEVELS) {
+    switch (variableType) {
+      case VAR_FIELDS:
+        return DETECTED_FIELD_VALUES_EXPR;
+      case VAR_METADATA:
+        return DETECTED_METADATA_VALUES_EXPR;
+      case VAR_LEVELS:
+        return DETECTED_LEVELS_VALUES_EXPR;
```
matyax, line 197: As usual, I would add a default case with a console error for good measure.

src/services/TagValuesProvider.ts
```diff
@@ -0,0 +1,103 @@
...
+      );
+      return { replace: true, values: [] };
+    }
+  } else {
+    logger.warn(
+      'getDetectedFieldValuesTagValuesProvider: fetchDetectedLabelValues is not defined in Loki datasource. Upgrade to Grafana 11.4 or higher.'
+    );
+    return { replace: true, values: [] };
+  }
+};
```
matyax, line 103: The amount of return statements made me a bit dizzy. Would you consider something like:

```suggestion
export const getDetectedFieldValuesTagValuesProvider = async (
  filter: AdHocFilterWithLabels,
  expr: string,
  sceneRef: SceneObject,
  timeRange: TimeRange,
  variable: typeof VAR_FIELDS | typeof VAR_METADATA | typeof VAR_LEVELS
): Promise<{
  replace?: boolean;
  values: MetricFindValue[];
}> => {
  const datasourceUnknownType = await getDataSourceSrv().get(getDataSource(sceneRef));
  // Narrow the DataSourceApi type to DataSourceWithBackend
  if (!(datasourceUnknownType instanceof DataSourceWithBackend)) {
    logger.error(new Error('getTagValuesProvider: Invalid datasource!'));
    throw new Error('Invalid datasource!');
  }

  // Assert datasource is Loki
  const lokiDatasource = datasourceUnknownType as DataSourceWithBackend<LokiQuery>;
  // Assert language provider is LokiLanguageProvider
  const languageProvider = lokiDatasource.languageProvider as LokiLanguageProviderWithDetectedLabelValues;

  let values: MetricFindValue[] = [];

  if (languageProvider && languageProvider.fetchDetectedLabelValues) {
    const options: FetchDetectedLabelValuesOptions = {
      expr,
      limit: 1000,
      timeRange,
      throwError: true,
    };

    const requestOptions: Partial<BackendSrvRequest> = {
      showErrorAlert: false,
    };

    try {
      let results = await languageProvider.fetchDetectedLabelValues(filter.key, options, requestOptions);
      // If the variable has a parser in the value, make sure we extract it and carry it over, this assumes the parser for the currently selected value is the same as any value in the response.
      // @todo is the parser always the same for the currently selected values and the results from detected_field/.../values?
      if (results && isArray(results)) {
        if (variable === VAR_FIELDS) {
          const valueDecoded = getValueFromFieldsFilter(filter, variable);
          values = results.map((v) => ({
            text: v,
            value: JSON.stringify({
              value: v,
              parser: valueDecoded.parser,
            }),
          }));
        } else {
          values = results.map((r) => ({ text: r }));
        }
      }
    } catch (e) {
      logger.error(e);
      logger.warn(
        'getDetectedFieldValuesTagValuesProvider: loki missing detected_field/.../values endpoint. Upgrade to Loki 3.3.0 or higher.'
      );
    }
  } else {
    logger.warn(
      'getDetectedFieldValuesTagValuesProvider: fetchDetectedLabelValues is not defined in Loki datasource. Upgrade to Grafana 11.4 or higher.'
    );
  }

  return { replace: true, values };
};
```

matyax (approved):
Re-tested it and looks great! Amazing work.

PR #849, opened 2024-10-22
Structured metadata: Changes to ad-hoc variable doesn't run detected_fields 

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/847


tests/exploreServicesBreakDown.spec.ts
```diff
@@ -403,19 +402,15 @@ test.describe('explore services breakdown page', () => {
...
-
     // Panel on the top should not
     await expect(page.getByTestId(/data-testid Panel header/).first()).not.toBeInViewport();
-
     // Wait for a bit for the requests to be made
     await page.waitForTimeout(250);
-
     // if this flakes we could just assert that it's greater then 3
-    expect(requestCount).toEqual(14);
+    expect(requestCount).toEqual(17);
```
gtk-grafana, line 413: Don't know why there is an extra row now, this shouldn't have anything to do with the functional change in this PR, I think it's a consequence of the updates to 11.3

tests/fixtures/explore.ts
```diff
@@ -82,7 +82,7 @@ export class ExplorePage {
   }
 
   async scrollToBottom() {
-    const main = this.page.locator('main#pageContent');
+    const main = this.page.locator('html');
```
gtk-grafana, line 85: scroll container appears to have changed in 11.3

tests/fixtures/explore.ts
```diff
@@ -44,7 +44,7 @@ export class ExplorePage {
   }
 
   async setDefaultViewportSize() {
-    await this.page.setViewportSize({ width: 1280, height: 720 });
+    await this.page.setViewportSize({ width: 1280, height: 680 });
```
gtk-grafana, line 47: window height also changed a bit in 11.3, which had the consequence of adding more lazy loaded elements into the viewport, bumping down the height a tiny bit so we don't have to update a bunch of assertions.

matyax:
I'm seeing that it correctly runs with filter changes except with patterns.

Uploading Test.mov…

gtk-grafana:
@matyax great catch, I was initially thinking patterns should be fine because the patterns endpoint doesn't take detected fields, but it turns out detected_fields does take patterns!

Should be fixed now

matyax (approved):
GJ!

PR #854, opened 2024-10-25
chore(intercept-banner): move into `container`

svennergr:
Moves the intercept banner into the main `container` so it gets the same margins as the rest of the app.
image


matyax:
Renders like this for some reason.

imagen

matyax:
Seems to be the flex grow attribute.

gtk-grafana:
@matyax good catch, I pushed a quick fix

matyax (approved):

PR #855, opened 2024-10-25
Logs panel: add button to copy link to log line

### LINKED ISSUE #772: Logs panel: Share link to log line
Allow users to share links to a specific log line within the logs panel.

matyax:
Requires https://github.com/grafana/grafana/pull/95330

This PR adds support for linking to log lines from the logs viz.

Closes #772

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -486,3 +486,4 @@ uuidv
 without
 pangea
 primarylabels
+Shortlink
--- a/src/Components/ServiceScene/CopyLinkButton.test.tsx
+++ b/src/Components/ServiceScene/CopyLinkButton.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
 import { CopyLinkButton } from './CopyLinkButton';
 import React from 'react';
 import userEvent from '@testing-library/user-event';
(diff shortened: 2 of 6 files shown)


matyax:
Testing coverage is pending, but I'd appreciate if you give it a try in the mean time.

gtk-grafana:
Looks pretty good, a couple of nits:
1. When copying a link in the logs panel there's no feedback that the link was copied, in classic Explore we show a toast on copy, in Explore Logs Table we show a tooltip
image
image

I prefer the tooltip, but if it's not possible with #95330 the toast will do for now, but we should create an issue to sync up the UIs if this is the case.

2. Interoperability with the table link: When switching back and forth between panels the links do not behave as expected, each panel has an independent line link. Ideally if you have a link to a log line in one viz, and a user switches to the other viz, that line should still be highlighted in the other viz. However this can be addressed in a new issue/PR and should not block release of this feature (feel free to assign to me). 

Otherwise LGTM, and thanks for cleaning things up!

matyax:
Thanks for the feedback! For point 2, I was thinking that we should update the table to use the `panelState` param that the viz, Dashboards, and Explore use. https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/logs/LogsPanel.tsx#L448-L450

Created https://github.com/grafana/explore-logs/issues/857

matyax:
Addressed point 1:

imagen

There's no `success` variant for `IconButton` but it does the trick.

gtk-grafana (approved):
Works locally, great work as always

gtk-grafana:
Oh, except I didn't check that this works with older Grafana versions, will this fail gracefully?

xenartro:
Yeah, it would just not render. On the bright side, whenever permalink support was added to LogsPanel, any URL that contains the parameter should at least scroll to the linked log line for free.

PR #868, opened 2024-11-04
ServiceSelectionScene: Manual query runners

gtk-grafana:
First step of breaking https://github.com/grafana/explore-logs/pull/841 into manageable chunks.

This PR does two things:
1. Converts the service selection scene query runners to "manual" `runQueriesMode`, and executes queries on activation, datasource, and time range updates.
2. Tears down the Volume query runner when updating the aggregated metric toggle. This is required to prevent duplicate queries from being run as the entire body is rebuilt, triggering activation for the existing panels.

Why:
Preparing for https://github.com/grafana/explore-logs/pull/841, we will need to execute panel queries manually to avoid duplicate queries, as we'll be adding the ability for users to change the label filters, which is used in the logs panel and volume queries, since the volume queries trigger the panel queries, and changing the labels triggers volume and panel queries as that variable is interpolated in both, we need to be able to skip executing the panel queries every time the labels change.

When:
Slated for merge to main 1.0.4 after 1.0.3 release which should go out early this week.


src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -677,7 +730,25 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
     }
   }
 
-  private updateBody() {
+  private getGridItems(): SceneCSSGridItem[] {
+    return this.state.body.state.children as SceneCSSGridItem[];
+  }
+
+  private getVizPanel(child: SceneCSSGridItem) {
+    return child.state.body as VizPanel | undefined;
```
matyax, line 738:
```suggestion
    return child.state.body instanceof VizPanel ? child.state.body  : undefined;
```

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -677,7 +730,25 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
...
+  }
+
+  private runPanelQuery(child: SceneCSSGridItem) {
+    if (child && child.isActive) {
```
matyax, line 742: If you need to test for child you should update the type in the function.

matyax (commented):
Looks good! Left a few comments.

matyax (approved):

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
gtk-grafana: Avoids the type assertion too. Good catch and fixed

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
gtk-grafana: Agreed, fixed

PR #871, opened 2024-11-04
Detected fields: Use detected_fields response to determine if avg_over_time query should be run

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/865


src/services/fields.ts
```diff
@@ -77,6 +78,23 @@ export function extractParserFromString(parserString?: string): ParserType {
...
+    case 'float':
+      return 'float';
+    case 'duration':
+      return 'duration';
+    case 'boolean':
+      return 'boolean';
+    case 'bytes':
+      return 'bytes';
+    default:
+      return 'string';
```
matyax, line 94:
```suggestion
    case 'int':
    case 'float':
    case 'duration':
    case 'boolean':
    case 'bytes':
      retyrb fieldString;
    default:
      return 'string';
```

matyax (approved):

PR #878, opened 2024-11-05
Labels variable: Combobox

gtk-grafana:
Second step of breaking https://github.com/grafana/explore-logs/pull/841 into more manageable chunks.

This PR converts the ad-hoc variables to the new `combobox` type. 
This also allows users to select multiple include filters on the breakdown views using the combobox. 
The combobox has not been added to the service selection scene, that will happen in the next PR


src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -748,7 +757,19 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
       const queryRunners = getQueryRunnerFromChildren(child);
       if (queryRunners.length === 1) {
         const queryRunner = queryRunners[0];
-        queryRunner.runQueries();
+
+        // If the scene was cached, the time range will still be the same as what was executed in the query
+        const requestTimeRange = queryRunner.state.data?.timeRange;
+        const sceneTimeRange = sceneGraph.getTimeRange(this);
+        const fromDiff = requestTimeRange
```
matyax, line 764: Could this diff be negative? Should be `Math.abs()` this?

src/services/TagKeysProviders.ts
```diff
@@ -0,0 +1,36 @@
...
+  const datasource = datasource_ as DataSourceWithBackend<LokiQuery>;
+
+  if (datasource && datasource.getTagKeys) {
+    const filters = joinTagFilters(variable);
+
+    const options: DataSourceGetTagKeysOptions<LokiQuery> = {
+      filters,
+    };
+
+    const result = (await datasource.getTagKeys(options)) as MetricFindValue[];
```
matyax, line 28:
```suggestion
    const tagKeys = await datasource.getTagKeys(options);
    const result: MetricFindValue[] = Array.isArray(tagKeys) ? tagKeys : [];
```
This seems to do the trick. A bit less tech debt for when we're tasked to remove type assertions.

matyax:
Not sure if this has been discussed, but if we're adding the "Indexed" prefix, I would very much rather change the tab name to "Indexed labels" and keep "Labels" in the filters.

imagen

imagen

Or use "Indexed" in both.

matyax:
I'm not getting options for the Combobox but I have a value for container in the breakdown.

imagen

matyax (approved):
Looks great and seems to be working as expected. Amazing addition!

src/services/query.test.ts
```diff
@@ -232,3 +233,67 @@ describe('renderLogQLLabelFilters', () => {
     );
   });
 });
+
+describe('joinTagFilters', () => {
+  it('joins multiple include', () => {
+    const adHoc = new AdHocFiltersVariable({
```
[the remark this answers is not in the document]
gtk-grafana: variable.state.filters currently doesn't have regex operators like:

```
 {
        key: 'service_name',
        value: 'service_value|service_value_2',
        operator: '=~',
      },
```

Although we should look into adding support for more operators and multiple values in the combobox in a later PR

PR #879, opened 2024-11-06
Shard splitting: improve combineResponses performance

matyax:
We recently noticed a performance degradation in the field breakdown pages, for some services, which was narrowed down to the `shouldCombine` function that was doing a lot of extra work for something that should have been only comparing the data frame refId.

Measuring `combineResponses()` in ms.

Before:

Before

After:

After


matyax:
The fix is incorrect as it doesn't support multiple series, but the problem has been narrowed down to the `shouldCombine` function, and in particular the successive `shallowCompare` calls.

PR #880, opened 2024-11-07
feat(combineResponses): improve label comparison performance

matyax:
This PR adds a performance optimization to the label comparison method, which is a function that gets called very often in the recombination code.

## Measurements before

1 minute of profiling

Before 2

Slightly different method but still comparing label keys.

Before 1

Start and end of `mergeResponses`
Worst cases original

## Measurements after

1 minute of profiling

After

Start and end of `mergeResponses`
Worst cases after

### DIFF
--- a/src/services/combineResponses.ts
+++ b/src/services/combineResponses.ts
@@ -7,7 +7,6 @@ import {
   Field,
   FieldType,
   QueryResultMetaStat,
-  shallowCompare,
 } from '@grafana/data';
 import { logger } from './logger';
 
@@ -180,10 +179,6 @@ function findSourceField(referenceField: Field, sourceFields: Field[], index: nu
     return candidates[0];
   }
 
-  if (referenceField.labels) {
-    return candidates.find((candidate) => shallowCompare(referenceField.labels ?? {}, candidate.labels ?? {}));
-  }
-
   return sourceFields[index];
 }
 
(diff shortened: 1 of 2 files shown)


src/services/combineResponses.ts
```diff
@@ -180,10 +179,6 @@ function findSourceField(referenceField: Field, sourceFields: Field[], index: nu
     return candidates[0];
   }
```
matyax, line 181: This was taken from the original grafana/grafana implementation, for non-Loki data sources. https://github.com/grafana/grafana/pull/90650

src/services/combineResponses.ts
```diff
@@ -272,3 +263,19 @@ function shouldCombine(frame1: DataFrame, frame2: DataFrame): boolean {
...
+
+function compareLabels(frame1: DataFrame, frame2: DataFrame) {
+  const field1 = frame1.fields.find((f) => f.type === FieldType.number);
+  const field2 = frame2.fields.find((f) => f.type === FieldType.number);
+  if (field1 === undefined || field2 === undefined) {
+    // should never happen
+    return false;
+  }
+  if (!frame1.name) {
+    frame1.name = JSON.stringify(field1.labels);
```
gtk-grafana, line 275: this is a **bunch** faster so let's ship this ASAP, but I wonder if we could optimize this further by avoiding serializing the labels. Maybe by using maps and adding labels one by one from each  object and returning false as soon as we find something that doesn't match, and then generating a uuid for the name for fields that do match?

gtk-grafana (approved):
So much better. Great work @matyax

src/services/combineResponses.ts
matyax: > Maybe by using maps and adding labels one by one from each object and returning false as soon as we find something 

Unless I don't understand the suggestion, that sounds essentially equal to what we were previously doing with shallowCompare https://github.com/grafana/grafana/blob/e86283e8b173b328c4710720b8f807f359f57a77/packages/grafana-data/src/dataframe/frameComparisons.ts#L102, and even with an slightly simplified shallowCompare that I implemented for this.

PR #881, opened 2024-11-07
Service selection: Label selection UI

gtk-grafana:
Third step of breaking https://github.com/grafana/explore-logs/pull/841 into more manageable chunks.
Parent branch: https://github.com/grafana/explore-logs/pull/878

WIP: Needs to implement updated designs from Joan

### DIFF
--- a/src/Components/ServiceSelectionScene/AddLabelToFiltersHeaderActionScene.tsx
+++ b/src/Components/ServiceSelectionScene/AddLabelToFiltersHeaderActionScene.tsx
@@ -1,0 +1,85 @@
+import { SceneComponentProps, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
+import { Button, useStyles2 } from '@grafana/ui';
+import { testIds } from '../../services/testIds';
+import { addToFilters, FilterType } from '../ServiceScene/Breakdowns/AddToFiltersButton';
+import { VAR_LABELS } from '../../services/variables';
+import { getLabelsVariable } from '../../services/variableGetters';
+import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from '../../services/analytics';
+import { GrafanaTheme2 } from '@grafana/data';
+import { css } from '@emotion/css';
+import React from 'react';
+
+export interface AddLabelToFiltersHeaderActionSceneState extends SceneObjectState {
+  name: string;
+  value: string;
+}
+
+export class AddLabelToFiltersHeaderActionScene extends SceneObjectBase<AddLabelToFiltersHeaderActionSceneState> {
+  constructor(state: AddLabelToFiltersHeaderActionSceneState) {
+    super({
+      ...state,
+    });
+  }
+
+  public static Component = ({ model }: SceneComponentProps<AddLabelToFiltersHeaderActionScene>) => {
+    const { value } = model.useState();
+    const isIncluded = false;
+    const isExcluded = false;
+
+    const styles = useStyles2(getStyles, isIncluded, isExcluded);
+    return (
+      <span className={styles.wrapper}>
+        <Button
+          tooltip={`Add ${value} to filters`}
+          variant={isIncluded ? 'primary' : 'secondary'}
+          fill={'outline'}
+          icon={'plus'}
+          size="sm"
+          aria-selected={isIncluded}
+          className={styles.includeButton}
+          onClick={() => (isIncluded ? () => model.onClick('clear') : model.onClick('include'))}
+          data-testid={testIds.exploreServiceDetails.buttonFilterInclude}
+        />
+      </span>
+    );
+  };
+
+  public getFilter() {
+    return { name: this.state.name, value: this.state.value };
+  }
+
+  public onClick = (type: FilterType) => {
+    const filter = this.getFilter();
+    if (!filter) {
+      return;
+    }
+
+    addToFilters(filter.name, filter.value, type, this, VAR_LABELS);
+
+    const variable = getLabelsVariable(this);
+    reportAppInteraction(USER_EVENTS_PAGES.service_selection, USER_EVENTS_ACTIONS.service_selection.add_to_filters, {
+      filterType: 'index-filters',
+      key: filter.name,
+      action: type,
+      filtersLength: variable?.state.filters.length || 0,
+    });
+  };
+}
+
+const getStyles = (theme: GrafanaTheme2, isIncluded: boolean, isExcluded: boolean) => {
+  return {
+    container: css({
+      display: 'flex',
+      justifyContent: 'center',
+    }),
+    includeButton: css({
+      borderRadius: 0,
+    }),
+    wrapper: css({
+      display: 'flex',
+      flexDirection: 'column',
+      justifyContent: 'center',
+      alignSelf: 'center',
+    }),
+  };
+};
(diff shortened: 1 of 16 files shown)


src/Components/IndexScene/CustomVariableValueSelectors.tsx
```diff
@@ -0,0 +1,43 @@
...
+  includeNames?: string[];
+  excludeNames?: string[];
+}
+
+export class CustomVariableValueSelectors extends SceneObjectBase<VariableValueSelectorsState> {
+  public static Component = CustomVariableValueSelectorsRenderer;
+}
+
+function CustomVariableValueSelectorsRenderer({ model }: SceneComponentProps<CustomVariableValueSelectors>) {
+  const variablesSetState = sceneGraph.getVariables(model)!.useState();
```
svennergr, line 22: I'm always a bit scared seeing `!`. Can we fail gracefully if not defined, e.g. just setting `variables` to an empty error.

src/Components/IndexScene/LayoutScene.tsx
```diff
@@ -169,8 +216,10 @@ function getStyles(theme: GrafanaTheme2) {
       },
     }),
     controlsWrapper: css({
+      label: 'controlsWrapper',
       display: 'flex',
       flexDirection: 'column',
+      marginTop: '3px',
```
svennergr, line 222: Can we do `theme.spacing(1)` or the equivalent here?

src/Components/IndexScene/ShowLogsButtonScene.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { GrafanaTheme2 } from '@grafana/data';
+import { css } from '@emotion/css';
+import { navigateToInitialPageAfterServiceSelection } from '../../services/navigate';
+import { getLabelsVariable } from '../../services/variableGetters';
+import { FilterOp } from '../../services/filterTypes';
+import { testIds } from '../../services/testIds';
+
+export interface ShowLogsButtonSceneState extends SceneObjectState {
+  disabled?: boolean;
+  hide?: boolean;
```
svennergr, line 13: Just for consistency, `disabled` is past tense, `hide` isn't. Should we change to `hidden`?

src/Components/IndexScene/ShowLogsButtonScene.tsx
```diff
@@ -0,0 +1,80 @@
...
+      ...state,
+    });
+
+    this.addActivationHandler(this.onActivate.bind(this));
+  }
+
+  onActivate() {
+    const labelsVar = getLabelsVariable(this);
+    labelsVar.subscribeToState((newState) => {
+      const hasPositiveFilter = !!newState.filters.find((f) => f.operator === FilterOp.Equal);
```
svennergr, line 27:
```suggestion
      const hasPositiveFilter = newState.filters.some((f) => f.operator === FilterOp.Equal);
```

svennergr (commented):
https://github.com/user-attachments/assets/d616eca7-f72b-4168-994d-044eb1e43abf

The counter in the tabs doesn't seem to update accordingly.

Left a few nits, but overall it's working great!

src/Components/IndexScene/CustomVariableValueSelectors.tsx
gtk-grafana: I'll take a look, this was lifted directly from scenes though: https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/components/VariableValueSelectors.tsx#L22

gtk-grafana: Nevermind, seems like it wasn't needed. Removed!

svennergr (commented):
This state doesn't reset correctly.


https://github.com/user-attachments/assets/fa320feb-089d-4e56-931c-d063ef016f8a

gtk-grafana:
> The counter in the tabs doesn't seem to update accordingly.

So it looks like allowing multiple includes in the UI conflicts with the tab counts.
For example I have an include filter in tab A, the volume query we run to get the number of panels will replace that value with .+ so we display all possible labels for the user to potentially filter on. If we run detected_labels with the same query, we'll get the correct counts for this and all tabs.
Then I switch to tab B, the label the volume query renders for the label in tab A is no longer `.+`, otherwise users could include a label that is mutually exclusive with the filter in tab A, dead-ending the user and returning no data for every future query. Now in the count for tab A, we'll show the number of include filters we've added, instead of the number of possible labels the user can select, so when on tab B, the label for tab A would be 1, but when on tab A the count would be n (the cardinality of that label before any filters are applied).

Since there's not an un-confusing way to let users know that the counts used in the tabs/dropdowns are only applicable before any filters are applied, we decided to remove the counts for now, which admittedly sucks, but is arguably better then confusing users with incorrect counts. :(

src/Components/ServiceSelectionScene/AddLabelToFiltersHeaderActionScene.tsx
```diff
@@ -0,0 +1,117 @@
...
+      const value = getValueFromAdHocVariableFilter(variable, f);
+      return f.key === this.state.name && value.value === this.state.value;
+    });
+
+    if (!filterInSelectedFilters) {
+      return { isIncluded: false, isExcluded: false };
+    }
+
+    return {
+      isIncluded: filterInSelectedFilters.operator === FilterOp.Equal,
```
matyax, line 54: Can something be included and excluded at the same time? If not, I would recommend changing this to a boolean or enum.
Can also be `included: boolean | null`. null === no state, included true, excluded false.

matyax (commented):
Looks and works perfectly, really nice work! Left a few minor comments around.

Just one thing that happened to me, that might happen to other people. When I clicked the "add to filters" buttons, I was confused for a minute expecting something to happen in the top section. Even with context about this feature, I was used to seeing filters being appended to the top, and the only change was the button becoming primary and enabled.

https://github.com/user-attachments/assets/b772e5cd-5bd6-40c7-8641-c09195e19ab2

For a follow up, can we consider adding the current filters as text, or at least as a tooltip in the button to make it more evident that you're about to show logs for the selected filters?

gtk-grafana:
> When I clicked the "add to filters" buttons, I was confused for a minute expecting something to happen in the top section. Even with context about this feature, I was used to seeing filters being appended to the top, and the only change was the button becoming primary and enabled.

@matyax yeah that's my bad, I had forgot to push my most recent commit which addressed exactly this, you've got aggregated metrics enabled, and we don't currently support more then a single label filter with agg metrics. We removed all those buttons when agg metrics is enabled.

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -164,6 +182,19 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
 
   private setTagProviders() {
     const labelsVar = getLabelsVariable(this);
+
+    labelsVar._getOperators = function () {
+      const wip = labelsVar.state._wip;
```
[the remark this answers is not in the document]
gtk-grafana: wip is the `AdHocFilterWithLabels` filter the user is currently editing, part of the `AdHocFiltersVariableState`

gtk-grafana:
Nope, please hold, I introduced a bug in the last commit.

matyax:
Oh, right. I'll disable it.

matyax (approved):
🚀

svennergr (approved):
👍

PR #885, opened 2024-11-12
Go to Explore button: keep visual preferences in Explore link

### LINKED ISSUE #884: Open in Explore button: respect displayed fields
As an Explore Logs user using displayed fields I would like to keep them when I open my current search in Explore.

matyax:
This PR extends the GoToExploreButton to respect the selected displayedFields, visualizationType, and table columns, which was a source of confusion for users switching between apps.

Part of #884
Requires https://github.com/grafana/grafana/pull/96242

### DIFF
--- a/src/Components/ServiceScene/ActionBarScene.tsx
+++ b/src/Components/ServiceScene/ActionBarScene.tsx
@@ -42,7 +42,7 @@ export class ActionBarScene extends SceneObjectBase<ActionBarSceneState> {
         <div className={styles.actions}>
           <Stack gap={1}>
             {config.featureToggles.appSidecar && <ToolbarExtensionsRenderer serviceScene={serviceScene} />}
-            <GoToExploreButton exploration={exploration} />
+            <GoToExploreButton exploration={exploration} sceneRef={serviceScene} />
           </Stack>
         </div>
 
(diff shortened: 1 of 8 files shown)


gtk-grafana:
@matyax Does this rely on core changes for displayed fields?
Also be good to create a follow-up PR for table to carry over any columns the user added in Explore Logs to Explore

matyax:
Yes, requires https://github.com/grafana/grafana/pull/96242 on the Grafana side. Without it, should silently do nothing.

> Also be good to create a follow-up PR for table to carry over any columns the user added in Explore Logs to Explore

Definitely, good idea. I'll add it to #884

gtk-grafana:
Well I verified that it silently does nothing correctly :P. I'll continue the review tomorrow.

gtk-grafana:
LGTM, but same worries here as in https://github.com/grafana/grafana/pull/96242,

Would be nice to show the "showing only selected fields" message more prominently then in the tool tip under the button in the top right corner which I imagine will get missed, potentially leading to confusion for users.

image
image

Might be good to review with UX before release

gtk-grafana:
Another idea, add a sticky positioned "pill" for each displayed field to the top of the logs panel which can be cleared out individually as well as having the button that clears them all.

svennergr:
Was wondering if we can use something similar to error popup per line, if lines are empty due to "wrong" selected fields.
image

Just throwing out ideas.

matyax:
Changing the Explore button in https://github.com/grafana/grafana/pull/96242 seems like a no brainer, let's do it.

For the other comments, I'm having a hard time understanding the relation with this PR which is meant to respect the selected fields in the Explore links. The query, time range, and selected fields in combination will asure that Explore will display the correct fields. After being in Explore, if the user changes the query, the displayed fields will be cleared out.

matyax:
Now also sending the "columns" param to Explore.

PR #892, opened 2024-11-14
Panel Menus

gtk-grafana:
Waiting on https://github.com/grafana/grafana/issues/96865 before taking out of draft, I can remove the css hacks and wrapping element if that proposal is accepted. 

Fixes: https://github.com/grafana/explore-logs/issues/890

Adds VizPanelMenu to all time-series
Adds link to explore to VizPanelMenu
Adds support for addToExplorations feature in VizPanelMenu

TODO:
* Do we want menu in logs/table panel?
* Do we want investigations link in LogsVolume and/or logs/table panel?
* Add to dashboard is not yet provided by core to plugins, but work was started in https://github.com/grafana/grafana/pull/94458

With investigations enabled:
image

Without investigations enabled: 
image


src/Components/Panels/ExploreLogsVizPanelMenu.tsx
```diff
@@ -0,0 +1,192 @@
...
+export const getPanelWrapperStyles = (theme: GrafanaTheme2) => {
+  return {
+    panelWrapper: css({
+      width: '100%',
+      height: '100%',
+      label: 'panel-wrapper',
+      display: 'flex',
+
+      // Need more specificity to override core style
+      'button.show-on-hover': {
```
gtk-grafana, line 182: Might want to add e2e test asserting that the menus are visible without hovering. Also might be a good idea to open a core issue/PR to allow consumers of VizPanelMenu to define if the menu is always visible, or only visible on hover

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
```diff
@@ -0,0 +1,194 @@
...
+        addToExplorations: new AddToExplorationButton({
+          labelName: this.state.labelName,
+          fieldName: this.state.fieldName,
+          frame: this.state.frame,
+        }),
+      });
+
+      // @todo rewrite the AddToExplorationButton
+      // Manually activate scene
+      this.state.addToExplorations?.activate();
```
svennergr, line 54: 👍  I can take a GH issue for that.

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
```diff
@@ -0,0 +1,194 @@
...
+  frame?: DataFrame;
+  labelName?: string;
+  fieldName?: string;
+  addToExplorations?: AddToExplorationButton;
+}
+
+/**
+ * @todo the VizPanelMenu interface is overly restrictive, doesn't allow any member functions on this class, so everything is currently inlined
+ */
+export class ExploreLogsVizPanelMenu
```
svennergr, line 37: I think we don't need the prepended "ExploreLogs".

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+      // @todo rewrite the AddToExplorationButton
+      // Manually activate scene
+      this.state.addToExplorations?.activate();
+
+      const items: PanelMenuItem[] = [
+        {
+          text: 'Explore',
+          iconClassName: 'compass',
+          shortcut: '',
```
svennergr, line 60: If it works, we should do `p x` to be similar to dashboards.

svennergr (commented):
Works good! I really like that.

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
gtk-grafana: I changed my mind, hopefully https://github.com/grafana/grafana/issues/96865 makes progress soon and this functionality will be tested in core, where it should be

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
gtk-grafana: then it conflicts with VizPanelMenu from scenes core

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
gtk-grafana: it does not work :(

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
svennergr: WDYT about just `PanelMenu`? Considering it's in the `Panel` directory, `Menu` might work too.

gtk-grafana: Nice. Done

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
```diff
@@ -0,0 +1,194 @@
...
+          onClick: () => onExploreClick(this),
```
svennergr: Requires some changes, but can we change this from `onClick` to `href`, i.e. change `onExploreLinkClick` to return a string instead?

That would mirror core functionality, and it would allow people to copy a link and CMD+Click on the link to topen in a new tab.

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
```diff
@@ -0,0 +1,194 @@
...
+    if (this.state.body) {
+      this.state.body.setItems(items);
+    }
+  }
+
+  public static Component = ({ model }: SceneComponentProps<ExploreLogsVizPanelMenu>) => {
+    const { body } = model.useState();
+
+    if (body) {
+      return body && <body.Component model={body} />;
```
svennergr:
```suggestion
      return <body.Component model={body} />;
```

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+const onExploreClick = (sceneRef: SceneObject) => {
+  const indexScene = sceneGraph.getAncestor(sceneRef, IndexScene);
+  const $data = sceneGraph.getData(sceneRef);
+  let queryRunner = getQueryRunnerFromChildren($data)[0];
+
+  // If we don't have a query runner, then our panel is within a SceneCSSGridItem, we need to get the query runner from there
+  if (!queryRunner) {
+    const sceneGridItem = sceneGraph.getAncestor(sceneRef, SceneCSSGridItem);
+    const queryProvider = sceneGraph.getData(sceneGridItem);
```
svennergr: AFAIK you can do `sceneGraph.getData(sceneRef)` and it will walk up the tree to find the next SceneDataProvider. With other words, no need to do `getAncestor`

src/Components/Panels/ExploreLogsVizPanelMenu.tsx
gtk-grafana: Updated, this should work here as the links are generated on activation of the menu (when the menu is opened), but it won't work for the existing explore button

svennergr: Yep, I know, looked into that before. I'll take another look in a bit 👍

config/loki-config.yaml
```diff
@@ -21,7 +21,6 @@ frontend:
 pattern_ingester:
   enabled: true
   metric_aggregation:
-    enabled: true
```
svennergr, line 24: huh?

docker-compose.local.yaml
```diff
@@ -22,7 +22,6 @@ services:
     extra_hosts:
       - 'host.docker.internal:host-gateway'
   loki:
-#    image: grafana/loki:main-12cd12a
```
svennergr, line 25: FWIW. not sure if intended.

src/Components/Panels/PanelMenu.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+const getExploreLink = (sceneRef: SceneObject) => {
+  const indexScene = sceneGraph.getAncestor(sceneRef, IndexScene);
+  const $data = sceneGraph.getData(sceneRef);
+  let queryRunner = getQueryRunnerFromChildren($data)[0];
+
+  // If we don't have a query runner, then our panel is within a SceneCSSGridItem, we need to get the query runner from there
+  if (!queryRunner) {
+    const sceneGridItem = sceneGraph.getAncestor(sceneRef, SceneCSSGridItem);
+    const queryProvider = sceneGraph.getData(sceneGridItem);
```
svennergr, line 104: Did this get lost?
https://github.com/grafana/explore-logs/pull/892#discussion_r1858648460

svennergr (approved):
The "Explore" link on distinct values does not include that value filter. Not really something for this PR, but good followup.

src/Components/Panels/PanelMenu.tsx
gtk-grafana: Nope, that change is what introduced the bugs on label value panels so I reverted

docker-compose.local.yaml
gtk-grafana: Yeah it was intended, although it could be moved to another PR

src/Components/Panels/PanelMenu.tsx
svennergr: Interesting, sounds good. Let's merge!

PR #908, opened 2024-11-20
Fix favoriting on label select

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/905
Fixes: https://github.com/grafana/explore-logs/issues/904

Should add any label to favorites when adding any ad hoc filter in the UI except for in the combobox.

Also fixes a bug with the service selection search where a custom filter would be actively filtering the labels returned in the UI, but the search input was empty and was not able to be cleared.

Now whenever labels are added (besides directly in the combobox), that label is favorited, and pushed to the top of the dropdowns, and the top of the service list on the next render (I don't think we want to change the order of the service selection list as users are adding filters).

I wasn't happy with subscribing to the combobox and favoriting whenever something was added as we didn't have the context needed to re-render components after adding things to local storage, adding full combobox support could be solved by subscribing in the ServiceScene, and ServiceSelection scene, and avoiding the re-rendering in the ServiceScene since nothing in the UI displays the favorites. 

We might want to refactor the components that consume local storage to listen to events or subscribe to some top-level state for re-rendering to get this in a more future proof state.

Need to follow up with addressing adding [combobox filters to favorites in another PR](https://github.com/grafana/explore-logs/issues/909)


src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -43,12 +44,17 @@ export function addToFilters(
   value: string,
   operator: FilterType,
   scene: SceneObject,
-  variableType?: VariableFilterType
+  variableType?: VariableFilterType,
+  skipFavorites?: boolean
```
gtk-grafana, line 48: This is gross, I'm adding implementation to prevent from ripping out more unit tests.

matyax: Definitely. Let's not do this.

matyax: You can mock '../../../services/store', doesn't that work?

src/services/store.ts
```diff
@@ -28,8 +31,35 @@ export function getFavoriteLabelValuesFromStorage(dsKey: string | unknown, label
   return labelValues;
 }
```
matyax, line 33: These are not "store" functions, right? Feel free to create a new service.

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -1092,3 +1083,17 @@ function getStyles(theme: GrafanaTheme2) {
     }),
   };
 }
+
+export function wrapWildcardSearch(input: string) {
```
matyax, line 1087: Maybe these can be `service/query.ts` functions?

matyax (commented):
Fixes the bug, but let's clean this up a bit.

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
gtk-grafana: Yeah good call, I was running on fumes last night. Fixed!

src/services/store.ts
gtk-grafana: Created a favorites service

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
gtk-grafana: Works for me. Done!

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -44,11 +46,16 @@ export function addToFilters(
   operator: FilterType,
   scene: SceneObject,
   variableType?: VariableFilterType
+  // skipFavorites?: boolean
```
matyax, line 49:
```suggestion
```

src/services/store.ts
```diff
@@ -52,8 +52,8 @@ export function addToFavoriteLabelValueInStorage(dsKey: string | unknown, labelN
   localStorage.setItem(key, JSON.stringify(servicesToStore));
 }
 
-export function removeFromFavoritesInStorage(dsKey: string, labelName: string, labelValue: string) {
-  if (!dsKey || !labelName || !labelValue) {
+export function removeFromFavoritesInStorage(dsKey: string | unknown, labelName: string, labelValue: string) {
```
matyax, line 55: Is it unknown or undefined? Seems unexpected that it could be unknown.

matyax (approved):
Looks good!

src/services/store.ts
gtk-grafana: yeah good catch, I was looking at the other functions in the store that do this, but it really should be a `VariableValue` instead (updated)

PR #920, opened 2024-11-25
Logs Panel: move log panel options and add sort order

gtk-grafana:
Moving Log panel options to the panel header, and adding logs direction:
image

WARNING: Breaking change for existing log wrap setting in local storage!

### DIFF
--- a/src/Components/ServiceScene/LogOptionsScene.tsx
+++ b/src/Components/ServiceScene/LogOptionsScene.tsx
@@ -22,13 +22,10 @@ export class LogOptionsScene extends SceneObjectBase<LogOptionsState> {
   static Component = LogOptionsRenderer;
 
   constructor(state: LogOptionsState) {
-    const logsSortOrderFromLocalStorage = getLogOption('sortOrder');
     super({
       ...state,
-      sortOrder: (logsSortOrderFromLocalStorage
-        ? logsSortOrderFromLocalStorage
-        : LogsSortOrder.Descending) as LogsSortOrder,
-      wrapLogMessage: Boolean(getLogOption('wrapLogMessage')),
+      sortOrder: getLogsPanelSortOrder(),
+      wrapLogMessage: Boolean(getLogOption<boolean>('wrapLogMessage', false)),
     });
   }
 
@@ -115,6 +112,10 @@ function LogOptionsRenderer({ model }: SceneComponentProps<LogOptionsScene>) {
       <LogsPanelHeaderActions vizType={visualizationType} onChange={onChangeVisualizationType} />
     </div>
   );
+}
+
+export function getLogsPanelSortOrder() {
+  return getLogOption<LogsSortOrder>('sortOrder', LogsSortOrder.Descending) as LogsSortOrder;
 }
 
 const getStyles = (theme: GrafanaTheme2) => ({
--- a/src/Components/ServiceScene/LogOptionsScene.tsx
+++ b/src/Components/ServiceScene/LogOptionsScene.tsx
@@ -90,7 +90,6 @@ function LogOptionsRenderer({ model }: SceneComponentProps<LogOptionsScene>) {
           ]}
           value={sortOrder}
           onChange={model.onChangeLogsSortOrder}
-          className={styles.buttonGroup}
         />
       </InlineField>
 
@@ -99,7 +98,6 @@ function LogOptionsRenderer({ model }: SceneComponentProps<LogOptionsScene>) {
           size="sm"
           value={wrapLinesText}
           onChange={model.handleWrapLinesChange}
-          className={styles.buttonGroup}
           options={[
             {
               label: 'Wrap',
@@ -126,18 +124,8 @@ const getStyles = (theme: GrafanaTheme2) => ({
     gap: theme.spacing(1),
     marginTop: theme.spacing(0.5),
   }),
-  input: css({
-    width: '100%',
-  }),
-  field: css({
-    label: 'field',
-    marginBottom: 0,
-  }),
   buttonGroupWrapper: css({
     margin: 0,
     alignItems: 'center',
-  }),
-  buttonGroup: css({
-    // padding: `0 4px 0 0`,
   }),
 });
--- a/src/Components/ServiceScene/LogOptionsScene.tsx
+++ b/src/Components/ServiceScene/LogOptionsScene.tsx
@@ -62,11 +62,17 @@ function LogOptionsRenderer({ model }: SceneComponentProps<LogOptionsScene>) {
   const { wrapLogMessage, onChangeVisualizationType, visualizationType, sortOrder } = model.useState();
   const { displayedFields } = model.getParentScene().useState();
   const styles = useStyles2(getStyles);
-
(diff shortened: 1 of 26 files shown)


src/Components/ServiceScene/LogOptionsScene.tsx
```diff
@@ -42,40 +59,85 @@ export class LogOptionsScene extends SceneObjectBase<LogOptionsState> {
...
     marginBottom: 0,
   }),
-  horizontalInlineSwitch: css({
-    padding: `0 4px 0 0`,
+  buttonGroupWrapper: css({
+    margin: 0,
+    alignItems: 'center',
+  }),
+  buttonGroup: css({
+    // padding: `0 4px 0 0`,
```
gtk-grafana, line 141: clean up

src/Components/Table/LogsHeaderActions.tsx
```diff
@@ -33,12 +38,11 @@ export function LogsPanelHeaderActions(props: {
 
 const styles = {
   visualisationType: css({
-    display: 'flex',
-    flex: '1',
-    justifyContent: 'space-between',
-    marginTop: '8px',
+    // display: 'flex',
```
gtk-grafana, line 41: clean up

src/Components/ServiceScene/LogOptionsScene.tsx
```diff
@@ -1,30 +1,47 @@
...
-  constructor(state?: Partial<LogOptionsState>) {
+  constructor(state: LogOptionsState) {
+    const logsSortOrderFromLocalStorage = getLogOption('sortOrder');
     super({
       ...state,
-      wrapLines: Boolean(getLogOption('wrapLines')),
+      sortOrder: (logsSortOrderFromLocalStorage
+        ? logsSortOrderFromLocalStorage
+        : LogsSortOrder.Descending) as LogsSortOrder,
+      wrapLogMessage: Boolean(getLogOption('wrapLogMessage')),
```
gtk-grafana, line 31: This is the breaking change for line wrap, not really necessary, but it's nice to have the options here consistent with the types.

src/Components/ServiceScene/LogOptionsScene.tsx
```diff
@@ -1,30 +1,47 @@
...
 
+/**
+ * The options rendered in the logs panel header
+ */
 export class LogOptionsScene extends SceneObjectBase<LogOptionsState> {
   static Component = LogOptionsRenderer;
 
-  constructor(state?: Partial<LogOptionsState>) {
+  constructor(state: LogOptionsState) {
+    const logsSortOrderFromLocalStorage = getLogOption('sortOrder');
```
svennergr, line 25: We could change `getLogOption` to `getLogOption<T>(localStorageKey: string, defaultValue: T)` to return `defaultValue` if it's not found. WDYT?

src/Components/ServiceScene/LogOptionsScene.tsx
svennergr: I think it's fine to do.

tests/exploreServicesBreakDown.spec.ts
```diff
@@ -3,6 +3,7 @@ import { ExplorePage, PlaywrightRequest } from './fixtures/explore';
 import { testIds } from '../src/services/testIds';
 import { mockEmptyQueryApiResponse } from './mocks/mockEmptyQueryApiResponse';
 import { LokiQuery } from '../src/services/lokiQuery';
+import exp = require('node:constants');
```
svennergr, line 6:
```suggestion
```

src/Components/ServiceScene/LogOptionsScene.tsx
```diff
@@ -1,30 +1,47 @@
...
+  handleWrapLinesChange = (type: 'wrap' | 'nowrap') => {
+    this.setState({ wrapLogMessage: type === 'wrap' });
+    setLogOption('wrapLogMessage', type === 'wrap');
+    this.getParentScene().setLogsVizOption({ wrapLogMessage: type === 'wrap' });
+  };
+
+  onChangeLogsSortOrder = (sortOrder: LogsSortOrder) => {
+    this.setState({ sortOrder: sortOrder });
+    setLogOption('sortOrder', sortOrder);
+    this.getParentScene().setLogsVizOption({ sortOrder: sortOrder });
```
svennergr, line 44: I think we need to either call this method initially, or set it in the logs panel directly. If I change ordering to "Oldest first", everything works fine. Reloading the page sets the toggle to "Oldest first", but logs are sorted by "newest first".

svennergr (commented):
Left a few comments, especially the sorting on page refresh needs a fix, but I really like the UX of it and using the RadioGroups for all options is really slick.

src/Components/ServiceScene/LogOptionsScene.tsx
```diff
@@ -1,30 +1,47 @@
...
+      wrapLogMessage: Boolean(getLogOption('wrapLogMessage')),
     });
   }
 
-  handleWrapLinesChange = (e: ChangeEvent<HTMLInputElement>) => {
-    const checked = e.target.checked;
-    this.setState({ wrapLines: checked });
-    setLogOption('wrapLines', checked);
-    this.getParentScene().setLogsVizOption({ wrapLogMessage: checked });
+  handleWrapLinesChange = (type: 'wrap' | 'nowrap') => {
```
matyax, line 35:
```suggestion
  handleWrapLinesChange = (wrap: boolean) => {
```
Also seems more compatible with `wrapLogMessage` which is also a boolean.

src/Components/ServiceScene/LogOptionsScene.tsx
gtk-grafana: Good call, added

src/Components/ServiceScene/LogOptionsScene.tsx
gtk-grafana: Fixed, I'll add this case to the e2e test as well

src/Components/ServiceScene/LogOptionsScene.tsx
gtk-grafana: Fixed!

svennergr (approved):
🚀 🚀 🚀 

I think this scrolling behavior is an issue with the logspanel itself, but worth calling out:

https://github.com/user-attachments/assets/f77849a5-9e97-4a82-9836-5865acab3d38

gtk-grafana:
@svennergr yeah the panel tries to keep the same lines in the viewport for better or for worse. It's probably more annoying in a QA context then in a real life situation

svennergr:
> @svennergr yeah the panel tries to keep the same lines in the viewport for better or for worse. It's probably more annoying in a QA context then in a real life situation

That's not the case in core Explore though, so somewhere is a disconnect, I think.

gtk-grafana:
oh, I didn't realize that! @matyax is this a bug in Explore, or Explore logs?

matyax:
I think this is the first time I see this behavior in the logs panel. Probably within LogsPanel or something about Scenes.

PR #924, opened 2024-11-26
Field labels: histogram option for numeric fields

gtk-grafana:
Adds a new option for numeric fields to allow users to toggle between histogram and time series visualization types.
Also adds grouping categories to the PanelMenu.
Saves last selection in local storage as default viz for numeric panels.

image
image

Without investigations:
Numeric panel:
image
Regular panel:
image

Fixes: https://github.com/grafana/explore-logs/issues/923

### DIFF
--- a/src/Components/Panels/PanelMenu.tsx
+++ b/src/Components/Panels/PanelMenu.tsx
@@ -90,6 +90,8 @@ function addHistogramItem(items: PanelMenuItem[], sceneRef: PanelMenu) {
       if (fieldsAggregatedBreakdownScene) {
         fieldsAggregatedBreakdownScene.rebuildAvgFields();
       }
+
+      onSwitchVizTypeTracking(newPanelType);
     },
   });
 }
@@ -188,6 +190,12 @@ const getExploreLink = (sceneRef: SceneObject) => {
 
 const onExploreLinkClickTracking = () => {
   reportAppInteraction(USER_EVENTS_PAGES.all, USER_EVENTS_ACTIONS.all.open_in_explore_menu_clicked);
+};
+
+const onSwitchVizTypeTracking = (newVizType: AvgFieldPanelType) => {
+  reportAppInteraction(USER_EVENTS_PAGES.service_details, USER_EVENTS_ACTIONS.service_details.change_viz_type, {
+    newVizType,
+  });
 };
 
 const getInvestigationLink = (addToExplorations: AddToExplorationButton) => {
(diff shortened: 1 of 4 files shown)


src/Components/Panels/PanelMenu.tsx
```diff
@@ -19,16 +21,77 @@ import { logger } from '../../services/logger';
...
+      const gridItem = sceneGraph.getAncestor(sceneRef, SceneCSSGridItem);
+      const viz = sceneGraph.getAncestor(sceneRef, VizPanel).clone();
+      const $data = sceneGraph.getData(sceneRef).clone();
+      const menu = sceneRef.clone();
+      const headerActions = Array.isArray(viz.state.headerActions)
+        ? viz.state.headerActions.map((o) => o.clone())
+        : viz.state.headerActions;
+      let body;
+
+      if (sceneRef.state.panelType !== AvgFieldPanelType.histogram) {
```
matyax, line 68: Can we add event tracking for this?

src/services/store.ts
```diff
@@ -231,3 +234,16 @@ export function getLogsVisualizationType(): LogsVisualizationType {
...
   localStorage.setItem(VISUALIZATION_TYPE_LOCALSTORAGE_KEY, type);
 }
+
+// Panel options
+const PANEL_OPTIONS_LOCALSTORAGE_KEY = `${pluginJson.id}.panel.option`;
+export interface PanelOptions {
+  panelType: AvgFieldPanelType;
+}
+export function getPanelOption<K extends keyof PanelOptions, V extends PanelOptions[K]>(option: K): V | null {
+  return localStorage.getItem(`${PANEL_OPTIONS_LOCALSTORAGE_KEY}.${option}`) as V | null;
```
matyax, line 244: This could probably use integration with narrowing.ts

src/services/store.ts
```diff
@@ -214,6 +201,22 @@ export function setDisplayedFields(sceneRef: SceneObject, fields: string[]) {
   localStorage.setItem(`${pluginJson.id}.${PREFIX}.logs.fields`, JSON.stringify(fields));
 }
 
+// Log panel options
```
matyax, line 204: Were these 2 functions just moved?

matyax (approved):
Looks good and works as expected! Would appreciate if you add usage tracking and sync with narrowing.ts changes before

src/services/store.ts
gtk-grafana: Yeah, might be good to break out the store a bit in the future, I was just organizing a bit

src/services/store.ts
gtk-grafana: Good catch, I removed the assertion and updated the getPanelOption to require providing all possible values for the panel options, might not work for options with many/infinite possible values, but provides better safety for these enum options.

PR #925, opened 2024-11-27
LogsVolumePanel: Add infinite scroll for logs and display visible range

### LINKED ISSUE #919: 1000 log line default results set limitations.
Would it be possible to extend the search past the default 100 log line limit without having to add more criteria? 

It would make it easier if we users could limit the result set to a smaller number of log lines, for instance 200 and an easy way to look at the next 200 or prev 200.  

I'm not sure if this is possible in Explore Logs.

matyax:
This PR integrates infinite scrolling support using the updated Logs Panel, and displays the visible range of logs, using two annotations to show the time of the oldes and newest log in the list.

The visible range of logs will be updated when infinite scroll is used.

Demo:

https://github.com/user-attachments/assets/a446fe5a-6d5d-4f25-adcb-f4292db56f2d

Requires https://github.com/grafana/grafana/pull/97095 for infinite scrolling + updating the visible range when new logs arrive.

Additionally, it fixes sharing link to log lines, which were being stripped from the URL, causing the Logs Panel to not receive the required parameters.

https://github.com/user-attachments/assets/f5ccd353-8e70-4c81-a3e1-39719ff00092

Closes #919

### DIFF
--- a/src/Components/ServiceScene/LogsPanelScene.tsx
+++ b/src/Components/ServiceScene/LogsPanelScene.tsx
@@ -17,7 +17,7 @@ import { getVariableForLabel } from '../../services/fields';
 import { VAR_FIELDS, VAR_LABELS, VAR_LEVELS, VAR_METADATA } from '../../services/variables';
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from '../../services/analytics';
 import { getAdHocFiltersVariable, getValueFromFieldsFilter } from '../../services/variableGetters';
-import { copyText, generateLogShortlink } from 'services/text';
+import { copyText, generateLogShortlink, resolveRowTimeRangeForSharing } from 'services/text';
 import { CopyLinkButton } from './CopyLinkButton';
 import { getLogsPanelSortOrder, LogOptionsScene } from './LogOptionsScene';
 
@@ -132,15 +132,15 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
   private handleShareLogLineClick = (event: MouseEvent<HTMLElement>, row?: LogRowModel) => {
     if (row?.rowId && this.state.body) {
       const parent = this.getParentScene();
-      const timeRange = sceneGraph.getTimeRange(this.state.body);
       const buttonRef = event.currentTarget instanceof HTMLButtonElement ? event.currentTarget : undefined;
+      const timeRange = resolveRowTimeRangeForSharing(sceneGraph.getTimeRange(this.state.body).state.value, row);
       copyText(
         generateLogShortlink(
           'panelState',
           {
             logs: { id: row.uid, displayedFields: parent.state.displayedFields },
           },
-          timeRange.state.value
+          timeRange
         ),
         buttonRef
       );
--- a/src/Components/ServiceScene/LogsPanelScene.tsx
+++ b/src/Components/ServiceScene/LogsPanelScene.tsx
@@ -118,7 +118,8 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
         .setOption('sortOrder', getLogsPanelSortOrder())
         .setOption('wrapLogMessage', Boolean(getLogOption<boolean>('wrapLogMessage', false)))
         .setOption('showLogContextToggle', true)
-
+        // @ts-expect-error
+        .setOption('enableInfiniteScrolling', true)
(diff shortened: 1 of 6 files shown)


matyax:
Struggling a bit to decide whether to use region or two annotations.

imagen

imagen

I don't like a lot the idea of covering the chart with a region.

zizzpudding:
I think we need the shaded area. It's not super obvious what that annotations are and that they are end points of a range. 

The infinite scrolling piece is awesome.

gtk-grafana:
I agree with Joan, the region should be shaded.

src/services/logsFrame.ts
```diff
@@ -180,3 +190,41 @@ export function getBodyName(logsFrame?: LogsFrame | null): string {
...
+}
+
+export const VISIBLE_RANGE_NAME = 'Visible range';
+export function getVisibleRangeFrame(start: number, end: number) {
+  const frame = arrayToDataFrame([
+    {
+      time: start,
+      timeEnd: start,
+      text: 'Time of the oldest log',
+      color: 'rgb(56, 113, 220, 0.7)',
```
gtk-grafana, line 215: nit: can we pull the theme and grab some more theme appropriate colors from there?
image

Blends in a bit on light mode

gtk-grafana (commented):
One small bug: when logs direction is oldest first, the first infinite scroll removes the "end" annotation from the logs volume panel

src/services/logsFrame.ts
```diff
@@ -180,3 +190,41 @@ export function getBodyName(logsFrame?: LogsFrame | null): string {
...
+  return { start, end };
+}
+
+export const VISIBLE_RANGE_NAME = 'Visible range';
+export function getVisibleRangeFrame(start: number, end: number) {
+  const frame = arrayToDataFrame([
+    {
+      time: start,
+      timeEnd: start,
+      text: 'Time of the oldest log',
```
gtk-grafana, line 214: Another nit: can we make the copy a tad more specific to this feature? 
`Oldest log rendered in logs panel`?


image

Maybe create an issue to add a button to the annotation region tooltip that scrolls to that specific log?

matyax: Definitely. Using a painted region, how does this sound? 

imagen

> Maybe create an issue to add a button to the annotation region tooltip that scrolls to that specific log?

Created https://github.com/grafana/grafana/issues/97257

src/services/logsFrame.ts
matyax: Looked at the theme but couldn't find a suitable color because I need an rgba with transparency. How do we feel about ☝️ ?

gtk-grafana: Yeah looks much better but I can't see the annotation region on this branch yet?

matyax: Pushed.

src/services/panel.ts
```diff
@@ -83,7 +83,7 @@ export function setLevelSeriesOverrides(levels: string[], overrideConfig: FieldC
 }
 
 export function syncLogsPanelVisibleSeries(panel: VizPanel, series: DataFrame[], sceneRef: SceneObject) {
-  const focusedLevels = getVisibleLevels(getLabelsFromSeries(series), sceneRef);
+  const focusedLevels = [...getVisibleLevels(getLabelsFromSeries(series), sceneRef)];
```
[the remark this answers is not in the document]
matyax: Ah good catch. No, it's a leftover from old iterations of this.

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -118,7 +119,10 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
         .setOption('sortOrder', getLogsPanelSortOrder())
         .setOption('wrapLogMessage', Boolean(getLogOption<boolean>('wrapLogMessage', false)))
         .setOption('showLogContextToggle', true)
-
+        // @ts-expect-error
+        .setOption('enableInfiniteScrolling', true)
```
gtk-grafana, line 123: can we comment what version of core is needed to remove these?

gtk-grafana (approved):
Looks good, works well, awesome feature!

src/Components/ServiceScene/LogsPanelScene.tsx
matyax: Definitely. Done.

PR #926, opened 2024-11-27
Chore: Better type safety with ts-reset

gtk-grafana:
Adding https://www.totaltypescript.com/ts-reset to Explore Logs.

The primary use case is for improved type-safety when dealing with values coming from local storage and when parsing JSON encoded strings. There is also some QoL improvements as well, take a look at the link above for more information.

TL;DR localStorage and JSON.parse now return `unknown` instead of `any`.
This PR implements additional runtime narrowing of types to make sure we're not displaying the wrong data, and malicious actors will have a harder time injecting unexpected data through url parameters or local storage.

### DIFF
--- a/src/services/narrowing.ts
+++ b/src/services/narrowing.ts
@@ -1,0 +1,78 @@
+const isString = (s: unknown) => (typeof s === 'string' && s) || '';
+
+const isPropString = <P extends string, T extends Record<P, T[P]>>(
+  prop: P,
+  object: T
+): object is T & Record<P, string> => {
+  return typeof object[prop] === 'string';
+};
+
+export function unknownToStrings(a: unknown): string[] {
+  let strings: string[] = [];
+  if (Array.isArray(a)) {
+    for (let i = 0; i < a.length; i++) {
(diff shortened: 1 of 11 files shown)


src/services/narrowing.ts
```diff
@@ -0,0 +1,83 @@
+import { SelectedTableRow } from '../Components/Table/LogLineCellComponent';
```
matyax, line 1: This is great! I think it would be important to unit test these new functions, because they are doing a lot of stuff, and would also document how we expect to use them.

gtk-grafana: Yeah agreed, I can add tests for the 6 exported functions in here when I get back from the holiday break.

matyax: I wouldn't mind doing it as well, btw.

gtk-grafana: It's all yours if you want it! :)

gtk-grafana:
@matyax tests look good, do you mind approving and we can get this merged and synced with https://github.com/grafana/explore-logs/pull/924?

PR #930, opened 2024-11-29
Queries: remove placeholder query and sanitize stream selector

matyax:
A placeholder string was introduced when creating queries, which was being sent in queries.

example

This has been removed, adding a function to clean up trailing commas and spaces in stream selectors, which can break queries.


gtk-grafana (approved):
Looks good, everything seems to work as expected, thanks for cleaning this up!

PR #931, opened 2024-11-29
Add support to generate OTEL logs in generate script

shantanualsi:
Desciption:
image

This PR adds support to generate OTEL logs from the generate script. It creates another service `loki-ingester-otel`. 
It defines another logger alongside AppLogger that pushes logs to the OTEL collector. 

Few notes:
- Dockerfile is modified that starts otel collector which is required to ingest logs
- Loki address in otel-config is configurable as per the environment used. Can be extracted as env-var later
- Adds slog as it provides a bridge for otel provider
- Some refactor to move files to appropriate directories
- There are a lot of improvements that can be made in the script related to structured metadata and labels. Can be made in subsequent PRs.

Fixes: https://github.com/grafana/explore-logs/issues/224


shantanualsi:
messed up the git commits 🤦 . I'll fix the branch and reopen the PR in a while

PR #937, opened 2024-12-04
Timeseries panels: Map field display names to color

gtk-grafana:
https://github.com/grafana/explore-logs/pull/936 introduced a new "summary" panel to the value breakdowns, but the colors of the series don't match up, which reduces the summary panel's ability to help use volume and other relationships between the series to decide what to filter.

While https://github.com/grafana/explore-logs/issues/939, should help, as that PR would allow users to interact more directly with the viz, having the same colors used in both the breakdowns and summary series will greatly reduce the effort required to match series in the summary panel to the breakdowns.

Any series throughout the app with the same field name will have the same color:
image
image

Except fields aggregated by level, which will show canonical error: red, warning: orange, etc
image

Summary panels will now match breakdown panels:
image


src/services/panel.ts
```diff
@@ -149,9 +230,13 @@ export function getQueryRunner(queries: LokiQuery[], queryRunnerOptions?: Partia
     });
   }
 
-  return getSceneQueryRunner({
-    queries: queries,
-    ...queryRunnerOptions,
+  return new SceneDataTransformer({
```
gtk-grafana, line 233: this breaks some stuff

src/services/panel.ts
```diff
@@ -94,6 +96,85 @@ export function syncLogsPanelVisibleSeries(panel: VizPanel, series: DataFrame[],
...
+    hash |= 0; // to 32bit integer
+  }
+  return hash;
+}
+
+export function setFixedColorByDisplayNameTransformation() {
+  return (source: Observable<DataFrame[]>) => {
+    return source.pipe(
+      map((data: DataFrame[]) => {
+        if (data?.[0]?.refId === LOGS_PANEL_QUERY_REFID) {
```
gtk-grafana, line 147: maybe would be better to exclude the transformer on the logs panel query?

matyax (approved):

gtk-grafana:
Going to wait on merge of parent branch and merge this into main to keep that parent PR from getting any bigger.

PR #944, opened 2024-12-09
Logs Panel: Combine wrapLogMessage with prettifyLogMessage

matyax:
This PR adds `prettifyLogMessage` to work in combination with `wrapLogMessage`. When `wrapLogMessage` is enabled, so will be `prettifyLogMessage`, expanding json log lines if there are.

https://github.com/user-attachments/assets/4915d8b2-5745-4319-8937-5f181acafab9


gtk-grafana (approved):
Nice

PR #951, opened 2024-12-10
Logs Tab: Show log line count

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/425

image

image

image

image

\@TODO:
* ✅  we're running this query on every activation of the ServiceScene, can we only run it when the query/timerange changes or if the current count is undefined?
* ✅  fix e2e assertions now that there's another query in the mix
* ✅  LogsCount needs to sum all infinite scroll pages
* ✅ Custom maxLines support

### DIFF
--- a/src/Components/ServiceScene/ActionBarScene.tsx
+++ b/src/Components/ServiceScene/ActionBarScene.tsx
@@ -109,8 +109,6 @@ const getCounter = (tab: BreakdownViewDefinition, state: ServiceSceneCustomState
       return state.patternsCount;
     case 'labels':
       return state.labelsCount;
-    // case 'logs':
-    //   return state.totalLogsCount
     default:
       return undefined;
   }
--- a/src/Components/ServiceScene/ActionBarScene.tsx
+++ b/src/Components/ServiceScene/ActionBarScene.tsx
@@ -52,7 +52,7 @@ export class ActionBarScene extends SceneObjectBase<ActionBarSceneState> {
                 key={index}
                 label={tab.displayName}
                 active={currentBreakdownViewSlug === tab.value}
-                counter={loadingStates[tab.displayName] ? undefined : getCounter(tab, { ...state })}
+                counter={loadingStates[tab.displayName] ? undefined : getCounter(tab, state)}
                 suffix={
                   tab.displayName === TabNames.logs
                     ? ({ className }) => LogsCount(className, logsCount, totalLogsCount)
@@ -90,8 +90,6 @@ const getCounter = (tab: BreakdownViewDefinition, state: ServiceSceneCustomState
       return state.patternsCount;
     case 'labels':
       return state.labelsCount;
-    case 'logs':
-      return state.totalLogsCount;
     default:
       return undefined;
   }
--- a/src/Components/ServiceScene/ActionBarScene.tsx
+++ b/src/Components/ServiceScene/ActionBarScene.tsx
@@ -10,15 +10,34 @@ import { css, cx } from '@emotion/css';
 import { BreakdownViewDefinition, breakdownViewsDefinitions, TabNames } from './BreakdownViews';
 import { config, usePluginLinks } from '@grafana/runtime';
 import { getLabelsVariable } from '../../services/variableGetters';
-import { LINE_LIMIT } from './Breakdowns/Patterns/PatternNameLabel';
+import { IndexScene } from '../IndexScene/IndexScene';
+import { LINE_LIMIT } from '../../services/query';
 
-export interface ActionBarSceneState extends SceneObjectState {}
+export interface ActionBarSceneState extends SceneObjectState {
+  maxLines?: number;
+}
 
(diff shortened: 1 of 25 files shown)


src/Components/ServiceScene/ActionBarScene.tsx
```diff
@@ -168,3 +175,55 @@ function ToolbarExtensionsRenderer(props: { serviceScene: SceneObject }) {
...
+
+function LogsCount(className: string | undefined, logsCount: number | undefined, totalCount: number | undefined) {
+  const styles = useStyles2(getLogsCountStyles);
+
+  if (logsCount !== undefined) {
+    const valueFormatter = getValueFormat('short');
+
+    const formattedLogsCount = valueFormatter(logsCount, 0);
+
+    if (logsCount < LINE_LIMIT || totalCount === undefined) {
```
gtk-grafana, line 187: Hmm, will this get messed up by infinite scrolling?

gtk-grafana: Not this, but the logsCount total does not sum multiple pages

gtk-grafana:
@matyax There doesn't seem to be any way for the app to know when a new page has been loaded in the LogsPanel with infinite scroll, the observable is not updated, so the panel never goes into the loading state, and there's nothing for us to subscribe to from the app.

Do we need to emit an event from the panel plugin?

matyax:
Would maybe this help? https://github.com/grafana/explore-logs/blob/main/src/Components/ServiceScene/LogsPanelScene.tsx#L128

src/services/query.ts
```diff
@@ -225,3 +228,5 @@ export function unwrapWildcardSearch(input: string) {
 export function sanitizeStreamSelector(expression: string) {
   return expression.replace(/\s*,\s*}/, '}');
 }
+
+export const LINE_LIMIT = 1000;
```
svennergr, line 232:
```suggestion
// default line limit; each data source can define it's own line limit too
export const LINE_LIMIT = 1000;
```

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -397,8 +430,29 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
   }
 
   private subscribeToLogsQuery() {
-    return this.state.$data?.subscribeToState((newState) => {
+    return this.state.$data?.subscribeToState((newState, prevState) => {
       this.updateLoadingState(newState, TabNames.logs);
+      if (newState.data?.state === LoadingState.Done || newState.data?.state === LoadingState.Streaming) {
+        const resultCount = newState.data.series[0].length;
```
svennergr, line 436:
```suggestion
        const resultCount = newState.data.series[0]?.length ?? 0;
```
Just as a safeguard if there's no series?

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -397,8 +430,29 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
+        }
+      }
+    });
+  }
+
+  private subscribeToLogsCountQuery() {
+    return this.state.$logsCount?.subscribeToState((newState) => {
+      if (newState.data?.state === LoadingState.Done) {
+        const value: number | undefined = newState.data.series[0]?.fields?.[1]?.values?.[0];
+        if (value !== undefined) {
```
svennergr, line 450: What happens if it is undefined and the `totalLogsCount` state is currently set? Should we reset to 0?

svennergr (approved):
Left a few nits. For my own taste, we could spare the displayed log count and just show the number of total logs, but that's on you and @zizzpudding, I'd say.

src/Components/ServiceScene/ServiceScene.tsx
gtk-grafana: If the instant query returns "Done" but does not return a result, I guess that would have to be a no-data state, so yeah we should set 0, but the total count won't get displayed if the logs query returns 0, so hopefully it doesn't matter either way.

Updated!

gtk-grafana:
@svennergr I chatted with @zizzpudding, we decided to move the log line count to the logs/table panel header, and additionally place the total log line count in the log volume panel header.

gtk-grafana:
@svennergr I updated the UI, however in my testing I discovered a routing bug that was causing labels containing `/` to redirect back to the same route on every label update, but more importantly I realized how "off" the instant query results are compared to the log volume query, particularly for low volume/ small duration queries.

image

So I added the check back in there to output the results of the logs panel query if the instant query is undefined, and the logs panel returned less then the line count. 

image

gtk-grafana:
Oops we forgot the table header

gtk-grafana:
@svennergr I added the count to the table as well.
@matyax I found a bug with the data frame: https://github.com/grafana/explore-logs/issues/966

Also we might want to remove the logs count in the footer now, but let's do that in another PR.

PR #955, opened 2024-12-11
fix(LogsVolumePanel): fix display of visible range when using cached data

matyax:
Fixes https://github.com/grafana/explore-logs/issues/953

https://github.com/user-attachments/assets/450debc4-81ea-4a67-b9a0-f1964812d22f


gtk-grafana (approved):
LGTM, thanks for the quick fix!

PR #956, opened 2024-12-12
Line filter: add case sensitive line filter state to local storage

gtk-grafana:
Part of: https://github.com/grafana/explore-logs/issues/952

Adds the line filter case sensitivity toggle value to local storage

### DIFF
--- a/src/Components/ServiceScene/LineFilterScene.tsx
+++ b/src/Components/ServiceScene/LineFilterScene.tsx
@@ -13,7 +13,6 @@ import { getLineFilterCase, setLineFilterCase } from '../../services/store';
 interface LineFilterState extends SceneObjectState {
   lineFilter: string;
   caseSensitive: boolean;
-  regex: boolean;
 }
 
 export class LineFilterScene extends SceneObjectBase<LineFilterState> {
@@ -23,7 +22,6 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
     super({
       lineFilter: state?.lineFilter || '',
       caseSensitive: getLineFilterCase(false),
-      regex: false,
       ...state,
     });
     this.addActivationHandler(this.onActivate);
--- a/src/Components/ServiceScene/LineFilterScene.tsx
+++ b/src/Components/ServiceScene/LineFilterScene.tsx
@@ -77,7 +77,7 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
     // Set value in local storage
     setLineFilterCase(caseSensitive);
 
-    this.updateFilter(this.state.lineFilter);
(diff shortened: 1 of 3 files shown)


src/Components/ServiceScene/LineFilterScene.tsx
```diff
@@ -20,7 +21,7 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
   constructor(state?: Partial<LineFilterState>) {
     super({
       lineFilter: state?.lineFilter || '',
-      caseSensitive: false,
+      caseSensitive: getLineFilterCase(false),
```
gtk-grafana, line 24: needs to be set in activation, the value is reset if line filter is empty

src/services/store.ts
```diff
@@ -236,6 +236,22 @@ export function setLogsVisualizationType(type: string) {
   localStorage.setItem(VISUALIZATION_TYPE_LOCALSTORAGE_KEY, type);
 }
 
+// Line filter options
+const LINE_FILTER_OPTIONS_LOCALSTORAGE_KEY = `${pluginJson.id}.linefilter.option`;
+export function setLineFilterCase(caseSensitive: boolean) {
+  let storedValue = caseSensitive.toString();
+  if (typeof caseSensitive === 'boolean' && !caseSensitive) {
```
matyax, line 243: Because of typescript, `typeof caseSensitive` will always be `boolean`, unless the types are incorrectly defined.

src/Components/ServiceScene/LineFilterScene.tsx
```diff
@@ -66,11 +67,17 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
...
     this.setState({
-      caseSensitive: newState === 'sensitive',
+      caseSensitive,
     });
 
-    this.updateFilter(this.state.lineFilter);
+    // Set value in local storage
+    setLineFilterCase(caseSensitive);
+
+    this.updateFilter(this.state.lineFilter, false);
```
matyax, line 80: Just FYI this call to updateFilter() used to be debounced.

matyax (approved):

matyax:
Should we make case sensitive the default?

src/Components/ServiceScene/LineFilterScene.tsx
gtk-grafana: Yeah I don't think there's a reason to debounce a button press toggle, plus it meant if you refreshed before the debounce the UI state wouldn't get added to local storage.

gtk-grafana:
@matyax 
> Should we make case sensitive the default?

I created a spot for this discussion: https://github.com/grafana/explore-logs/issues/962

PR #958, opened 2024-12-12
Line Filter: Add regex option to line filter

gtk-grafana:
Part of: https://github.com/grafana/explore-logs/issues/952
Parent branch: https://github.com/grafana/explore-logs/pull/956

Search for IPs:
image

case insensitive regex
image

case sensitive regex
image

### DIFF
--- a/src/Components/ServiceScene/LineFilterScene.tsx
+++ b/src/Components/ServiceScene/LineFilterScene.tsx
@@ -22,10 +22,9 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
   static Component = LineFilterRenderer;
 
   constructor(state?: Partial<LineFilterState>) {
-    const caseSensitive = getLineFilterCase(false);
     super({
       lineFilter: state?.lineFilter || '',
-      caseSensitive,
+      caseSensitive: getLineFilterCase(false),
       regex: getLineFilterRegex(false),
       ...state,
     });
--- a/src/Components/ServiceScene/LineFilterScene.tsx
+++ b/src/Components/ServiceScene/LineFilterScene.tsx
@@ -43,7 +43,6 @@ export class LineFilterScene extends SceneObjectBase<LineFilterState> {
     const caseSensitiveMatches = caseSensitive
       ? lineFilterString.match(/\|=.`(.+?)`/)
       : lineFilterString.match(/`\(\?i\)(.+)`/);
-    const regexMatches = lineFilterString.match(/\|~.+\`(.*?)\`/);
 
     // If the existing query is case sensitive, overwrite the users options for case sensitivity
     if (caseSensitiveMatches && caseSensitiveMatches.length === 2) {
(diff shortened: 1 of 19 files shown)


src/services/store.ts
```diff
@@ -247,11 +247,25 @@ export function setLineFilterCase(caseSensitive: boolean) {
   localStorage.setItem(`${LINE_FILTER_OPTIONS_LOCALSTORAGE_KEY}.caseSensitive`, storedValue);
 }
 
+export function setLineFilterRegex(regex: boolean) {
+  let storedValue = regex.toString();
+  if (typeof regex === 'boolean' && !regex) {
```
matyax, line 252: Same. This function wil always be called with booleans or the signature must reflect the possible types for `regex`.

src/Components/ServiceScene/RegexIcon.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const RegexIcon = (props: Props) => {
+  const theme = useTheme2();
+  const fill = props.regex ? theme.colors.text.maxContrast : theme.colors.text.disabled;
+  const styles = getStyles(theme, fill);
+
+  return (
+    <span className={styles.container} title={`${props.regex ? 'String comparison' : 'Regex matching' }`}>
+      <svg
+        onClick={() => props.onRegexToggle(props.regex ? 'match' : 'regex')}
```
matyax, line 21: Nitpicky, but having `onClick` listeners would make these not `Icon` but `IconButton`. 

Also we should probably enable accessibility rules for this repo, because elements with onClick listeners should be keyboard accessible, so this should be wrapped in a button. Same for LineFilterIcon.

src/Components/ServiceScene/LineFilterScene.tsx
```diff
@@ -1,27 +1,32 @@
...
 export class LineFilterScene extends SceneObjectBase<LineFilterState> {
   static Component = LineFilterRenderer;
 
   constructor(state?: Partial<LineFilterState>) {
+    const caseSensitive = getLineFilterCase(false);
     super({
       lineFilter: state?.lineFilter || '',
-      caseSensitive: getLineFilterCase(false),
+      caseSensitive,
+      regex: getLineFilterRegex(false),
```
matyax, line 29:
```suggestion
      caseSensitive: getLineFilterCase(false),
      regex: getLineFilterRegex(false),
```

matyax, line 29: One issue with `regex` is that we're setting it based on the stored preference and not the filter contents.

matyax (commented):
It's almost there. There is an issue to resolve when you share or reuse a URL, because the `regex` state doesn't depend on the filter contents. Also it seems that even with regex turned out we're producing and interpreting regexes (the URL I open in incognito with the regex correctly produces log results).

https://github.com/user-attachments/assets/e192cdf7-7537-4358-8426-f6269e84c338

matyax:
After seeing https://github.com/grafana/explore-logs/pull/956 and this PR, I think we should change the approach:

- Regex on, show "case sensitive/case insensitive" toggle. Changing it changes the regular expression.
- Regex off, don't show sensitiveness toggle, always produce a Line filter.
- Parse URL: if the URL contains a regex, set `regex` to true, parse the regular expression, parse case sensitiveness and also set the state.

src/Components/ServiceScene/RegexIcon.tsx
gtk-grafana: Good catch, should be fixed now

PR #960, opened 2024-12-12
Keybindings: support time range copy/paste 

gtk-grafana:
Add support for `t c` and `t v` in Explore Logs.

Related: https://github.com/grafana/scenes/pull/999
Also ran into this bug while testing: https://github.com/grafana/grafana/issues/97892

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -225,7 +225,7 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
   private subscribeToPasteTimeEvent = async () => {
     const copiedRange = await getCopiedTimeRange();
 
-    if (!copiedRange.isError) {
+    if (copiedRange.isError) {
       return;
     }
 
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -226,26 +226,27 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     const copiedRange = await getCopiedTimeRange();
 
     if (!copiedRange.isError) {
-      const timeRange = sceneGraph.getTimeRange(this);
-      const to = typeof copiedRange.range.to === 'string' ? copiedRange.range.to : undefined;
-      const from = typeof copiedRange.range.from === 'string' ? copiedRange.range.from : undefined;
+      return;
+    }
 
-      const newRange = rangeUtil.convertRawToRange(copiedRange.range);
+    const timeRange = sceneGraph.getTimeRange(this);
+    const to = typeof copiedRange.range.to === 'string' ? copiedRange.range.to : undefined;
+    const from = typeof copiedRange.range.from === 'string' ? copiedRange.range.from : undefined;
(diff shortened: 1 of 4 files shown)


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -221,6 +222,33 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
...
+      } else {
+        logger.error(new Error('Invalid time range from clipboard'), {
+          msg: 'Invalid time range from clipboard',
+          sceneTimeRange: typeof timeRange,
+          to: to ?? '',
+          from: from ?? '',
+        });
+      }
+    }
+  };
```
matyax, line 250: As usual, one less level of indentation.

```suggestion
  private subscribeToPasteTimeEvent = async () => {
    const copiedRange = await getCopiedTimeRange();

    if (!copiedRange.isError) {
      return;
    }
    const timeRange = sceneGraph.getTimeRange(this);
    const to = typeof copiedRange.range.to === 'string' ? copiedRange.range.to : undefined;
    const from = typeof copiedRange.range.from === 'string' ? copiedRange.range.from : undefined;

    const newRange = rangeUtil.convertRawToRange(copiedRange.range);

    if (timeRange && newRange) {
      timeRange.setState({
        value: newRange,
        to,
        from,
      });
    } else {
      logger.error(new Error('Invalid time range from clipboard'), {
        msg: 'Invalid time range from clipboard',
        sceneTimeRange: typeof timeRange,
        to: to ?? '',
        from: from ?? '',
      });
    }
  };
```

src/services/keyboardShortcuts.ts
```diff
@@ -63,6 +64,26 @@ export function setupKeyboardShortcuts(scene: IndexScene) {
     }),
   });
 
+  // Copy time range
+  keybindings.addBinding({
+    key: 't c',
+    onTrigger: () => {
+      const timeRange = sceneGraph.getTimeRange(scene);
+      setWindowGrafanaSceneContext(timeRange);
```
matyax, line 72: Not entirely sure what this is doing.

src/services/keyboardShortcuts.ts
```diff
@@ -146,3 +167,64 @@ export function toggleVizPanelLegend(vizPanel: VizPanel): void {
...
+      (window as any).__grafanaSceneContext = prevScene;
+    }
+  };
+}
+
+// taken from /Users/galen/projects/grafana/grafana/public/app/core/utils/timePicker.ts
+type CopiedTimeRangeResult = { range: RawTimeRange; isError: false } | { range: string; isError: true };
+
+// modified to narrow types from clipboard
+export async function getCopiedTimeRange(): Promise<CopiedTimeRangeResult> {
```
matyax, line 210: A couple of unit tests wouldn't hurt.

src/services/keyboardShortcuts.ts
gtk-grafana: you are not alone :P 
I added the docblock from the definition in scenes, the important bit `This is also used from TimeSrv to access scene time range.`

matyax (approved):
💯

PR #964, opened 2024-12-17
Logs Volume: Set relative height and allow to collapse

matyax:
The Logs Volume panel has been updated to support being collapsed. In addition, its height has been set to be 20% of the viewport height or 100px.

https://github.com/user-attachments/assets/6b8b3450-6494-4be1-9302-ab88575c0d10


gtk-grafana (approved):
Looks good, works well locally

PR #965, opened 2024-12-18
fix(panelmenu): `Investigations` causing multiple same keys

svennergr:
We used `ADD_TO_INVESTIGATION_MENU_DIVIDER_TEXT` / `Investigations` for the divider and group, which cause React to complain about multiple same keys:

image


gtk-grafana (approved):
whoops, sorry about that! LGTM

PR #967, opened 2024-12-18
Logs panel: update service data when receiving new logs

### LINKED ISSUE #966: Table: Infinite scroll breaking table summary and line counts
On the `logsPanelQuery`, the series length is not being adjusted, but values are. This is causing the table count in the footer to show the wrong number:
Image

and the density calculations are wrong

Image

Image

CC: @matyax

matyax:
Before these changes, the ServiceScene was keeping the original data frame but with mutated fields after merging, so it would have an original length of 1000 but potentially many more field values.

With these changes, the table will have access to the updated data frame.

imagen

Without this, paginated results will be lost or not available when switching back and forth from table to logs.

Fixes #966

### DIFF
--- a/src/Components/ServiceScene/LogsPanelScene.tsx
+++ b/src/Components/ServiceScene/LogsPanelScene.tsx
@@ -167,6 +167,11 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
       logsCount: newLogs[0].length,
     });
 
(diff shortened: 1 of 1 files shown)


matyax:
Note: sometimes when infinite scroll is triggered, the scroll position is being reset to the top. It should be fixed by the additional change in https://github.com/grafana/grafana/pull/98203

gtk-grafana (approved):
LGTM

PR #968, opened 2024-12-19
fix(panel-menu): menu throwing error in logs table

svennergr:
Opening panel menu was throwing an error in Logs Table because no `VizPanel` ancestor could be found. Interestingly `findObjectOfType` returned a `VizPanel` instance, but `sceneGraph.getAncestor` threw an error.

Change the behavior to always use `sceneGraph.getAncestor` and catch the error in case there's not `VizPanel`.

### DIFF
--- a/src/Components/Panels/PanelMenu.tsx
+++ b/src/Components/Panels/PanelMenu.tsx
@@ -59,22 +59,6 @@ export class PanelMenu extends SceneObjectBase<PanelMenuState> implements VizPan
   constructor(state: Partial<PanelMenuState>) {
     super({ ...state, addExplorationsLink: state.addExplorationsLink ?? true });
     this.addActivationHandler(() => {
-      const viz = findObjectOfType(this, (o) => o instanceof VizPanel, VizPanel);
-
-      this.setState({
-        explorationsButton: new AddToExplorationButton({
-          labelName: this.state.labelName,
-          fieldName: this.state.fieldName,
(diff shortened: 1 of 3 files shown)


src/Components/Panels/PanelMenu.tsx
```diff
@@ -59,22 +59,6 @@ export class PanelMenu extends SceneObjectBase<PanelMenuState> implements VizPan
   constructor(state: Partial<PanelMenuState>) {
     super({ ...state, addExplorationsLink: state.addExplorationsLink ?? true });
     this.addActivationHandler(() => {
-      const viz = findObjectOfType(this, (o) => o instanceof VizPanel, VizPanel);
```
matyax, line 62: We should expect that findObjectOfType doesn't find what we're looking for and account for that case. Similar to when we use `Array.find()`.

matyax (approved):

src/Components/Panels/PanelMenu.tsx
gtk-grafana: Isn't the problem here that we do find something with findObjectOfType, it's just the wrong viz panel?

gtk-grafana: looks like findObject will recursively check all of the children of each parent, so it grabs the viz panel for the logs volume instead

gtk-grafana (approved):
LGTM, thanks for adding the e2e test as well

PR #973, opened 2025-01-02
Add support to generate OTEL logs in generate script

shantanualsi:
This PR adds support to generate OTEL logs from the generate script. It creates another service loki-ingester-otel, distributor-otel, among others.
It defines another logger alongside AppLogger that pushes logs to the OTEL collector.

Few notes:

- Dockerfile is modified that starts otel collector which is required to ingest logs
- Loki address in otel-config is configurable as per the environment used. Can be extracted as env-var later
- Adds slog as it provides a bridge for otel provider
- Some refactor to move files to appropriate directories
- There are a lot of improvements that can be made in the script related to structured metadata and labels. Can be made in subsequent PRs.

Fixes: https://github.com/grafana/explore-logs/issues/224

Previous PR with few comments: https://github.com/grafana/explore-logs/pull/931 (closed due to unclean git history)


shantanualsi:
@gtk-grafana to [your comment ](https://github.com/grafana/explore-logs/pull/931#issuecomment-2511873927)on the original PR, I have modified `yarn generate-logs` to run inside docker because I'm using `otelcol-contrib` which would need platform specific downloads if I put the shell script outside the dockerfile. Let me know if there are any concerns in this.

generator/otel-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+processors:
+  batch:
+    timeout: 1s
+    send_batch_size: 1024
+
+exporters:
+  otlphttp/logs:
+    # change this depending on where you're running loki (can be made configurable)
+    endpoint: http://localhost:3100/otlp
```
svennergr, line 17: `yarn server` will not produce any otel logs as it is using `localhost` by default while Loki does not run in the same container.

Should we default to 
```suggestion
    endpoint: http://host.docker.internal:3100/otlp
```
or make it configurable?

shantanualsi: sure that makes sense. thanks!

shantanualsi: i think better to make it configurable somehow, i need to fix the dockerfile after this commit to use `localhost`.. will push the commit in sometime

PR #974, opened 2025-01-02
feat(patterns): use grafana's calculated `interval` as `step`

svennergr:
Optimizes the patterns graph. 

Due to defaulting to no step, graphs can sometimes look very spikey:
image

while they actually shouldn't be when used with a correct `step`:

image


gtk-grafana (approved):
Good catch!

PR #983, opened 2025-01-07
Breakdowns: Add share menu

gtk-grafana:
Adds a share button similar to button in classic Explore:

image

image

Fixes: https://github.com/grafana/explore-logs/issues/771
Related: https://github.com/grafana/scenes/pull/1022


src/Components/IndexScene/ShareButtonScene.tsx
```diff
@@ -0,0 +1,264 @@
...
+    });
+  } else {
+    appEvents.publish({
+      type: AppEvents.alertError.name,
+      payload: ['Error generating shortened link'],
+    });
+  }
+};
+
+export const copyStringToClipboard = (string: string) => {
```
matyax, line 238: There's https://github.com/grafana/explore-logs/blob/main/src/services/text.ts#L5

matyax (approved):

src/Components/IndexScene/ShareButtonScene.tsx
matyax: Any reason not to use the existing function?

gtk-grafana: Oh snap I never sent my comment:
copyText seems to require a button to append the textarea for fallback when clipboard.writeText isn't available, we should probably refactor `copyText` to work like `copyStringToClipboard` instead of the other way around. 

Created https://github.com/grafana/explore-logs/issues/986 to track, I can clean this up in the morning

matyax: No worries. IIRC the other function has a backwards compatibility feature that makes it a bit more resilient.

PR #985, opened 2025-01-07
Gtk grafana/issues/984/logs panel url sync

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/984

### DIFF
--- a/src/Components/ServiceScene/LogsPanelScene.tsx
+++ b/src/Components/ServiceScene/LogsPanelScene.tsx
@@ -5,6 +5,8 @@ import {
   sceneGraph,
   SceneObjectBase,
   SceneObjectState,
+  SceneObjectUrlSyncConfig,
+  SceneObjectUrlValues,
   SceneQueryRunner,
   VizPanel,
 } from '@grafana/scenes';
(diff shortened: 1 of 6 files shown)


src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -20,28 +22,93 @@ import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from '..
...
+    } catch (e) {
+      // URL Params can be manually changed and it will make JSON.parse() fail.
+      logger.error(e, { msg: 'LogOptionsScene: updateFromUrl unexpected error' });
+    }
+
+    if (Object.keys(stateUpdate).length) {
+      this.setState({ ...stateUpdate });
+    }
+  }
+
```
svennergr, line 100: Do we want to store in store if we update from URL? I navigated from a link, had `wrapping: true`, went to service selection, selected a new service, and `wrapping: false`.
Would be similar question for sort order.

gtk-grafana: I would argue we only want to update local storage state when the user makes an action, e.g. if I'm viewing a link that looks better with line wrap off, I wouldn't expect viewing that link to change my settings next time I start a new investigation

svennergr (approved):
LGTM

src/Components/ServiceScene/LogsPanelScene.tsx
gtk-grafana: But this isn't a hill I'm trying to die on, I can forsee situations where either situation is more desirable 🤷

svennergr: Yea, same. Let's go with the current state.

PR #987, opened 2025-01-07
chore: clean up copy texts

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/986

### DIFF
--- a/src/services/text.ts
+++ b/src/services/text.ts
@@ -2,29 +2,16 @@ import { locationService } from '@grafana/runtime';
 import { logger } from './logger';
 import { dateTime, LogRowModel, TimeRange } from '@grafana/data';
 
-export const copyText = async (
-  text: string,
-  buttonRef?: React.MutableRefObject<HTMLButtonElement | null> | HTMLButtonElement
-) => {
+export const copyText = (string: string) => {
   if (navigator.clipboard && window.isSecureContext) {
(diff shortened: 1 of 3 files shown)


gtk-grafana:
So interestingly enough we copied both methods from core Grafana, they both seem to have fallbacks, and neither seem to have test coverage (but it's a PITA testing clipboards in both playwright and jest).

Probably be good to manually test this a bit more manually, I don't think we need the buttonref implementation as we're not copying from within a modal or something that hijacks focus.

matyax:
> I don't think we need the buttonref implementation as we're not copying from within a modal or something that hijacks focus.

Yeah, agree.

src/services/text.ts
```diff
@@ -2,29 +2,16 @@ import { locationService } from '@grafana/runtime';
 import { logger } from './logger';
 import { dateTime, LogRowModel, TimeRange } from '@grafana/data';
 
-export const copyText = async (
-  text: string,
-  buttonRef?: React.MutableRefObject<HTMLButtonElement | null> | HTMLButtonElement
-) => {
+export const copyText = (string: string) => {
   if (navigator.clipboard && window.isSecureContext) {
```
matyax, line 6: Much better 👌

matyax (approved):

PR #994, opened 2025-01-09
fix: use urlUtil instead of UrlSearchParams

gtk-grafana:
Fixes bug introduced by URLSearchParams which will replace whitespace with `+` instead of `%20`
Fixes bug in line filter migration where escaped chars would not get unescaped.

### DIFF
--- a/src/services/migrations.ts
+++ b/src/services/migrations.ts
@@ -17,16 +17,7 @@ function removeEscapeChar(value: string, caseSensitive: boolean) {
       // Since there is no reverse of the escapeRegExp method provided by lodash we're essentially building our own "unescapeRegExp"
       const nextChar = stringArray[index + 1];
       const isNextCharRegex = charsEscapedByEscapeRegExp.includes(nextChar);
-      const filterChar = !(char === '\\' && isNextCharRegex);
-
-      console.log('char', {
-        char,
-        nextChar,
-        isNextCharRegex,
-        filterChar,
-      });
-
-      return filterChar;
+      return !(char === '\\' && isNextCharRegex);
     })
     .join('');
 }
--- a/src/services/migrations.ts
+++ b/src/services/migrations.ts
@@ -5,10 +5,29 @@ import { LineFilterOp } from './filterTypes';
 import { ServiceScene } from '../Components/ServiceScene/ServiceScene';
 import { urlUtil } from '@grafana/data';
 
-function removeEscapeChar(value: string) {
+function removeEscapeChar(value: string, caseSensitive: boolean) {
+  const charsEscapedByEscapeRegExp = ['^', '$', '.', '*', '+', '?', '(', ')', '[', ']', '{', '}', '|'];
+  if (!caseSensitive) {
+    charsEscapedByEscapeRegExp.push('\\');
+  }
   return value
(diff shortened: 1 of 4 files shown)


src/services/migrations.ts
```diff
@@ -3,16 +3,27 @@ import { getLineFiltersVariable } from './variableGetters';
 import { LineFilterCaseSensitive } from '../Components/ServiceScene/LineFilter/LineFilterScene';
 import { LineFilterOp } from './filterTypes';
 import { ServiceScene } from '../Components/ServiceScene/ServiceScene';
+import { urlUtil } from '@grafana/data';
+
+function removeEscapeChar(value: string) {
+  return value
```
matyax, line 9: Curious about the unusual approach. Why not a regex?

tests/exploreServicesBreakDown.spec.ts
```diff
@@ -1455,23 +1455,31 @@ test.describe('explore services breakdown page', () => {
...
+        'page_url%3D%22https:%2F%2Fgrafana%5C.net%2Fexplore%5C%3Fleft%3D%5C%7B%22datasource%22:%22grafanacloud-prom%22,%22queries%22:%5C%5B%5C%7B%22datasource%22:%5C%7B%22type%22:%22prometheus%22,%22uid%22:%22grafanacloud-prom%22%5C%7D,%22expr%22:%22max%20by%20%5C%28kube_cluster_name,%20kube_namespace%5C%29%20%5C%28quantile_over_time%5C%280%5C.85,%20kubernetes_state_pod_age%5C%7Bplatform%3D%22data%22,kube_namespace%21~%22data-dev%5C%7Cdata-stg%22,pod_phase%3D%22pending%22%5C%7D%5C%5B5m%5C%5D%5C%29%5C%29%20%3E%20600%22,%22refId%22:%22A%22%5C%7D%5C%5D,%22range%22:%5C%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%5C%7D%5C%7D%22%60';
+      const decodedAndUnescaped =
+        'page_url="https://grafana.net/explore?left={"datasource":"grafanacloud-prom","queries":[{"datasource":{"type":"prometheus","uid":"grafanacloud-prom"},"expr":"max by (kube_cluster_name, kube_namespace) (quantile_over_time(0.85, kubernetes_state_pod_age{platform="data",kube_namespace!~"data-dev|data-stg",pod_phase="pending"}[5m])) > 600","refId":"A"}],"range":{"from":"now-1h","to":"now"}}"';
+      await explorePage.gotoServicesOldUrlLineFilters('tempo-distributor', false, urlEncodedAndEscaped);
```
matyax, line 1476: Took me a while to understand the trick with the string encoded/decoded.

matyax (approved):

src/services/migrations.ts
gtk-grafana: Was struggling to get the regex to work last night, I'll take another look

gtk-grafana: Nevermind, I made this implementation more complicated to fix the user-entered escape char case, so I think it's beyond my regex abilities without dumping a bunch more time into this.

PR #997, opened 2025-01-14
Link extensions: Add support for line filters

gtk-grafana:
Pull line filters from link extension queries into explore logs.

Note: Updates scenes to 5.37.0

### DIFF
--- a/src/services/extensions/links.test.ts
+++ b/src/services/extensions/links.test.ts
@@ -191,7 +191,79 @@ describe('contextToLink', () => {
     const expectedLineFiltersUrlString =
       '&var-lineFilters=caseSensitive%2C0%7C__gfp__%3D%7C%22+%28%3Fi%29caller%22' +
       // &var-lineFilters=caseSensitive,1|__gfp__=|+(?i)caller.+\\\\n
-      '&var-lineFilters=caseSensitive%2C1%7C__gfp__%3D%7C+%28%3Fi%29caller.%2B%5C%5C%5C%5Cn';
+      '&var-lineFilters=caseSensitive%2C1%7C__gfp__%3D%7C+%28%3Fi%29caller.%2B%5C%5Cn';
+
+    expect(config).toEqual({
+      path: `/a/grafana-lokiexplore-app/explore/cluster/eu-west-1/logs?var-ds=123abc&from=1675828800000&to=1675854000000${expectedLabelFiltersUrlString}${expectedLineFiltersUrlString}`,
+    });
+  });
+  it('should parse case sensitive non-regex line-filter containing double quotes', () => {
+    const links = linkConfigs;
+    const target: { refId: string } & Partial<LokiQuery> = {
+      expr: '{cluster="eu-west-1"} |= "thread \\\\\\"main\\\\\\""',
+      datasource: {
+        type: 'loki',
+        uid: '123abc',
+      },
+      refId: 'A', // Ensure refId is defined
+    };
+    const config = links?.[0].configure?.({
+      timeRange: {
+        from: dateTime('2023-02-08T04:00:00.000Z'),
+        to: dateTime('2023-02-08T11:00:00.000Z'),
+      },
+      pluginId: 'grafana-lokiexplore-app',
+      timeZone: 'browser',
+      id: 0,
+      title: 'test',
+      dashboard: {
+        tags: [],
+        title: 'test',
+        uid: 'test',
+      },
+      targets: [target],
+    });
+
+    const expectedLabelFiltersUrlString = '&var-filters=cluster%7C%3D%7Ceu-west-1';
+
+    // &var-lineFilters=caseSensitive,0|__gfp__=|thread \"main\"
+    const expectedLineFiltersUrlString = '&var-lineFilters=caseSensitive%2C0%7C__gfp__%3D%7Cthread+%5C%22main%5C%22';
+
+    expect(config).toEqual({
+      path: `/a/grafana-lokiexplore-app/explore/cluster/eu-west-1/logs?var-ds=123abc&from=1675828800000&to=1675854000000${expectedLabelFiltersUrlString}${expectedLineFiltersUrlString}`,
+    });
+  });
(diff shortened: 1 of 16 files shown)


src/services/extensions/links.test.ts
```diff
@@ -34,4 +34,239 @@ describe('contextToLink', () => {
...
+      timeZone: 'browser',
+      id: 0,
+      title: 'test',
+      dashboard: {
+        tags: [],
+        title: 'test',
+        uid: 'test',
+      },
+      targets: [target],
+    });
```
matyax, line 62: I think this part repeats exactly on every test, ritght? Nit: pulling it to a constant would make reading the tests a bit easier.

src/services/extensions/links.ts
```diff
@@ -1,12 +1,13 @@
...
 import { PluginExtensionLinkConfig, PluginExtensionPanelContext, PluginExtensionPoints } from '@grafana/data';
 
-import { SERVICE_NAME, VAR_DATASOURCE, VAR_FIELDS, VAR_LABELS } from 'services/variables';
+import { SERVICE_NAME, VAR_DATASOURCE, VAR_FIELDS, VAR_LABELS, VAR_LINE_FILTERS } from 'services/variables';
 import pluginJson from '../../plugin.json';
 import { LabelType } from '../fieldsTypes';
 import { getMatcherFromQuery } from '../logqlMatchers';
 import { LokiQuery } from '../lokiQuery';
 import { FilterOp } from '../filterTypes';
+// import { sceneUtils } from '@grafana/scenes';
```
matyax, line 10:
```suggestion
```

src/services/extensions/links.ts
```diff
@@ -81,6 +82,17 @@ function contextToLink<T extends PluginExtensionPanelContext>(context?: T) {
       params
     );
   }
+
+  if (lineFilters) {
+    for (const lineFilter of lineFilters) {
+      params = appendUrlParameter(
+        UrlParameters.LineFilters,
+        `${lineFilter.key}|${escapeURLDelimiters(lineFilter.operator)}|${escapeURLDelimiters(lineFilter.value)}`,
```
matyax, line 90: Unrelated, but seeing `|` as the query delimiter made me wonder about the following query, producing results with "case sensitivity" (quoted because it's just natural line filters) and no results with case sensitiveness disabled.

`{service_name=`mimir/ingester`}    |= \"|\" | json | logfmt | drop __error__, __error_details__ `

imagen

`{service_name=`mimir/ingester`}    |~ \"(?i)\\\\|\" | json | logfmt | drop __error__, __error_details__ `

imagen

Seems to be being processed as a regex? If so, I think that's the problem with "case insensitive" being a regex under the hood.



@cyriltovena / @trevorwhitney would you consider adding an actual case insensitive line filter in LogQL?

src/services/extensions/links.ts
```diff
@@ -120,3 +133,26 @@ export function appendUrlParameter(
 export function replaceSlash(parameter: string): string {
   return parameter.replace(/\//g, '-');
 }
+
+// Manually copied over from @grafana/scenes so we don't need to import scenes to build links
```
matyax, line 137: Are these covered by unit tests? Maybe indirectly in links.test.ts, but just double checking.

src/services/logqlMatchers.ts
```diff
@@ -89,8 +103,82 @@ export function getMatcherFromQuery(query: string): Filter[] {
       type: selectorPosition.contains(matcherPosition) ? LabelType.Indexed : undefined,
     });
   }
+}
+
+function parseLineFilters(query: string, lineFilters: LineFilterType[]) {
```
matyax, line 108: Same question about unit tests, direct or indirect, for these functions.

matyax:
Possibly unrelated but I'm seeing this:

imagen

matyax (commented):
Hopefully I didn't misunderstand the PR, but I would not be getting this to work:

https://github.com/user-attachments/assets/918c6dfa-c4d1-42c2-9440-c51e194b24ec

src/services/logqlMatchers.ts
gtk-grafana: same as the scenes methods above, covered by tests in `links.test.ts` and e2e

src/services/extensions/links.ts
gtk-grafana: The queries I get for this are 
```
{service_name=`grafana/grafana`} |= "|" | json | logfmt | drop __error__, __error_details__
```
image

and 

```
{service_name=`grafana/grafana`} |~ "(?i)\\|" | json | logfmt | drop __error__, __error_details__
```

image

the LogQL looks correct, but it is confusing that the case-insensitive query returns different results.

If you query more then just `|`, it does appear to work as expect:
```
{service_name=`grafana/grafana`} |~ "(?i)\\| json \\|" | json | logfmt | drop __error__, __error_details__
```
image

seems to return the same result as
```
{service_name=`grafana/grafana`} |= "| json |" | json | logfmt | drop __error__, __error_details__
```

image

gtk-grafana: @svennergr is going to chat with Loki folks as this looks like a bug in Loki and is reproducible on main/outside explore logs

src/services/extensions/links.ts
```diff
@@ -16,14 +16,16 @@ export const ExtensionPoints = {
...
+export type LinkConfigs = Array<
   {
     targets: string | string[];
     // eslint-disable-next-line deprecation/deprecation
   } & Omit<PluginExtensionLinkConfig<PluginExtensionPanelContext>, 'type' | 'extensionPointId'>
-> = [
+>;
+
+// `plugin.addLink` requires these types; unfortunately, the correct `PluginExtensionAddedLinkConfig` type is not exported with 11.2.x
+// TODO: fix this type when we move to `@grafana/data` 11.3.x
```
svennergr, line 27: AFAIK we can upgrade dependencies without needing to upgrade minimum grafana version.

svennergr (approved):
Works for me! LGTM 🚀

PR #999, opened 2025-01-14
Link extensions: add support for fields

gtk-grafana:
Support fields in link extensions.

Notes:
* metadata is detected by fields that occur before the parsers
* the field parser is determined by the parsers used in the query
* * this will result in sub-optimal queries as we don't override the parser passed in the (JSON encoded) ad-hoc filter value
* * we'd need to call the detected_fields api to figure out if we should use one parser or the other, and this is not currently used in classic explore
* * * we could potentially update the parser whenever we got an update in the detected_fields response within Explore Logs, but I'm concerned that we could potentially use a different set of parsers then the user added in their query, which could change the label set between explore and explore logs, resulting in bugs.
* regex fields are not supported and will be dropped
* only the first field (`LabelFilter`) expression in each PipeLineExpr will be used, so `field1="value1" or field2="value2"` will only return `field1="value1"` in Explore Logs
* we check the dataframe field `labelTypes` to figure out the LabelType (indexed, structuredmetadata, parsed), but we only check the first row, which could cause us to incorrectly identify a structured metadata label as parsed


src/services/fieldsTypes.ts
```diff
@@ -1,7 +0,0 @@
-// Warning: This file is included in the main bundle with Grafana in order to provide link extension support in Grafana core, in an effort to keep Grafana loading quickly, please do not add any unnecessary imports to this file and run the bundle analyzer before committing any changes!
```
gtk-grafana, line 1: why?

gtk-grafana: undo

src/services/logqlMatchers.ts
```diff
@@ -167,18 +209,133 @@ function parseLineFilters(query: string, lineFilters: LineFilterType[]) {
...
+    const position = NodePosition.fromNode(matcher);
+    const expression = position.getExpression(query);
+
+    // Skip error expression, it will get added automatically when explore logs adds a parser
+    if (expression.substring(0, 9) === `__error__`) {
+      continue;
+    }
+
+    // @todo we need to use detected_fields API to get the "right" parser for a specific field
+    // Currently we just check to see if there is a parser before the current node, this means that queries that are placing metadata filters after the parser will query the metadata field as a parsed field, which will lead to degraded performance
```
gtk-grafana, line 257: I guess in the app we could check if a field is actually metadata once we execute the detected_fields call and move the filter from the fields variable to the metadata variable?

matyax (approved):
Looks great!

PR #1010, opened 2025-01-21
Regex labels: support queries from explore

gtk-grafana:
This PR adds support for all labels in Explore Logs coming from an Explore query.

The ability to write arbitrary regex queries within Explore Logs is not supported yet, but this PR is the first step towards providing support in the UI for adding regex label queries.
Field regex operators are not supported yet, but should be a fast follow.

Fixes: https://github.com/grafana/explore-logs/issues/1008

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -89,6 +89,7 @@ import { getCopiedTimeRange, PasteTimeEvent, setupKeyboardShortcuts } from '../.
 import { LokiDatasource } from '../../services/lokiQuery';
 import {
   includeOperators,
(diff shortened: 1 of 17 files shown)


matyax:
Just checking if I understand correctly, but this actually doesn't fix https://github.com/grafana/explore-logs/issues/1008, right?

matyax (approved):
Looks good and works well locally!

gtk-grafana:
> Just checking if I understand correctly, but this actually doesn't fix https://github.com/grafana/explore-logs/issues/1008, right?

It should, that issue is just scoped to pulling any label selector from explore, not writing arbitrary regex queries in Explore Logs

PR #1013, opened 2025-01-23
feat: include adaptive logs exemptions component (extension)

darrenjaneczek:
Embed an explicit exposed component extension from the `grafana-adaptivelogs-app` Grafana cloud plugin, which conditionally contributes a contextual button to the log volume panel. 

Notes:
- If the  grafana-adaptivelogs-app is not present (rather, a new enough build that will contribute this extension), nothing will appear in the log volume actions header.
- If a datasource not managed by the adaptive logs service, nothing will appear...

The purpose of this button is to give the user the ability to create a temporary exemption for the selected label selector, to disable adaptive logs and collect more logs related to the currently displayed log volume. This will be handled by the plugin exposing this component extension.

This plugin merely needs to collect the data source and label selector (from the ad-hoc filter), and pass that context along. Additional context is included as part of `contextHints={['log-volume-panel', 'header-action']}`, which in the future will be used to place this component in different places and adjust the way it looks. 

As per @lukasztyrala:

image

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -508,3 +508,4 @@ linefilter
 queryfrontend
 timesync
 openmetrics
+adaptivelogs
--- a/project-words.txt
+++ b/project-words.txt
@@ -509,3 +509,4 @@ queryfrontend
 timesync
 openmetrics
 adaptivelogs
+subselector
--- a/src/Components/ServiceScene/LogsVolumeActions.tsx
+++ b/src/Components/ServiceScene/LogsVolumeActions.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+
+import { SceneComponentProps, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
+import { usePluginComponent } from '@grafana/runtime';
+import { VAR_LABELS } from 'services/variables';
+import { getAdHocFiltersVariable } from 'services/variableGetters';
+import { AdHocVariableFilter } from '@grafana/data';
+import { getDataSource } from 'services/scenes';
+
+interface LogsVolumeActionsState extends SceneObjectState {}
+
+export class LogsVolumeActions extends SceneObjectBase<LogsVolumeActionsState> {
+  static Component = Component;
+}
+
+type SelectorFilterItem = Pick<AdHocVariableFilter, 'key' | 'operator' | 'value'>;
+
+type TemporaryExemptionsButtonProps = {
+  dataSourceUid?: string;
+  selectorFilters?: SelectorFilterItem[];
+};
+
+function Component({ model }: SceneComponentProps<LogsVolumeActions>) {
+  const { component: TemporaryExemptionsButton, isLoading } = usePluginComponent<TemporaryExemptionsButtonProps>(
+    'grafana-adaptivelogs-app/temporary-exemptions-button/v1'
+  );
+
+  const labelsVar = getAdHocFiltersVariable(VAR_LABELS, model);
+  const { filters } = labelsVar.useState();
+  const selectorFilters = filters.map(({ key, operator, value }: AdHocVariableFilter) => ({ key, operator, value }));
+
+  const dataSourceUid = getDataSource(model);
+
+  if (isLoading || !TemporaryExemptionsButton) {
+    return null;
+  }
+
+  return <TemporaryExemptionsButton dataSourceUid={dataSourceUid} selectorFilters={selectorFilters} />;
+}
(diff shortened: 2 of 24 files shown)


src/Components/ServiceScene/LogsVolumeActions.tsx
```diff
@@ -0,0 +1,49 @@
...
+  static Component = Component;
+}
+
+type SelectorFilterItem = Pick<AdHocVariableFilter, 'key' | 'operator' | 'value'>;
+
+type TemporaryExemptionsProps = {
+  /** Currently selected data source */
+  dataSourceUid?: string;
+  /** The selector fields that we are considering temporary exemptions for */
+  selectorFilters?: SelectorFilterItem[];
```
svennergr, line 22: Just a nit, but can we call those `streamSelectors` or `labelFilters`?

svennergr (commented):
Don't we need to add this to `addedComponents`? https://grafana.com/developers/plugin-tools/reference/plugin-json#extensions-1:~:text=Description-,addedComponents,-object

Also, I'm not super sure about the implications adding this to the plugin.json's `dependencies`. @darrenjaneczek do you know more about this? https://grafana.com/developers/plugin-tools/reference/plugin-json#extensions says "A list of the exposed component ids that your plugin depends on.", which might be too restrictive?

svennergr:
Also @darrenjaneczek, how would one go about testing this? I don't even see `grafana-adaptivelogs-app/temporary-exemptions/v1` being used anywhere yet?

darrenjaneczek:
> Also @darrenjaneczek, how would one go about testing this? I don't even see `grafana-adaptivelogs-app/temporary-exemptions/v1` being used anywhere yet?

I haven't published my branch in a PR yet, it is still a work in progress.
Once I tidy a few things up, I will post a draft PR, and some quick instructions on getting it started up. 

Or we can consider a live session?

darrenjaneczek:
> Don't we need to add this to `addedComponents`? https://grafana.com/developers/plugin-tools/reference/plugin-json#extensions-1:~:text=Description-,addedComponents,-object
> 
> Also, I'm not super sure about the implications adding this to the plugin.json's `dependencies`. @darrenjaneczek do you know more about this? https://grafana.com/developers/plugin-tools/reference/plugin-json#extensions says "A list of the exposed component ids that your plugin depends on.", which might be too restrictive?

The equivalent of "addedComponents" will be done on the adaptivelogs app. Except instead of being an "added" component, it uses the newer explicity "exposed" component mechanism.

This is in the plugin.json of adaptivelogs.
(https://grafana.com/developers/plugin-tools/reference/plugin-json#extensions-1)
```
  "extensions": {
    "exposedComponents": [
      {
        "id": "grafana-adaptivelogs-app/temporary-exemptions/v1",
        "title": "Adaptive Logs Temporary Exemptions",
        "description": "Helps the user provide temporary exemptions to Adaptive Logs."
      }
    ]
  }
```

The plugin extensions are optional dependencies -- in the sense that the plugin that declares the dependency won't be compromised if the dependency isn't met -- however, it is mandatory that we express the dependency if we intend to use it. If we do not specify the exposed component as a dependency in explorelogs, then the infrastructure will refuse to let us use it in explorelogs. The plugin extension declaration system is _very particular_ about its metadata. If the adaptivelogs plugin is absent, then no component is returned by the hook, and explorelogs will carry on just fine. We get a similar effect if the adaptivelogs plugin is disabled. See the video below.

[dependency behavior.webm](https://github.com/user-attachments/assets/14e2d455-ba7d-443f-a358-997e3877ebd1)

src/Components/ServiceScene/LogsVolumeActions.tsx
```diff
@@ -0,0 +1,49 @@
...
+type SelectorFilterItem = Pick<AdHocVariableFilter, 'key' | 'operator' | 'value'>;
+
+type TemporaryExemptionsProps = {
+  /** A list of string identifiers to provide context cue of where this component is being embedded and how we might want to consider displaying it */
+  contextHints?: string[];
+  /** Currently selected data source */
+  dataSourceUid?: string;
+  /** The stream selector, broken down into a list of structured subselector filter items */
+  streamSelector?: SelectorFilterItem[];
+};
```
svennergr, line 25:
```suggestion
type StreamSelector = Pick<AdHocVariableFilter, 'key' | 'operator' | 'value'>;

type TemporaryExemptionsProps = {
  /** A list of string identifiers to provide context cue of where this component is being embedded and how we might want to consider displaying it */
  contextHints?: string[];
  /** Currently selected data source */
  dataSourceUid?: string;
  /** The stream selector, broken down into a list of structured subselector filter items */
  streamSelector?: StreamSelector[];
};
```

svennergr (approved):
LGTM

project-words.txt
```diff
@@ -508,3 +508,4 @@ linefilter
 queryfrontend
 timesync
 openmetrics
+adaptivelogs
```
svennergr, line 511:
```suggestion
adaptivelogs
subselector
```

PR #1021, opened 2025-01-29
Patterns: Not configured state not working

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/1005

We've had a few users complain as well in the feedback form that patterns are broken, it's likely that this is because the patterns not configured state is no longer showing.

This PR simply shows the error state if we get an error response from the API, and adds an e2e test to ensure we don't break it again in the future.


matyax:
In a local test, I block the request and it stays in a loading state.

imagen

gtk-grafana:
I'm not sure that's the best way to test if the patterns are configured in Loki, but it's easy enough to run the `onDataChange` method on activation. Should be fixed now.

matyax:
Cool! Yeah, at least it would be cool to show something if the request fails for whatever reason.

PR #1035, opened 2025-02-04
feat: open in Explore logs button

kozhuhds:
Exposes a `OpenInExploreLogsButton` component which takes properties and creates a LinkButton with a href to navigate to the Explore Logs app from outside.

### DIFF
--- a/src/Components/OpenInExploreLogsButton.tsx
+++ b/src/Components/OpenInExploreLogsButton.tsx
@@ -8,6 +8,7 @@ export interface OpenInExploreLogsButtonProps {
   from?: string;
   to?: string;
   returnToPreviousSource?: string;
+  renderButton?: (props: { href: string }) => React.ReactElement<any>;
 }
 
 export default function OpenInExploreLogsButton({
@@ -16,6 +17,7 @@ export default function OpenInExploreLogsButton({
   from,
   to,
   returnToPreviousSource,
+  renderButton,
 }: OpenInExploreLogsButtonProps) {
   const setReturnToPrevious = useReturnToPrevious();
 
@@ -45,11 +47,15 @@ export default function OpenInExploreLogsButton({
     return null;
   }
 
+  if (renderButton) {
+    return renderButton({ href });
+  }
+
   return (
     <LinkButton
       variant="secondary"
       href={href}
-      onMouseDown={() => setReturnToPrevious(returnToPreviousSource || 'previous')}
+      onClick={() => setReturnToPrevious(returnToPreviousSource || 'previous')}
     >
       Open in Explore logs
     </LinkButton>
--- a/src/Components/OpenInExploreLogsButton.tsx
+++ b/src/Components/OpenInExploreLogsButton.tsx
@@ -1,0 +1,57 @@
+import { useReturnToPrevious } from '@grafana/runtime';
+import { LinkButton } from '@grafana/ui';
+import React, { useMemo } from 'react';
+
+export interface OpenInExploreLogsButtonProps {
+  datasourceUid?: string;
+  labelMatchers: Array<{ name: string; value: string }>;
+  from?: string;
+  to?: string;
+  returnToPreviousSource?: string;
+}
+
+export function OpenInExploreLogsButton({
+  datasourceUid,
+  labelMatchers,
+  from,
+  to,
+  returnToPreviousSource,
+}: OpenInExploreLogsButtonProps) {
+  const setReturnToPrevious = useReturnToPrevious();
+
+  const href = useMemo(() => {
+    const mainLabel = labelMatchers[0];
+
+    if (!mainLabel) {
+      return null;
+    }
+
+    const url = new URL(
+      `${window.location.origin}/a/grafana-lokiexplore-app/explore/${mainLabel.name}/${mainLabel.value}/logs`
+    );
+
+    datasourceUid && url.searchParams.set('var-datasource', datasourceUid);
+    from && url.searchParams.set('from', from);
+    to && url.searchParams.set('to', to);
+
+    labelMatchers.forEach((labelMatcher) => {
+      url.searchParams.append('var-filters', `${labelMatcher.name}|=|${labelMatcher.value}`);
+    });
+
+    return url.toString();
+  }, [datasourceUid, from, to, labelMatchers]);
+
+  if (!href) {
+    return null;
+  }
+
+  return (
+    <LinkButton
+      variant="secondary"
+      href={href}
+      onClick={() => setReturnToPrevious(returnToPreviousSource || 'previous')}
+    >
+      Open in Explore logs
+    </LinkButton>
+  );
+}
--- a/src/Components/OpenInExploreLogsButton.tsx
+++ b/src/Components/OpenInExploreLogsButton.tsx
@@ -10,7 +10,7 @@ export interface OpenInExploreLogsButtonProps {
   returnToPreviousSource?: string;
 }
 
-export function OpenInExploreLogsButton({
+export default function OpenInExploreLogsButton({
   datasourceUid,
   labelMatchers,
   from,
@@ -49,7 +49,7 @@ export function OpenInExploreLogsButton({
     <LinkButton
       variant="secondary"
       href={href}
-      onClick={() => setReturnToPrevious(returnToPreviousSource || 'previous')}
+      onMouseDown={() => setReturnToPrevious(returnToPreviousSource || 'previous')}
     >
       Open in Explore logs
     </LinkButton>
--- a/src/plugin.json
+++ b/src/plugin.json
@@ -68,6 +68,13 @@
         "description": "Open current query in the Explore Logs view"
       }
     ],
+    "exposedComponents": [
+      {
+        "id": "grafana-lokiexplore-app/open-in-explore-logs-button/v1",
+        "title": "Open in Explore logs button",
+        "description": "A button that opens a logs view in the Explore logs app."
+      }
+    ],
     "extensionPoints": [
       {
         "id": "grafana-lokiexplore-app/metric-exploration/v1"
--- a/src/services/extensions/exposedComponents.tsx
+++ b/src/services/extensions/exposedComponents.tsx
@@ -1,10 +1,27 @@
-import { OpenInExploreLogsButton } from 'Components/OpenInExploreLogsButton';
+import { LinkButton } from '@grafana/ui';
+import { OpenInExploreLogsButtonProps } from 'Components/OpenInExploreLogsButton';
+import React, { lazy, Suspense } from 'react';
+const EntityAssertionsWidget = lazy(() => import('Components/OpenInExploreLogsButton'));
+
+function SuspendedOpenInExploreLogsButton(props: OpenInExploreLogsButtonProps) {
+  return (
+    <Suspense
+      fallback={
+        <LinkButton variant="secondary" disabled>
+          Open in Explore logs
+        </LinkButton>
+      }
+    >
+      <EntityAssertionsWidget {...props} />
+    </Suspense>
+  );
+}
 
 export const exposedComponents = [
   {
     id: `grafana-lokiexplore-app/open-in-explore-logs-button/v1`,
     title: 'Open in Explore logs button',
     description: 'A button that opens a logs view in the Explore logs app.',
-    component: OpenInExploreLogsButton,
+    component: SuspendedOpenInExploreLogsButton,
   },
 ];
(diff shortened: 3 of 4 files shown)


src/Components/OpenInExploreLogsButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const href = useMemo(() => {
+    const mainLabel = labelMatchers[0];
+
+    if (!mainLabel) {
+      return null;
+    }
+
+    const url = new URL(
+      `${window.location.origin}/a/grafana-lokiexplore-app/explore/${mainLabel.name}/${mainLabel.value}/logs`
```
gtk-grafana, line 32:
```suggestion
      `${window.location.origin}/a/grafana-lokiexplore-app/explore/${mainLabel.name}/${replaceSlash(mainLabel.value)}/logs`
```

src/services/extensions/exposedComponents.tsx
```diff
@@ -0,0 +1,27 @@
+import { LinkButton } from '@grafana/ui';
+import { OpenInExploreLogsButtonProps } from 'Components/OpenInExploreLogsButton';
+import React, { lazy, Suspense } from 'react';
+const EntityAssertionsWidget = lazy(() => import('Components/OpenInExploreLogsButton'));
```
gtk-grafana, line 4: We still have to include the entire `'Components/OpenInExploreLogsButton'` component in the main bundle since you load the interface from this same component. Maybe move the `OpenInExploreLogsButtonProps` into a new file?

src/Components/OpenInExploreLogsButton.tsx
gtk-grafana: Otherwise a label value with a slash, like loki/distributor will break the routing

svennergr: This will fail if Grafana runs with a suburl. I think 
```suggestion
      `${config.appSubUrl}/a/grafana-lokiexplore-app/explore/${mainLabel.name}/replaceSlash(${mainLabel.value})/logs`
```
with `config` from `@grafana/runtime` should do the trick, though.

src/Components/OpenInExploreLogsButton/types.ts
```diff
@@ -0,0 +1,8 @@
+export interface OpenInExploreLogsButtonProps {
+  datasourceUid?: string;
+  labelMatchers: Array<{ name: string; value: string }>;
```
ifrost, line 3: In theory you can also use [AbstractLabelMatcher](https://github.com/grafana/grafana/blob/fb79be4a4330ae765ce9a4d386ce0c0ccdd160aa/packages/grafana-data/src/types/query.ts#L45) from grafana/data. It's `internal` but it's been there for a while and we can promote it to `alpha`

src/services/extensions/exposedComponents.tsx
```diff
@@ -0,0 +1,27 @@
...
+          Open in Explore logs
+        </LinkButton>
+      }
+    >
+      <OpenInExploreLogsButton {...props} />
+    </Suspense>
+  );
+}
+
+export const exposedComponents = [
```
ifrost, line 20: Have you considered creating a single component that allows linking to any app that register an extension point instead of having single button specific to single app? Something similar to https://github.com/grafana/explore-profiles/blob/62720ad1cbd1346c772a521c133988d043851f55/src/links.ts#L42

src/Components/OpenInExploreLogsButton/types.ts
kozhuhds: yeah, I thought about it but it has a special operator value that requires additional mapping to `=` and `!=` syntax, so I decided to just go with this option for now

gtk-grafana (dismissed):
The PR looks fine, but I'm not sure how to test this out locally, can we at least add some test coverage so this isn't inadvertently broken in a future PR? 

Unit tests verifying that the href is created correctly should be fine

src/Components/OpenInExploreLogsButton/OpenInExploreLogsButton.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      to: 'now',
+    };
+
+    render(<OpenInExploreLogsButton {...props} />);
+
+    const linkButton = screen.getByRole('link', { name: /open in explore logs/i });
+    expect(linkButton).toBeInTheDocument();
+    expect(linkButton).toHaveAttribute(
+      'href',
+      'http://localhosta/grafana-lokiexplore-app/explore/job/test-job/logs?var-datasource=test-datasource&from=now-1h&to=now&var-filters=job%7C%3D%7Ctest-job'
```
gtk-grafana, line 36:
```suggestion
      'http://localhost/a/grafana-lokiexplore-app/explore/job/test-job/logs?var-datasource=test-datasource&from=now-1h&to=now&var-filters=job%7C%3D%7Ctest-job'
```

src/Components/OpenInExploreLogsButton/OpenInExploreLogsButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const url = new URL(
+      `${config.appSubUrl || config.appUrl}a/grafana-lokiexplore-app/explore/${mainLabel.name}/${mainLabel.value}/logs`
+    );
+
+    datasourceUid && url.searchParams.set('var-datasource', datasourceUid);
+    from && url.searchParams.set('from', from);
+    to && url.searchParams.set('to', to);
+
+    labelMatchers.forEach((labelMatcher) => {
+      url.searchParams.append('var-filters', `${labelMatcher.name}|=|${replaceSlash(labelMatcher.value)}`);
```
gtk-grafana, line 33: Is the calling application responsible for escaping label value content? Is only the equals operator supported?

gtk-grafana: The Explore Logs application assumes that the content of variables is not escaped, but this means you cannot pass in regex operations (unless you specify the content is pre-escaped, look at the `addAdHocFilterUserInputPrefix` method in `variables.ts`), everything in the value will be interpolated as a literal char to match, regex operation or not.

gtk-grafana: So this should be fine until a situation arises where you want to pass in a label value like `cluster.+` and expect it to match all label values starting with `cluster`

gtk-grafana: Regex support was added in the UI for labels and fields in https://github.com/grafana/explore-logs/pull/1030 (very recently).

kozhuhds: kk, then it makes sense to add `AbstractLabelMatcher`, will do

kozhuhds: @gtk-grafana not sure we need to escape value here, I wrote some tests that cover regex cases and it seems like the explore logs app can process these kinds of encoded URLs without any issues.

gtk-grafana: @kozhuhds as long as the chars in the label values are interpreted as literal chars, everything should work as expected, but if you want to support regex chars then I think you'll need to escape the content and add the prefix that denotes a value is pre-escaped

kozhuhds: @gtk-grafana I found some similar code and updated the place accordingly so now it handles all the cases even with pipes etc, thank you

src/Components/OpenInExploreLogsButton/OpenInExploreLogsButton.test.tsx
```diff
@@ -34,7 +34,7 @@ describe('OpenInExploreLogsButton', () => {
     expect(linkButton).toBeInTheDocument();
     expect(linkButton).toHaveAttribute(
       'href',
-      'http://localhost:3000/a/grafana-lokiexplore-app/explore/job/test-job/logs?var-datasource=test-datasource&from=now-1h&to=now&var-filters=job%7C%3D%7Ctest-job'
+      'http://localhost:3000/a/grafana-lokiexplore-app/explore/job/test-job/logs?var-datasource=test-datasource&from=now-1h&to=now&var-filters=job%7C%3D%7C__CV%CE%A9__test-job%2Ctest-job'
```
gtk-grafana, line 37: If you want to export the `encodeFilter` method from links.test.ts it might help make this a bit easier to read/maintain

src/Components/OpenInExploreLogsButton/OpenInExploreLogsButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    const url = new URL(
+      `${config.appSubUrl || config.appUrl}a/grafana-lokiexplore-app/explore/${mainLabel.name}/${replaceSlash(
+        mainLabel.value
+      )}/logs`
+    );
+
+    datasourceUid && url.searchParams.set('var-datasource', datasourceUid);
+    from && url.searchParams.set('from', from);
+    to && url.searchParams.set('to', to);
```
svennergr, line 45: Just a nit, but can we move this into proper `if` conditions?

src/Components/OpenInExploreLogsButton/OpenInExploreLogsButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    labelMatchers.forEach((labelMatcher) => {
+      url.searchParams.append(
+        'var-filters',
```
svennergr, line 49: Also nit, but for `filters` we could import `VAR_LABELS` from `services/variables`. That would make it a bit more resilient.

src/Components/OpenInExploreLogsButton/OpenInExploreLogsButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+const operatorMap = {
+  [AbstractLabelOperator.Equal]: LabelFilterOp.Equal,
+  [AbstractLabelOperator.NotEqual]: LabelFilterOp.NotEqual,
+  [AbstractLabelOperator.EqualRegEx]: LabelFilterOp.RegexEqual,
+  [AbstractLabelOperator.NotEqualRegEx]: LabelFilterOp.RegexNotEqual,
+};
+
+export default function OpenInExploreLogsButton({
+  datasourceUid,
+  labelMatchers,
```
svennergr, line 18: Since we only set the `VAR_LABELS` variable, can we rename this to `streamSelectors`? `VAR_LABELS` will only be used for indexed labels/stream selectors.

src/services/extensions/exposedComponents.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { OpenInExploreLogsButtonProps } from 'Components/OpenInExploreLogsButton/types';
+import React, { lazy, Suspense } from 'react';
+const OpenInExploreLogsButton = lazy(() => import('Components/OpenInExploreLogsButton/OpenInExploreLogsButton'));
+
+function SuspendedOpenInExploreLogsButton(props: OpenInExploreLogsButtonProps) {
+  return (
+    <Suspense
+      fallback={
+        <LinkButton variant="secondary" disabled>
+          Open in Explore logs
```
svennergr, line 11:
```suggestion
          Open in Explore Logs
```

src/plugin.json
```diff
@@ -68,6 +68,13 @@
         "description": "Open current query in the Explore Logs view"
       }
     ],
+    "exposedComponents": [
+      {
+        "id": "grafana-lokiexplore-app/open-in-explore-logs-button/v1",
+        "title": "Open in Explore logs button",
```
svennergr, line 74:
```suggestion
        "title": "Open in Explore Logs button",
```

src/plugin.json
```diff
@@ -68,6 +68,13 @@
...
+        "description": "A button that opens a logs view in the Explore logs app."
```
svennergr, line 75:
```suggestion
        "description": "A button that opens a logs view in the Explore Logs app."
```

src/Components/OpenInExploreLogsButton/OpenInExploreLogsButton.tsx
```diff
@@ -0,0 +1,92 @@
...
```
gtk-grafana, line 27: i.e.

```suggestion
  streamSelectors,
```

gtk-grafana (approved):

PR #1040, opened 2025-02-05
E2E fix: add more deterministic pod value to generator, update e2e

gtk-grafana:
I made some assumptions about the format and count of pod label values, which were being randomly generated.
Updating the `pod` values for tempo-ingester in the generator to have a more consistent and differentiable value format, so we can assert that metadata regex is working as expected.


matyax (approved):

gtk-grafana:
@svennergr can you take a look when you're back in office?

PR #1045, opened 2025-02-07
Filters: Expression builder

gtk-grafana:
Fixes https://github.com/grafana/explore-logs/issues/1044

Still need to clean up and test, but it's close!

### DIFF
--- a/generator/log/values.go
+++ b/generator/log/values.go
@@ -119,6 +119,7 @@ func ForAllClusters(namespace, svc model.LabelValue, cb func(model.LabelSet, pus
 				"__stream_shard__": model.LabelValue(shards[clusterInt%len(shards)]),
 				"namespace":        namespace,
 				"service_name":     svc,
+				"file":             "C:\\Grafana\\logs\\" + namespace + ".txt",
 			}, RandStructuredMetadata(string(svc), i))
 		}
 	}
--- a/src/services/ExpressionBuilder.ts
+++ b/src/services/ExpressionBuilder.ts
@@ -1,19 +1,31 @@
 import { AdHocFilterWithLabels } from '@grafana/scenes';
 import { AdHocVariableFilter } from '@grafana/data';
-import { isOperatorExclusive, isOperatorInclusive, isOperatorRegex } from './operators';
-import { FilterOp, LabelFilterOp } from './filterTypes';
+import { isOperatorExclusive, isOperatorInclusive, isOperatorNumeric, isOperatorRegex } from './operators';
+import { FilterOp, FilterOpType, LabelFilterOp, NumericFilterOp } from './filterTypes';
 import { Dictionary, groupBy, trim } from 'lodash';
 import { EMPTY_VARIABLE_VALUE, isAdHocFilterValueUserInput, stripAdHocFilterUserInputPrefix } from './variables';
+import { getValueFromFieldsFilter } from './variableGetters';
 
 type Key = string;
 type Value = string;
-type CombinedFiltersValues = { operator: LabelFilterOp; values: Value[] };
+type CombinedFiltersValues = { operator: LabelFilterOp | NumericFilterOp; values: Value[] };
 type CombinedFiltersValuesByKey = Record<Key, CombinedFiltersValues>;
 type CombinedOperatorFilters = Record<Key, Value>;
+type OperatorFilters = Record<Key, Value[]>;
 
 interface Options {
   debug?: boolean;
+
+  /**
+   * Separator between filters with different keys or operators
+   */
   filterSeparator?: string;
+
+  /**
+   * Separator between filters with the same keys and operators.
+   * Only applicable if joinMatchFilters is false?
+   */
+  valueSeparator?: string;
 
   /**
    * Prefix of the logQL expression
@@ -25,12 +37,18 @@ interface Options {
    * i.e. multiple filters with operators: '=' | '!=' => '=~' | '!~'
    */
   joinMatchFilters: boolean;
+
+  /**
+   * Sets if the value is JSON encoded
+   */
+  encodeFilters: boolean;
 }
 
-export class filtersToLogQL {
+export class ExpressionBuilder {
   private filters: AdHocFilterWithLabels[];
   private options: Options;
-  constructor(filters: AdHocFilterWithLabels[], options: Options = { joinMatchFilters: true }) {
+
+  constructor(filters: AdHocFilterWithLabels[], options: Options = { joinMatchFilters: true, encodeFilters: false }) {
     this.filters = filters;
     this.options = options;
   }
@@ -51,6 +69,8 @@ export class filtersToLogQL {
       }
     });
 
+    // @todo numeric
+
     return adHocFilters;
   }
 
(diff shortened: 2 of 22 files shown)


src/services/ExpressionBuilder.ts
```diff
@@ -0,0 +1,636 @@
+import { AdHocFilterWithLabels, sceneUtils } from '@grafana/scenes';
```
matyax, line 1: Can you please add unit tests for this new service?

src/services/extensions/links.test.ts
```diff
@@ -669,3 +810,7 @@ function getPath(options: {
     options.expectedFieldsUrlString ?? ''
   }${options.expectedLevelsFilterUrlString ?? ''}`;
 }
+
+function addAdHocFilterUserInputPrefixAndValueLabel(value: string, valueLabel = value) {
```
matyax, line 814: This seems to be always called with a single argument? I.e. no valueLabel. Also not a huge fan of this function's name.

src/services/ExpressionBuilder.ts
gtk-grafana: Tests were in query.test.ts, moved to `ExpressionBuilder.test.ts` in https://github.com/grafana/explore-logs/pull/1045/commits/98fd579bc89becd59dee4907a8727be2c01d32eb

matyax: Ah, perfect, thanks!

src/services/ExpressionBuilder.ts
```diff
@@ -0,0 +1,640 @@
...
+    }
+
+    return '';
+  }
+
+  /**
+   * Returns merged filters separated by pipe
+   */
+  public getMetadataExpr(): string {
+    this.options = {
```
matyax, line 134: Feel free to decide whether to chage it in this or a follow up PR, but I find this approach very unusual and misleading for anyone consuming this class: options are set by the constructor, but then when someone calls one of the methods, any option set when the instance was created is going to be silently overriden? Feels like it could use a refactor to change how these options are set and used. 

For example, instead of overwriting any previous option set by the constructor, make it an argument to `this.getLabelsExpr()`.

gtk-grafana: Agreed, fixed in https://github.com/grafana/explore-logs/pull/1045/commits/60497085e08f09a1149a5e75300810dd1f8a0401

gtk-grafana: Aaaaaand updated the type def in [9a91e17](https://github.com/grafana/explore-logs/pull/1045/commits/9a91e177fa785b5b818a075333cca455159d860f) :P

generator/log/values.go
```diff
@@ -119,6 +119,7 @@ func ForAllClusters(namespace, svc model.LabelValue, cb func(model.LabelSet, pus
 				"__stream_shard__": model.LabelValue(shards[clusterInt%len(shards)]),
 				"namespace":        namespace,
 				"service_name":     svc,
+				"file":             "C:\\Grafana\\logs\\" + namespace + ".txt",
```
gtk-grafana, line 122: @svennergr this is lovely for e2e tests and internal testing, but maybe not quite as great on play since it makes the data more nonsensical. 

Is it worth adding a command line flag to gate this label so we can keep it out of play?

src/services/ExpressionBuilder.ts
matyax: Cool, thanks! I just checked and it seems that ExpressionBuilder is not being instantiated with custom options, so maybe remove them from the constructor? I still feel that exposing them through the constructor and then overwriting them with calls to different methods defeats the purpose of them being instance options. Maybe it should be something internal that changes depending on what expression needs to be generated.
Not a huge deal, on the nitpicky side, but I think it would make a more consistent implementation, specially if stuff like the filterSeparator doesn't really change.

matyax:
I was looking at the URL parameters and manually editing to see how the app behaves and found that by changing the URL we can set the operators to be any arbitrary string:

imagen

Maybe we should narrow operators when read from the URL? 

The URL I used: http://localhost:3000/a/grafana-lokiexplore-app/explore/filename/-var-log-pods-pop-observability_agent-deployment-65dbf7c95-fmp4v_9ddfa05c-f7bf-4b8b-b960-70cdad372a37-agent-0.log/logs?from=now-3h&to=now&var-ds=fdtb1jrsqq3uof&var-filters=service_name%7C%3Cscript%3Ealert%28%27xss%27%29%3C%2Fscript%3E%7Cagent-deployment&var-filters=filename%7C%3D%7C%2Fvar%2Flog%2Fpods%2Fpop-observability_agent-deployment-65dbf7c95-fmp4v_9ddfa05c-f7bf-4b8b-b960-70cdad372a37%2Fagent%2F0.log&patterns=%5B%5D&var-fields=&var-levels=&var-metadata=&var-patterns=&var-lineFilterV2=&var-lineFilters=&urlColumns=%5B%5D&visualizationType=%22logs%22&displayedFields=%5B%5D&timezone=browser&var-all-fields=&sortOrder=%22Ascending%22&wrapLogMessage=true

generator/log/values.go
gtk-grafana: Created: https://github.com/grafana/explore-logs/issues/1049

src/services/ExpressionBuilder.ts
gtk-grafana: Yeah, but then we have to make the options object optional everywhere if it's not set in the constructor.

I would argue the class is built for more generic use-cases that we don't currently have implemented, and the public methods besides `getLabelsExpr` and `getJoinedLabelsFilters` are entirely unnecessary syntactic sugar methods that just keep the options out of the generic constructor.

TL;DR if you want to set options in the constructor, call `getLabelsExpr` or `getJoinedLabelsFilters` directly.

If we really don't like overwriting the options in the public helper methods then we should export default options for specific use-cases (METADATA_PARSING_OPTIONS, FIELD_PARSING_OPTIONS), and always call the generic methods.

matyax (approved):
Everything works as expected and couldn't find any issue with these changes, gw!

src/services/ExpressionBuilder.ts
matyax: Maybe most of what I see is a byproduct of being an instantiable class. I was just thinking that by the way it works, every public method could be static and have the filters and options be arguments.

gtk-grafana: Agreed that would be a simpler API, but I'm not sure the juice is worth the squeeze with the current complexity of the class. But something to look at when we migrate the pattern filters or whatever else gets added next.

src/services/query.ts
```diff
@@ -218,36 +176,7 @@ export function renderLogQLLineFilter(filters: AdHocFilterWithLabels[]) {
...
-  // If the filter value is an empty string, we don't want to wrap it in backticks!
-
-  return `${filter.key}${filter.operator}${value}`;
-}
-
-export function renderRegexLabelFilter(key: string, values: string[], operator: FilterOp) {
-  return `${key}${operator}"${values.join('|')}"`;
-}
-
+// @todo worth migrating into the ExpressionBuilder class?
```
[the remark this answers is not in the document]
gtk-grafana: Created https://github.com/grafana/explore-logs/issues/1052, since this is converting AppliedPatterns to logQL, and not ad-hoc filters to logQL, I think we need another abstraction so we don't violate the SRP

PR #1047, opened 2025-02-10
Logs: apply query direction in query runner

matyax:
When the core Logs Panel created a new request for infinite scrolling, the direction was missing, resulting in a query always with `backwards` direction. With the direction applied when creating the query runner, now the panel has that property defined and available when a new request is created for infinite scrolling.

Can be reproduced with `Oldest first` sort order.

Before:

https://github.com/user-attachments/assets/ba851fb0-f3ce-443b-9c3c-cdf18ddbba00

After:

https://github.com/user-attachments/assets/07c00507-f19c-4939-8282-dcb806f31b38


gtk-grafana (approved):
Code looks good and can confirm this fixes locally! Good catch.

Might be worth adding some e2e coverage for infinite scroll at some point, as right now if something were to change in the implementation it might not get caught.

PR #1057, opened 2025-02-11
Table: add manage columns button to table header

gtk-grafana:
Add button to show column management drawer as it's currently buried in the column header menu, which is not very discoverable. 

Fixes: https://github.com/grafana/explore-logs/issues/1051

image


matyax (commented):
I was going to comment on the button style, but it looks great, it's just the focused style that makes it look bigger than it is in your screenshot.

But the biggest thing that I noticed is that there's a huge delay between when I click the button and the columns manager appears, which seems to also cause a full re-render of the table. The whole interaction feels very heavy in terms of performance as it is.

https://github.com/user-attachments/assets/bb54e330-a6d4-4372-b150-ef21855250e0

gtk-grafana:
@matyax yeah, the state had to get pulled up out of react and into the scene, and whenever the scene state changes the whole table is re-rendered.

Maybe that's an argument for creating a new row of options above the able, but still with the react component (which would look a bit different then the logs panel)

src/Components/Table/Context/TableColumnsContext.tsx
```diff
@@ -201,7 +208,7 @@ export const TableColumnContextProvider = ({
         filteredColumns,
         columns,
         setColumns: handleSetColumns,
-        visible: visible,
+        visible: isColumnManagementActive,
         setVisible: handleSetVisible,
         clearSelectedLine: handleClearSelectedLine,
```
matyax, line 213: This is passing a completely new object on any re-render, maybe this is causing the re-render I just noticed today? I wouldn't be surprised.

matyax:
Nevermind. What I noticed in my previous comment is not new. If we have time, it could be interesting to dig a bit on the root cause of these full re-renders, maybe there's a low hanging fruit around missing memoization somewhere.

matyax (approved):
Another great improvement 👌

PR #1058, opened 2025-02-11
Service selection: Pagination

gtk-grafana:
Fixes: https://github.com/grafana/explore-logs/issues/1046

Desktop:
image
image

Mobile:
image


svennergr (approved):
I like it. I'm wondering if we want to move the page number to the URL as well. Not necessarily in this PR, but would be good, I think?

Anyway, this is LGTM.

matyax:
Ahhh this is very cool ☄️ , well done!

InaSong:
Excellent support! So quick and so cool!

PR #1074, opened 2025-02-13
Multiple include: Levels & Labels

gtk-grafana:
Providing the ability to add multiple include filters in a value breakdown.

image

Related:
* [UX: Add more prominent way to navigate back to the label name drilldown](https://github.com/grafana/explore-logs/issues/1086)
* [Multiple include: fields](https://github.com/grafana/logs-drilldown/pull/1091)


src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -185,18 +333,11 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
-      lastFilterEvent: undefined,
-    });
-    navigateToDrilldownPage(PageSlugs.labels, sceneGraph.getAncestor(this, ServiceScene));
-  }
-
   private build(): LayoutSwitcher {
     const variable = getLabelGroupByVariable(this);
     const variableState = variable.state;
-    const labelBreakdownScene = sceneGraph.getAncestor(this, LabelBreakdownScene);
     const tagKey = String(variableState?.value);
```
gtk-grafana, line 339: replace with getTagKey()

src/Components/ServiceScene/Breakdowns/Panels/ValueSummary.tsx
```diff
@@ -47,6 +63,14 @@ export class ValueSummaryPanelScene extends SceneObjectBase<ValueSummaryPanelSce
     const viz = buildValueSummaryPanel(this.state.title, { levelColor: this.state.levelColor });
     const height = getValueSummaryHeight(collapsed);
 
+    // @todo remove after fields support: tmp gate for fields
+    const key = this.state.key;
```
gtk-grafana, line 67:
```suggestion
    const key = this.state.tagKey;
```

src/Components/ServiceScene/Breakdowns/Panels/ValueSummary.tsx
```diff
@@ -78,6 +102,110 @@ export class ValueSummaryPanelScene extends SceneObjectBase<ValueSummaryPanelSce
...
     );
   }
+
+  /**
+   * Syncs legend with labels
+   */
+  private extendTimeSeriesLegendBus = (context: PanelContext, key: string) => {
+    const $data = sceneGraph.getData(this);
+    const dataFrame = $data.state.data?.series;
+    // @todo after fields support
```
gtk-grafana, line 112: cleaned up in child branch

src/services/ExpressionBuilder.ts
```diff
@@ -603,22 +623,31 @@ export class ExpressionBuilder {
    * Groups all filters by operator and key
    */
   private groupFiltersByKey(filters: AdHocVariableFilter[]): Record<FilterOpType, Dictionary<AdHocFilterWithLabels[]>> {
-    const positiveMatch = filters.filter(
+    let filteredFilters: AdHocVariableFilter[] = filters.filter((f) => !this.options.ignoreKeys?.includes(f.key));
+
+    // We need at least one inclusive filter
+    if (this.options.type === 'indexed') {
+      if (filteredFilters.length < 1) {
+        filteredFilters = filters;
```
gtk-grafana, line 631: So if we have multiple include filters containing the exclude key, we would run the query with all of them. Maybe this is a reason to exclude the "primary" filter from breakdown views so we know that we're never excluding the only include filter.

But folks might want write a regex include e.g. `cluster=~"us-.+"`, and then want to exclude stuff in the UI... but right now I believe this will behave as the current version, including will hide all other options, and excluding will remove that panel from the view (if they have no other inclusion label filters).

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
gtk-grafana: Cleaned up in child branch

gtk-grafana:
Found a bug: When adding a field, the parser is not updated, leading to no results view when there are logs.

gtk-grafana:
Found another bug: When adding or removing another filter in the combobox in the value breakdown when there is an active filter for the current breakdown value, there is a duplicate logsPanelQuery and the current value breakdown query isn't excluding the key. 

Looks like it was introduced by the "fix" in https://github.com/grafana/logs-drilldown/commit/cf5751d31c672ec4fbb1f36b0e9cdc53d924e588, maybe we need to move the parser to a variable?

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -24,14 +30,24 @@ import {
...
+  isIncluded?: boolean;
+  isExcluded?: boolean;
 }
 
 export class AddFilterEvent extends BusEventBase {
-  constructor(public operator: FilterType | NumericFilterType, public key: string, public value: string) {
+  constructor(
+    public source: 'legend' | 'filterButton' | 'variable',
+    public operator?: FilterType | NumericFilterType,
+    public key?: string,
```
matyax, line 49:
```suggestion
    public operator: FilterType | NumericFilterType,
    public key: string,
```
As far as I can see, these should be defined.

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -60,14 +73,45 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
+
+    // Run query on activate
+    this.runQuery();
+    this.setSubs();
+  }
+
+  /**
+   * Set variable & event subscriptions
+   */
+  private setSubs() {
```
matyax:
```suggestion
  private setSubscriptions() {
```

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
```diff
@@ -77,47 +121,151 @@ export class LabelValuesBreakdownScene extends SceneObjectBase<LabelValueBreakdo
...
+            const filterExpression = renderLevelsFilter(newState.filters, [this.getTagKey()]);
+
+            if (partialFilterExpression !== filterExpression) {
+              const queryRunner = this.getSceneQueryRunner();
+              queryRunner?.runQueries();
+            }
+          }
+        })
+      );
+    }
```
matyax, line 214: First, double checking if these conditionals arent inverted. If the tag key is level, subscribe to labels?

Then, it looks like it could be unified as:

```suggestion
    const variable = this.getTagKey() === LEVEL_VARIABLE_VALUE ? getLabelsVariable(this) : getLevelsVariable(this);
    const renderer = this.getTagKey() === LEVEL_VARIABLE_VALUE ? renderLogQLLabelFilters : renderLevelsFilter;
    this._subs.add(
      variable.subscribeToState(async (newState, prevState) => {
        if (!areArraysEqual(newState.filters, prevState.filters)) {
          // Check to see if excluding this label changes the query string before running
          // If the filter change was for the label we're looking at, there's no need to re-run the query
          const partialFilterExpression = this.removeValueLabelFromVariableInterpolation();
          const filterExpression = renderer(newState.filters, [this.getTagKey()]);

          if (partialFilterExpression !== filterExpression) {
            const queryRunner = this.getSceneQueryRunner();
            queryRunner?.runQueries();
          }
        }
      })
    );
```

src/Components/ServiceScene/Breakdowns/Panels/ValueSummary.tsx
```diff
@@ -9,17 +10,32 @@ import {
...
+import { FilterType } from '../AddToFiltersButton';
+import { toggleLabelFromFilter } from '../../../../services/labels';
 
 const SUMMARY_PANEL_SERIES_LIMIT = 100;
 
 interface ValueSummaryPanelSceneState extends SceneObjectState {
   body?: SceneFlexLayout;
   title: string;
   levelColor?: boolean;
+  // @todo make required after adding field support
```
matyax: 💡

src/services/ExpressionBuilder.ts
```diff
@@ -41,14 +41,27 @@ interface Options {
...
+
+  /**
+   * Keys to ignore
+   */
+  ignoreKeys?: string[];
+
+  /**
+   * Filter type
+   */
+  type: 'indexed' | 'field';
```
matyax, line 53: If you change `type`  to `filterType`, both comments would be not needed. The rest of the code should be slightly more verbose too.

matyax (dismissed):
Great work! I'd appreciate if you can address/respond to the open comments before merging. Thanks!

src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1074/commits/8c3deb8b188c704926b25ba7a014c8c54aa148a1

src/Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene.tsx
gtk-grafana: Yeah good call, much cleaner!

Added in https://github.com/grafana/logs-drilldown/pull/1074/commits/e568e30b6cc24d3c2669a4808e5fbeddfccdbeee

src/services/ExpressionBuilder.ts
gtk-grafana: Changed in https://github.com/grafana/logs-drilldown/pull/1074/commits/9c114c7293e92ec6c3101d3e9db392330d97fba9

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -524,10 +539,15 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
 
   private subscribeToTimeRange() {
     return sceneGraph.getTimeRange(this).subscribeToState(() => {
-      this.state.$patternsData?.runQueries();
-      this.state.$detectedLabelsData?.runQueries();
-      this.state.$detectedFieldsData?.runQueries();
-      this.state.$logsCount?.runQueries();
+      // Hack: Wait for expressions in value breakdowns to run and then reset the expression builders,
+      // if they are both executed on the same callstack one of the queries will be interpolated incorrectly.
+      setTimeout(() => {
```
matyax: :(

tests/exploreServicesJsonMixedBreakDown.spec.ts
```diff
@@ -9,7 +9,7 @@ const jsonFmtFieldName = 'status';
 const metadataFieldName = 'pod';
 const serviceName = 'nginx-json-mixed';
 // const levelName = 'cluster'
-test.describe('explore nginx-json-mixed breakdown pages ', () => {
+test.describe.only('explore nginx-json-mixed breakdown pages ', () => {
```
matyax, line 12:
```suggestion
test.describe('explore nginx-json-mixed breakdown pages ', () => {
```

matyax (approved):
Tested the usual happy paths and looks ready to go 👌

PR #1077, opened 2025-02-14
Service Selection Pagination: Generate options based on total count

matyax:
This PR uses the totalCount of values to generate the options for the pagination dropdown, using the total count to determine the amount of options and syncing the visible value to the actual value. For example 60 services will generate 3 options, 40 will generate up to 40, and less than 20 will show the actual number. Internally it still uses the same values, but the visible values change according to the total.

For example:

imagen

imagen

imagen

https://github.com/user-attachments/assets/f8fb84b8-d6d8-4932-bdc4-b07c1f8b3b17


gtk-grafana (approved):
Looks good, works well locally!

PR #1079, opened 2025-02-17
Logs Panel: Display error message when logs fail to load

matyax:
This PR subscribes to the logs query and displays an error component when the log query fails, with the objective of improving how we surface errors to the user, making it clear that the query failed and providing a escape hatch to clear all filters.

Before:

After:

### DIFF
--- a/src/Components/ServiceScene/LogsPanelError.tsx
+++ b/src/Components/ServiceScene/LogsPanelError.tsx
@@ -28,5 +28,5 @@ function getMessageFromError(error: string) {
     return 'Logs could not be retrieved due to invalid filter parameters. Please review your filters and try again.';
   }
 
-  return 'Logs could not be retrieved. Please review your filters or try a different time interval.';
+  return 'Logs could not be retrieved. Please review your filters or try a different time range.';
 }
--- a/src/Components/ServiceScene/LogsPanelError.tsx
+++ b/src/Components/ServiceScene/LogsPanelError.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import { GrotError } from 'Components/GrotError';
+import { Button } from '@grafana/ui';
+
+interface Props {
+  clearFilters(): void;
+  error: string;
+}
+
+export const LogsPanelError = ({ clearFilters, error }: Props) => {
+  return (
+    <GrotError>
+      <div>
+        <p>
+          <strong>No logs found.</strong>
+        </p>
+        <p>{getMessageFromError(error)}</p>
+        <Button variant="secondary" onClick={clearFilters}>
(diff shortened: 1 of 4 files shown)


src/Components/ServiceScene/LogsPanelError.tsx
```diff
@@ -0,0 +1,32 @@
...
+    </GrotError>
+  );
+};
+
+function getMessageFromError(error: string) {
+  if (error.includes('parse error')) {
+    return 'Logs could not be retrieved due to invalid filter parameters. Please review your filters and try again.';
+  }
+
+  return 'Logs could not be retrieved. Please review your filters or try a different time interval.';
```
svennergr, line 31:
```suggestion
  return 'Logs could not be retrieved. Please review your filters or try a different time range.';
```

svennergr (commented):
I think it should replace the whole logs tab, or even higher level. 
There's not really a need to show log volume histogram, when the query expression errors out.

image

matyax:
Definitely. I will also look for a couple other possible error responses to map to a readable message. Feel free to suggest if you remember any in particular.

matyax:
@svennergr What do you think about collapsing Volume on error?

zizzpudding:
Is there a way to make the make a suggestion to recover from this error without clearing all? "Invalid filter parameters" is pretty cryptic.

I'm unsure if this is an issue with the line filter? or if it's returning an empty set because the line filter was invalid. 

In the first case. a simple error message "there are not logs that match your filter, clear all" 

In the second case, 
if there is a way to indicate which filter caused the erro ,highlight it (red border) and give a message to fix or clear that one (something like "contains invalid characters" etc. 

Also link to the some documentation for creating a Regex expression for Loki would be helpful.

gtk-grafana:
@zizzpudding Loki doesn't really tell us what part of the query failed, it just tells us that it was unable to parse the query. Also this PR is only addressing when we get an error back from Loki, not when there are no results.

We can follow up on client-side regex errors in line filters on this issue: [parse error](https://github.com/grafana/explore-logs/issues/989). Updating the docs is good call out, I'll add a line-item to that epic.

gtk-grafana (approved):
Looks good, the only unexpected thing is that when I have the volume collapsed before running into an error, it gets uncollapsed after fixing the error.

Still a lot we can do here, but still worth merging

matyax:
> Looks good, the only unexpected thing is that when I have the volume collapsed before running into an error, it gets uncollapsed after fixing the error.

You read my mind. I was just about to do that.

> Still a lot we can do here, but still worth merging

Agree. This is just the initial step to better handle and surface errors to the user.

PR #1082, opened 2025-02-18
Query runner: Allow direction to be updated when sort order changes

matyax:
Follow up to https://github.com/grafana/explore-logs/pull/1047 to allow the direction to be updated according to the current sort order:

Before:

https://github.com/user-attachments/assets/7a2534c9-4222-4a5e-a31c-c997288aa4ab

After:

https://github.com/user-attachments/assets/534d87f4-3694-42bb-9d6a-3b4a4cd11857


matyax:
I'll try to cover this with an e2e so we don't regress again.

matyax:
Just verified that the e2e update catches regressions with waitForRequest:

imagen

tests/fixtures/explore.ts
```diff
@@ -176,15 +176,18 @@ export class ExplorePage {
...
+        (resp) => {
+          const post = resp.request().postDataJSON();
+          const queries = post?.queries as LokiQuery[];
+          if (queries && test(queries[0])) {
+            callback(queries[0]);
+            return true;
+          }
+          return false;
+        },
+        { timeout: 30000 }
```
gtk-grafana, line 189: Does this timeout change anything, I think 30s is the default?

Issue #1053, opened 2025-02-10
Filters: Support queries from Explore containing "or"


gtk-grafana:
Impacts line filters and field filters.

As of (https://github.com/grafana/logs-drilldown/pull/1091), for positive field filters, it should just work as expected if the filters are added to the fields variable query string.

For negative field filters and line filters, we'll either need to be convert to a regex filter, or "or" needs to be added to the UI.

PR #1091, opened 2025-02-20
Multiple include: Fields & Metadata

gtk-grafana:
Adding support for multiple includes in the value breakdown UI.
Parent branch: https://github.com/grafana/logs-drilldown/pull/1074

image


matyax (commented):
Looks great so far. A few things that I noticed:

- Going to the fields/labels breakdown triggers an unnecessary logsPanelQuery, just by opening the tab. It's the first query that is being executed.

https://github.com/user-attachments/assets/523c139a-dbda-42b6-b04b-9ec1688ba245

- The active state of filters in Log Details in inconsistent. Sometimes it shows as an active filter, other it doesnt. When clicked, it seems to remove filters but never add.

imagen

imagen

- When filtering for detected_level, applies the filter and selects the time series in Logs Volume, but it doesn't show in the top section.

imagen

- Removed the detected_level filter from log details, but it stayed in the top filters.

imagen

- It's not parsing multiple filters when coming from Explore

https://github.com/user-attachments/assets/1a66cbb9-e1b4-4f35-9d65-f5394ee4d346

gtk-grafana:
Thanks for taking a look @matyax, good catches all around.

> * Going to the fields/labels breakdown triggers an unnecessary logsPanelQuery, just by opening the tab. It's the first query that is being executed.

Yeah the logs panel query is not manually executed yet, it runs on every route/variable update in main right now as well.

Good catch with the logs panel filters, I'll take a look at this and the table filters as well here this morning.

> It's not parsing multiple filters when coming from Explore
https://github.com/grafana/logs-drilldown/issues/1053
Yeah, right now there's no way to specify an or relation in field/line filters in the UI

matyax:
> Yeah the logs panel query is not manually executed yet, it runs on every route/variable update in main right now as well.

Not sure I follow. The `logsPanelQuery` ran to get the logs, but it will be executed again just by going to a tab. Is that needed for something?

gtk-grafana:
I think (?) it's tech debt at this point from when we were parsing the labels/labelTypes from the log panel query, but it's possible those are still being used somewhere. I'll create an issue to convert that query runner to manually execute and we can investigate if that dataframe is still being used in the other scenes.

gtk-grafana:
@matyax logs panel should be fixed in https://github.com/grafana/logs-drilldown/pull/1091/commits/ce0d51587edfc757249703bb9706ca5f177eba89. I also fixed a bug where all negative filters were being generated with the "or" syntax (test coverage in https://github.com/grafana/logs-drilldown/pull/1091/commits/69b777f1b8939ae429e4291386f0e74fd231cdb4)

matyax (approved):
Fantastic 👌

matyax:
Saw the !level btw, it's okay for now, and more importantly, it works.

Issue #1098, opened 2025-02-24
Fliters: support multiple positive filters from explore queries


gtk-grafana:
In https://github.com/grafana/logs-drilldown/pull/1091, we added support for multiple include filters in the UI, let's make sure we support queries like this coming from explore as well.

PR #1103, opened 2025-02-25
Link extensions: Support multiple include filters

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1098


matyax (approved):
It does the trick 👌

Issue #1096, opened 2025-02-21
Detected level: Clear filters button is disabled


gtk-grafana:
Only the last include filter should be disabled, not detected_level.

<img width="441" alt="Image" src="https://github.com/user-attachments/assets/4f0a6d0f-77ab-4e7b-9c7d-37174dbc5ab9" />

PR #1105, opened 2025-02-25
fix: clear filters icon not working with detected_level

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1096


src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
```diff
@@ -299,6 +300,9 @@ export class SelectLabelActionScene extends SceneObjectBase<SelectLabelActionSce
 
   public clearFilters = (variableType: InterpolatedFilterType) => {
```
matyax, line 301: It'd be nice to add a regression so this doesn't happen again, and is one less thing to manually QA.

matyax (approved):
Works as expected 👌

PR #1107, opened 2025-02-26
Browser history on filter change

gtk-grafana:
Adding browser history events before setting filters to state, when triggered by a user action.
Note: This doesn't catch changes made directly in the combobox variables, but should cover everything? else: patterns, line filters, ([levels in another PR](https://github.com/grafana/logs-drilldown/pull/1108)), and panel filter buttons.

Related: https://github.com/grafana/logs-drilldown/issues/1109


matyax:
> Create issue to support for combobox variables

I don't think that's an expectation. Or at least that was my first reaction upon reading that.

Even with breakdown filters, now with multiple selection support, I don't think I would expect history support if I'm not navigating away.

matyax (approved):
Works as expected 👌

Issue #1110, opened 2025-02-26
Link extension: include/exclude empty fields displays encoded value in combobox


gtk-grafana:
When hitting "go queryless" with a query like:
```
{service_name="tempo-ingester"} | json | logfmt | msg=""
```

You'll see 

<img width="572" alt="Image" src="https://github.com/user-attachments/assets/a2d8d3dc-8282-42a5-9637-6b9557bf40b2" />

Looks like the valueLabels are missing, let's fix that

PR #1111, opened 2025-02-26
fix: empty label values missing labelValues

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1110


src/services/extensions/links.ts
```diff
@@ -83,6 +83,14 @@ export function stringifyAdHocValues(value?: string): string {
   return addAdHocFilterUserInputPrefix(replaceEscapeChars(value));
 }
 
+export function stringifyAdHocValueLabels(value?: string): string {
+  if (!value) {
+    return '""';
```
matyax, line 88:
```suggestion
    return EMPTY_VARIABLE_VALUE;
```

Issue #1116, opened 2025-02-27
AdHocVariables: Move custom value option to top


gtk-grafana:
<img width="686" alt="Image" src="https://github.com/user-attachments/assets/c2028975-2c50-479f-b3ac-e268447c6bc3" />

gtk-grafana:
Fixed with https://github.com/grafana/scenes/issues/1065

PR #1117, opened 2025-02-27
AdHocVariables: Move custom value option to top for regex operations

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1116


gtk-grafana:
I think we get this for free if scenes can update to use the `combobox` [component](https://developers.grafana.com/ui/latest/index.html?path=/story/forms-combobox--async-options-with-labels&args=createCustomValue:!true) instead of the deprecated select.

gtk-grafana:
Closing in favor of https://github.com/grafana/scenes/pull/1250

PR #1127, opened 2025-03-18
Logs errors: improve error messages

### LINKED ISSUE #1121: Log panel error state: more explicit messages
When anything goes wrong with the Logs query, we display a panel that says "No logs found", but the real reason behind the failure can be a wrongly generated query or other types of error such as "response larger than max message size".

When the "no logs found" message appears it can be very confusing to the users, specially if they're seeing logs in Logs Volume. So we should show something like:

- On error: Failed to retrieve logs 
- On "response larger tan max message size", a friendlier user version of this message.

```
{
    "results": {
        "logsPanelQuery": {
            "error": "response larger than the max message size (121109328 vs 104857600)",
            "errorSource": "plugin",
            "status": 500
        }
    }
}
```

matyax:
The current error state can be potentially confusing to the users by not being clear of the underlying situation. With these changes we're making a difference between:

- General query errors
- Response too large
- No data returned.

Error:

no data

No data:

query error

Fixes #1121

### DIFF
--- a/src/Components/ServiceScene/LogsPanelScene.tsx
+++ b/src/Components/ServiceScene/LogsPanelScene.tsx
@@ -127,7 +127,6 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
     const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
     this._subs.add(
       serviceScene.subscribeToState((newState, prevState) => {
-        console.log(newState.$data?.state.data);
         if (newState.$data?.state.data?.state === LoadingState.Error) {
           this.handleLogsError(newState.$data?.state.data);
         } else if (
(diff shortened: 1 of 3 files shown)


src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -126,8 +127,14 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
     const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
     this._subs.add(
       serviceScene.subscribeToState((newState, prevState) => {
+        console.log(newState.$data?.state.data);
```
svennergr, line 130:
```suggestion
```

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -151,9 +158,26 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
   }
 
   handleLogsError(data: PanelData) {
+    const errorResponse = data.errors?.length ? data.errors[0].message : data.error?.message;
```
svennergr, line 161: I think we should log the error response using our logger.

svennergr (approved):
Left two comments. Overall LGTM👍

src/Components/ServiceScene/LogsPanelScene.tsx
matyax: Great idea.

PR #1135, opened 2025-03-19
E2E: Address flakey test

gtk-grafana:
Looks like https://github.com/grafana/logs-drilldown/pull/1127 introduced some flake in this e2e test, although I'm confused as to how it ever worked, but in all of the failures locally and in CI the new empty logs view was being rendered.

Telling the test to run the logs panel query seems to have fixed it locally.


matyax (approved):
💯

PR #1151, opened 2025-03-31
JSON: DrillUp breadcrumbs

gtk-grafana:
Adds parent nodes of current drilldown as breadcrumbs in viz "header".
Tweaks the box shadow styles to better give the appearance of the "header" being higher on the z-axis.

image

Notes: 10lb of **** in 5lb bag problem is addressed with horizontal scroll, which is not a good solution, but if we expand the height of the header we lose the ability to make the timestamp sticky as we need an absolute offset. 
I figured this was a reasonable compromise for MVP.


src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -388,15 +434,50 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
+            size={'sm'}
+            onClick={() => jsonFiltersSupported && this.setNewRootNode(rootKeyPath)}
+            variant={'secondary'}
+            fill={'outline'}
+            disabled={!filters.length}
+            name={keyPath[0].toString()}
+          >
+            {this.getKeyPathString(keyPath, filters.length ? '' : ':')}
+          </Button>
+          {filters.length > 0 && <span className={breadCrumbDelimiter}>{'>'}</span>}
```
matyax, line 458: I would try what we use in the actual breadcrumb components which is `<Icon name="angle-right" />`

gtk-grafana: Good call, fixed in https://github.com/grafana/logs-drilldown/pull/1151/commits/78368ba2cd2ad450dbdbeb9a078e225073f47817

matyax (approved):
🚀 ☄️

Issue #1152, opened 2025-04-01
Case sensitive toggle causes uncaught error


matyax:
Clicking ![Image](https://github.com/user-attachments/assets/0073619f-c5bf-47c5-97fc-8c8793fd9eeb) causes an uncaught error.

![Image](https://github.com/user-attachments/assets/550d6001-6f87-447b-909d-5ea264455a91)

PR #1153, opened 2025-04-01
Line filter: Fix error being thrown when toggling case sensitivity with empty value

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1152


svennergr (approved):
LGTM. Does that mean the `AdHocFilterWithLabels` type in scenes is wrong? Because `filter.value` says it is `string`, not `string | undefined`

gtk-grafana:
@svennergr It sure looks that way as this function is called directly from the adhoc variable expressionBuilder callback, but the line filter is a custom renderer and does some non-standard ad-hoc variable stuff, so I would need to reproduce with a base element to be sure we're not sneaking in an undefined value somehow.

PR #1162, opened 2025-04-02
chore: (grafana 11.6) - update to latest grafana 11.6 and latest plugin libraries, remove comments

### LINKED ISSUE #921: Panel menus: Pull in core Grafana changes after 11.6 release
Pull in changes introduced in https://github.com/grafana/grafana/issues/96865.
Clean up wrapping element and css hacks added in https://github.com/grafana/explore-logs/pull/892

L2D2Grafana:
# Description
Fixes #921 

- Update grafana to 11.6
- Update grafana plugin libraries
- Add github actions from the latest create plugin to auto update grafana dependencies,
- Remove unused code
- Use grafana emptystate 
Screenshot 2025-04-02 at 9 19 32 AM


.github/workflows/bundle-stats.yml
```diff
@@ -0,0 +1,21 @@
+name: Bundle Stats
```
gtk-grafana, line 1: Oh nice, will this mean we can remove bundlewatch? Where do we set the bundle size limits?

package.json
```diff
@@ -102,11 +102,11 @@
   "dependencies": {
     "@bsull/augurs": "^0.6.0",
     "@emotion/css": "^11.10.6",
-    "@grafana/data": "^11.3.0",
+    "@grafana/data": "^11.6.0",
     "@grafana/lezer-logql": "^0.2.7",
-    "@grafana/runtime": "^11.3.0",
+    "@grafana/runtime": "^11.6.0",
```
gtk-grafana, line 107: Do we need to update the grafanaDependency version in the plugin.json now?

cypress.json
```diff
@@ -1,3 +0,0 @@
-{
-  "video": false
```
gtk-grafana, line 2: and stay dead!

.github/workflows/bundle-stats.yml
gtk-grafana: I remember hearing that this check will fail until it's been merged into main, that could be the download artifacts error?

gtk-grafana: Let's create an issue to remove bundlewatch once we verify that bundle stats are working as expected

src/Components/GrotError.tsx
```diff
@@ -1,54 +1,29 @@
...
-      display: 'flex',
-      justifyContent: 'center',
-      alignItems: 'center',
-    }),
-    wrap: css({
-      margin: '0 auto',
-    }),
-  };
-};
+import { useStyles2, EmptyState } from '@grafana/ui';
```
gtk-grafana, line 6: whoa the EmptyState is animated?? So cool! 
Can we create a follow-up issue to follow up to show the "clear filters" view instead if there are more filters then a single label filter?
image

image

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -117,16 +117,5 @@ const getStyles = (theme: GrafanaTheme2) => ({
     width: '100%',
     height: '100%',
     label: 'panel-wrapper-table',
-
```
gtk-grafana, line 120: There's still 1 more of these in the PanelMenu.tsx.
Also I think we'll need to remove all of the comments and add `.setShowMenuAlways(true)` to all of the panels
``` 
// .setShowMenuAlways(true)
```

Search for 11.5 in the code and they should all come up.

gtk-grafana: aaaand another one in PatternsFrameScene.tsx: 
```
    // Hide header on hover hack
    '.show-on-hover': {
      display: 'none',
    },
```

L2D2Grafana: Oh wow, I missed so many ty! Updated!
Should I also delete these functions in `TagKeysProviders.ts`
```
// @todo delete after min supported grafana is upgraded to >=11.6
  // see ced526b3e37baded9082ffc3c2378a21201801b6 before this all got messed up
  const fetchDetectedFieldsFn =
    (datasource &&
      typeof languageProvider.fetchDetectedFields === 'function' &&
      languageProvider.fetchDetectedFields.bind(languageProvider)) ||
    function (opts: DetectedFieldQueryOptions) {
      return fetchDetectedFields(datasource, opts);
    };
```

package.json
L2D2Grafana: Oh I thought you wanted grafana/ui, data, runtime updated as well. Should I revert all those updates?

src/Components/ServiceScene/LogsTableScene.tsx
L2D2Grafana: lol nm, found the issue you already created for this.

src/Components/GrotError.tsx
L2D2Grafana: https://github.com/grafana/logs-drilldown/issues/1164

.github/workflows/bundle-stats.yml
L2D2Grafana: https://github.com/grafana/logs-drilldown/issues/1165

L2D2Grafana: Removed it for now and updated the ticket.

package.json
gtk-grafana: No, I think this is correct, but if we are using assets from Grafana 11.6 that could cause issues if users are running Grafana < 11.6?

src/Components/NoLokiSplash.tsx
```diff
@@ -10,13 +10,11 @@ export const NoLokiSplash = () => {
...
     <div className={styles.wrap}>
       <div className={styles.graphicContainer}>
-        <SVG
-          src={
-            theme.isDark
-              ? `/public/plugins/grafana-lokiexplore-app/img/grot_loki.svg`
-              : `/public/plugins/grafana-lokiexplore-app/img/grot_loki.svg`
-          }
-        />
+        {React.createElement(SVG as any, {
```
gtk-grafana, line 13: I don't love the type assertion, plus let's get rid of this unnecessary ternary

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -280,42 +280,37 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
...
+      .setOption(
+        'prettifyLogMessage',
+        options.prettifyLogMessage ?? Boolean(getLogOption<boolean>('wrapLogMessage', false))
+      )
+      .setMenu(
+        new PanelMenu({
+          investigationOptions: { type: 'logs', getLabelName: () => `Logs: ${getPrettyQueryExpr(serviceScene)}` },
+        })
+      )
+      .setOption('showLogContextToggle', true)
```
gtk-grafana, line 305: Looks like the logs panel is still only showing the menu icon on hover
```suggestion
      .setOption('showLogContextToggle', true)
      .setShowMenuAlways(true)
```

package.json
gtk-grafana: Yeah, the panel menus are show on hover, and more importantly the fields combobox detected_fields call is broken on 11.3. The fields call might work in 11.4, but the panel menus will be broken until 11.5 or 11.6.

src/Components/NoLokiSplash.tsx
L2D2Grafana: Strange I was getting a typescript error but after `yarn install` it has gone away and I'm able to keep using <SVG>

gtk-grafana:
Looks like we missed `showMenuAlways={true}` in the LogsTableScene.tsx `PanelChrome` component, but other then that this looks good to go!

L2D2Grafana:
> showMenuAlways={true}

Ty! So I tried adding showMenuAlways={true} and for some reason that throws a typescript error that I need to define width and height. I see that the width and height are specified in TableWrap.tsx. I could move that up to the top level. The fun thing is it renders correct with type script error.

gtk-grafana:
> Ty! So I tried adding showMenuAlways={true} and for some reason that throws a typescript error that I need to define width and height. I see that the width and height are specified in TableWrap.tsx. I could move that up to the top level. The fun thing is it renders correct with type script error.

Good catch, that was my fault, I added the type to the PanelChrome `Collapsible` interface, but we probably wanted to add them to the `AutoSize` interface. You can @ts-expect-error for now instead of changing the implementation, and add a @todo in the code when a PR for https://github.com/grafana/grafana/issues/103486 is merged.

src/Components/ServiceScene/Breakdowns/Patterns/PatternsFrameScene.tsx
```diff
@@ -250,9 +250,5 @@ export function overrideToFixedColor(key: keyof typeof palette): FieldColor {
 const styles = {
   container: css({
     width: '100%',
-    // Hide header on hover hack
-    '.show-on-hover': {
```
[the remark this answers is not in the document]
L2D2Grafana: Put back!

Issue #971, opened 2024-12-20
Layout Scene: Reorganize variables


gtk-grafana:
With [line filters being moved](https://github.com/grafana/explore-logs/pull/963) into the variables in the header, we've got quite a lot going on when queries get more complex:

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/716c1354-4a02-4e21-b8d6-486a005b530d" />

While at least the variables are in the same order as the logQL query now, I think we should do a better job illustrating that the order of operations is a function of how loki will find the logs you are searching for.

Also things look "messy", and we could do a better job collapsing and expanding variables so the header doesn't take up so much vertical space.

Some things we discussed:
* Move labels "combobox" on it's own line
* Move the feedback link to the left of the time picker
* put datasource and datapicker/refresh on the same line.
* Adding borders and indentiation to better separate each variable type
* Unifying styles, we've got "pill" style for the combobox and patterns, but "input" style on the fields and the line filters.

Some things we haven't discussed:
* Drag and drop all variables to change the filter order (within each row, users shouldn't be able to put a line filter or field variable before a metadata variable)

CC @zizzpudding

PR #1168, opened 2025-04-03
Filters: Remove padding and combine level with fields

matyax:
This PR removes the left margin from the control rows, which was initially added to give a sense of hierarchy between these filters and the indexed labels on top.

margin

After making this change, it became evident that we can make a little bit of a better use of the available space, by combining the level filters with the field filters. With these changes, the level filter will not wrap, while the fields filter can freely grow.

combined

Related with #971


L2D2Grafana (approved):
lgtm, thanks for jumping right on this issue from slack! 
Screenshot 2025-04-03 at 7 40 32 AM
In super small I wouldn't mind stacking them. Also seems like the patterns aren't responsive in small, perhaps a something to look into later

gtk-grafana:
LGTM, but agree with Liza we should put stuff on it's own row for smaller viewports
image

matyax:
Updated with responsiveness. Also added a breakpoint to the top header filters which were collapsing in a weird way.

https://github.com/user-attachments/assets/c67fa92a-2347-49a0-abe8-32f95c008bc2

PR #1175, opened 2025-04-04
chore: (provisioning) - add loki datasource and env

L2D2Grafana:
# Description
Provision loki-dev-005 and add `.env.template` file. 
Are there internal docs I should update? 🤔


gtk-grafana:
e2e looks broken on this branch: 
image
Are we setting the cloud db as default?

gtk-grafana:
> interesting that the playwright videos are not saved in the github action. I'd be a fan of uncommenting this in the playwright config.

Yeah if you can get it working go for it, I just download the trace and run it locally....

oh the traces are gone from the drone output. wtf?

gtk-grafana (approved):

Issue #1156, opened 2025-04-01
"Copy link to log line" does not respect GF_SERVER_ROOT_URL subpaths


proever:
I set `GF_SERVER_ROOT_URL=https://<some_internal_host>/monitoring/`.

When I am at `https://<some_internal_host>/monitoring/a/grafana-lokiexplore-app/explore/service/test_server/logs?patterns=%5B%5D&from=now-15m&to=now&var-ds=loki&var-filters=service_name%7C%3D%7Ctest_server&var-fields=&var-levels=&var-metadata=&var-patterns=&var-lineFilterV2=&var-lineFilters=&timezone=browser&var-all-fields=&urlColumns=%5B%5D&visualizationType=%22logs%22&displayedFields=%5B%5D&sortOrder=%22Descending%22&wrapLogMessage=false`

I click "copy link to log line" and get `https://<some_internal_host>/a/grafana-lokiexplore-app/explore/service/test/logs?patterns=%5B%5D&from=2025-04-01T16%3A51%3A07.236Z&to=2025-04-01T16%3A51%3A07.238Z&var-ds=loki&var-filters=service_name%7C%3D%7Ctest&var-fields=&var-levels=&var-metadata=&var-patterns=&var-lineFilterV2=&var-lineFilters=&timezone=browser&var-all-fields=&urlColumns=%5B%5D&visualizationType=%22logs%22&displayedFields=%5B%5D&sortOrder=%22Descending%22&wrapLogMessage=false&panelState=%7B%22logs%22%3A%7B%22id%22%3A%22logsPanelQuery_1743526267237293568_edae8f94%22%2C%22displayedFields%22%3A%5B%5D%7D%7D`

In other words, the subpath `/monitoring/` is not being respected. 

The button above, which copies a link to the current page (of all logs), works correctly: `https://<some_internal_host>/monitoring/a/grafana-lokiexplore-app/explore/service/test/logs?patterns=%5B%5D&from=now-15m&to=now&var-ds=loki&var-filters=service_name%7C%3D%7Ctest&var-fields=&var-levels=&var-metadata=&var-patterns=&var-lineFilterV2=&var-lineFilters=&timezone=browser&var-all-fields=&urlColumns=%5B%5D&visualizationType=%22logs%22&displayedFields=%5B%5D&sortOrder=%22Descending%22&wrapLogMessage=false`

gtk-grafana:
[...]

PR #1176, opened 2025-04-05
SubPath: Fix url sharing and line filter migrations

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1156
Fixes: https://github.com/grafana/logs-drilldown/issues/1177

Also the line-filter migration from the old url structure was using "location" instead of the location service which was causing a redirect that contained the subpath twice.

Any use of "location" not from the location service is at risk of url bugs like this with subpath. Ideally we could add this as a lint rule so this type of bug will not get committed in the future.

Additionally I updated the e2e grafana instance to exist in a subpath so e2e tests will hopefully catch future bugs with urls missing subpath (which led to catching the line filter migration bug).


.eslintrc
```diff
@@ -1,3 +1,16 @@
...
+  },
+  "overrides": [
+    {
+      "files": ["src/**/*.test.{ts,tsx}"],
+      // Allow location to be used in test files
+      "rules": {
+        "no-restricted-globals": 0
+      }
+    }
+  ]
```
L2D2Grafana, line 15: Very cool!
Screenshot 2025-04-07 at 2 28 19 PM

L2D2Grafana (commented):
Any tips on how to verify this? When I click around are there sub-paths in the app I can trigger somewhere 🤔

.eslintrc
gtk-grafana: > Any tips on how to verify this? When I click around are there sub-paths in the app I can trigger somewhere 🤔

Yeah after deleting your grafana image and running `yarn server` the grafana instance at port 3001 should be in a subpath /grafana

gtk-grafana: To test in another instance you'd need to set GF_SERVER_ROOT_URL and GF_SERVER_SERVE_FROM_SUB_PATH config values: https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#root_url

L2D2Grafana (approved):
lgtm, finally got my local to navigate around with a sub path and no issues!
Screenshot 2025-04-08 at 10 26 57 AM

matyax (approved):
Changes look good and works as expected in my tests 👍

.eslintrc
matyax: Great idea adding this 👌

Issue #1126, opened 2025-03-18
Logs/Table: sync displayed fields


matyax:
It would be ideal that we respect the displayed fields when switching between display modes.

gtk-grafana:
What's the difference between this and https://github.com/grafana/logs-drilldown/issues/857? Should we close one of these as duplicate?

CC: @matyax

gtk-grafana:
@matyax I think so, and https://github.com/grafana/logs-drilldown/issues/1195 should be moved to grafana/grafana?

matyax:
Moved. Closing as completed.

PR #1189, opened 2025-04-10
feat: (LogPanelTable) - sync the display fields and urlColumns between the logs panel and table

L2D2Grafana:
- #1126 
- Sync the displayFields with the urlColumns
- e2e test to verify url

https://github.com/user-attachments/assets/cfde05f8-c1a6-4b22-a10e-7933b305cc1c

### DIFF
--- a/src/Components/ServiceScene/LogsTableScene.tsx
+++ b/src/Components/ServiceScene/LogsTableScene.tsx
@@ -12,8 +12,9 @@ import { getLogsPanelFrame } from './ServiceScene';
 import { getVariableForLabel } from '../../services/fields';
 import { PanelMenu } from '../Panels/PanelMenu';
 import { LogLineState } from '../Table/Context/TableColumnsContext';
+import { DEFAULT_URL_COLUMNS } from '../Table/constants';
 
-let defaultUrlColumns = ['timestamp', 'body'];
+let defaultUrlColumns = DEFAULT_URL_COLUMNS;
 
 interface LogsTableSceneState extends SceneObjectState {
   menu?: PanelMenu;
@@ -46,7 +47,6 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
   // on activate sync displayed fields with url columns
   onActivateSyncDisplayedFieldsWithUrlColumns = () => {
     const parentModel = this.getParentScene();
-
     parentModel.setState({
       urlColumns: Array.from(new Set([...defaultUrlColumns, ...parentModel.state.displayedFields])),
     });
--- a/src/Components/ServiceScene/LogsTableScene.tsx
+++ b/src/Components/ServiceScene/LogsTableScene.tsx
@@ -34,7 +34,29 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
     this.setState({
       menu: new PanelMenu({ addInvestigationsLink: false }),
     });
+    this.onActivateSyncDisplayedFieldsWithUrlColumns();
   }
+
+  private getParentScene() {
+    return sceneGraph.getAncestor(this, LogsListScene);
+  }
+
+  // on activate sync displayed fields with url columns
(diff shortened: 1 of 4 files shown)


gtk-grafana:
Overall looking good and working as expected, I really like the idea of being able to switch more easily between the views without needing to reconfigure columns. 

A few bugs/regressions with the log line and time columns to clean up and we should be good to go:

If I have nothing visualized, and I toggle the table and back to the logs panel, the log line is stripped out and I see the "show original log line" button. I would expect this to be a noop.

https://github.com/user-attachments/assets/86637aa7-1ac6-497d-8c22-eb1008b55192

The timestamp currently is always displayed in the logs panel, so it's a bit jarring to have it removed from the Table:
image
image

When taking a logs panel query to Explore, the Time column is always stripped out, which doesn't occur on latest main:
image

gtk-grafana:
@L2D2Grafana that is much closer to what I expect, except at the [0:29-0:33] timestamp in your video, I would expect the panel in Explore to have the same columns as the panel in Logs Drilldown

L2D2Grafana:
> @L2D2Grafana that is much closer to what I expect, except at the [0:29-0:33] timestamp in your video, I would expect the panel in Explore to have the same columns as the panel in Logs Drilldown

I opened a can of worms 🪱 updating the url state. I still have one more known bug to fix opening in explore.

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -34,7 +39,68 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
...
+
+  private getParentScene() {
+    return sceneGraph.getAncestor(this, LogsListScene);
   }
+
+  // on activate sync displayed fields with url columns
+  onActivateSyncDisplayedFieldsWithUrlColumns = () => {
+    const searchParams = new URLSearchParams(locationService.getLocation().search);
+    const urlColumnsParam = searchParams.get('urlColumns');
+    const urlColumnsUrl = urlColumnsParam ? (JSON.parse(decodeURIComponent(urlColumnsParam)) as string[]) : null;
```
matyax, line 53: This seems like a good candidate to use https://github.com/grafana/logs-drilldown/blob/main/src/services/narrowing.ts that was introduced by @gtk-grafana to have more confidence on the actual type of stuff like user input, parsed json, etc.
We could add a narrowArray(type: string) or narrowStringArray().

matyax:
This is super cool! It makes a huge difference going back and forth between visualizations.

src/Components/ServiceScene/LogsTableScene.tsx
L2D2Grafana: Updated! Interesting concept, I took a look they all kind of seem like type validations. I wonder about naming them isValidType or isEmpty.

src/services/narrowing.ts
```diff
@@ -123,4 +123,19 @@ export function narrowFilterOperator(op: string): LabelFilterOp | NumericFilterO
   }
 }
 
+export function narrowUrlParam(param: unknown): string[] | null {
```
matyax, line 126: Excellent! My only suggestion would be to make it generic so we can reuse this narrowing elsewhere, maybe like:

```suggestion
export function narrowStringsArray(param: unknown): string[] | null {
```

I wanted to make it generic but I couldn't find an easy way.

gtk-grafana: Yeah if there was a generic narrowing strategy it would be the most imported typescript npm package, I think this is something that needs a bit of hand-holding

gtk-grafana: there is a `unknownToStrings` method in narrowing already

L2D2Grafana: I tried `unknownToStrings` but it's a slightly different use case where the param are passed in as strings and not an array so it never gets past 
```
if (Array.isArray(a)) {
```

L2D2Grafana: Updated the func name.

Issue #989, opened 2025-01-09
Line filters: Validation


gtk-grafana:
Can we validate the users input before sending the query?

See: https://github.com/grafana/explore-logs/pull/963#issuecomment-2579712305

Frontend validation:
Regex validation: We can test regex locally, but will that line up with loki response?
Non regex: Can we check for invalid chars that will break queries? Or should we be escaping them?

Backend validation:
Can we run an instant query or something fast to check query against loki?

Needs some investigation.

le0pard:
Re2js is heavy (due to Unicode maps), but it is not 3.22MB. You counted all variations (commonjs, ecmascript, umd)  + source map. I think you will use only one of variation

![Image](https://github.com/user-attachments/assets/fd0843b8-f1ed-4086-86c8-37346535b1c4)

gtk-grafana:
@le0pard my apologies and good catch!
Yeah 300Kb is **much** more feasible, I'll try to find some time to look into lazy loading a separate bundle with that package since it looks like many users aren't touching the regex options.

le0pard:
Yep, I definitely will do the same, if this is not major functionality for an app. `await import` should support some builders and do automatic code splitting

gtk-grafana:
I'll take a quick swing here at a PoC with frontend re2 validation.

PR #1190, opened 2025-04-14
Line filters: Client regex validation

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/989

Asynchronously loads re2JS package when regex is toggled on line filter inputs in an effort to keep load times down for users that will not add regex line filters.

Displays error message from re2JS in tooltip under the input.
image
image

Follow ups:
We could potentially prevent the query from executing when an invalid regex is detected, but this would add a bunch of complexity to this PR, and I wasn't certain that the version of re2 used in Loki would exactly map to the frontend library, and didn't want to risk a regression preventing a user from submitting a query that might work on the backend, but throws an error in the client.

Since Loki will validate the query before executing, preventing queries with invalid regex shouldn't impact billing or have significant performance impact, so the benefits of preventing queries with invalid regex line filters are mostly limited to keeping the frontend clean:

image

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/LineFilterInput.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LineFilterInput.tsx
@@ -1,6 +1,6 @@
 import { css } from '@emotion/css';
 import { Icon, IconButton, Input, Tooltip, useStyles2 } from '@grafana/ui';
-import React, { ChangeEvent, HTMLProps, useState } from 'react';
+import React, { ChangeEvent, HTMLProps, useCallback, useEffect, useState } from 'react';
 import { GrafanaTheme2 } from '@grafana/data';
 import { narrowErrorMessage } from '../../../services/narrowing';
 
@@ -20,41 +20,55 @@ export const LineFilterInput = ({ value, onChange, placeholder, onClear, suffix,
   const [invalid, setInvalid] = useState(false);
   const [errorMessage, setErrorMessage] = useState('');
 
-  const load = async () => {
+  const validate = useCallback(
+    (value: string) => {
+      if (value) {
+        try {
+          re2JS?.compile(value);
+          if (invalid) {
+            setInvalid(false);
+            setErrorMessage('');
+          }
+        } catch (e) {
+          const msg = narrowErrorMessage(e);
+          if (!invalid) {
+            setInvalid(true);
+          }
+
+          if (msg && msg !== errorMessage) {
+            setErrorMessage(msg);
+          }
+        }
+      } else if (invalid) {
+        setInvalid(false);
+        setErrorMessage('');
+      }
+    },
+    [errorMessage, invalid]
+  );
+
+  const load = useCallback(async () => {
     re2JS = null;
     re2JS = (await import('re2js')).RE2JS;
-  };
+  }, []);
 
-  const initValidation = (value: string) => {
-    if (re2JS === undefined && regex) {
-      load().then(() => validate(value));
-    } else if (regex && re2JS) {
-      validate(value);
-    }
-  };
-
-  const validate = (value: string) => {
-    if (value) {
-      try {
-        re2JS?.compile(value);
-        if (invalid) {
-          setInvalid(false);
-          setErrorMessage('');
-        }
-      } catch (e) {
-        const msg = narrowErrorMessage(e);
-        if (!invalid) {
-          setInvalid(true);
-        }
-
-        if (msg && msg !== errorMessage) {
-          setErrorMessage(msg);
-        }
+  const initValidation = useCallback(
+    (value: string) => {
+      if (re2JS === undefined && regex) {
+        load().then(() => validate(value));
+      } else if (regex && re2JS) {
+        validate(value);
       }
-    } else if (invalid) {
+    },
+    [load, regex, validate]
+  );
+
+  useEffect(() => {
+    if (!regex) {
       setInvalid(false);
+      setErrorMessage('');
     }
-  };
+  }, [regex]);
 
   initValidation(value);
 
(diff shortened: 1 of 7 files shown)


src/Components/ServiceScene/Breakdowns/LineFilterInput.tsx
```diff
@@ -1,41 +1,96 @@
...
-          ) : undefined}
-          {suffix && suffix}
-        </span>
+  const [invalid, setInvalid] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const load = async () => {
+    re2JS = null;
+    re2JS = (await import('re2js')).RE2JS;
+  };
```
L2D2Grafana, line 26: This could use a useCallback() or loading check to only run once

src/Components/ServiceScene/Breakdowns/LineFilterInput.tsx
```diff
@@ -1,41 +1,96 @@
...
+
+  const initValidation = (value: string) => {
+    if (re2JS === undefined && regex) {
+      load().then(() => validate(value));
+    } else if (regex && re2JS) {
+      validate(value);
+    }
+  };
```
L2D2Grafana, line 34: initValidation  is another candidate for useCallback  or perhaps useEffect to run on value change

gtk-grafana:
@L2D2Grafana good catch with the fields, looks like newer versions of Loki return a 400 when sent an invalid regex, but older ones, like [in play](https://play.grafana.org/api/datasources/uid/ddhr3fttaw8aod/resources/detected_fields?query=%7Bservice_name%3D%22tempo-distributor%22%7D%20%20%20%20%7C~%20%22(%22%20%7C%20json%20%7C%20logfmt%20%7C%20drop%20__error__%2C%20__error_details__%20&start=2025-04-15T12%3A03%3A30.795Z&end=2025-04-15T12%3A18%3A30.795Z) just return an empty object.

I'll open an issue and fix the loading state in another PR since that is present on main.

gtk-grafana:
RE: toast message: Grafana core seems to throw those whenever a 4/5xx is received from the datasource. I discussed preventing running invalid queries in this PR summary as a follow-up, I argue for now we should continue running the queries until we have a bit more confidence that the frontend validation will never say something is invalid that would currently run in Loki.

L2D2Grafana (approved):
Thanks for the updates!

src/Components/ServiceScene/Breakdowns/LineFilterInput.tsx
```diff
@@ -1,41 +1,110 @@
...
+  );
+
+  useEffect(() => {
+    if (!regex) {
+      setInvalid(false);
+      setErrorMessage('');
+    }
+  }, [regex]);
+
+  initValidation(value);
```
matyax, line 73: Do you need this to run on every re-render?

gtk-grafana: nah we can wrap in a useEffect and run on init, good catch

matyax: Yeah, better, becuse it might be running on every keystroke.

gtk-grafana: Should be cleaned in https://github.com/grafana/logs-drilldown/pull/1190/commits/797a27fe9814a797034c2068959f31461edb586f, it'll still run once on every keystroke, but not twice!

matyax:
Not related with this PR, but it seems that the types don't reflect what we're getting, since it can be undefined.

imagen

src/Components/ServiceScene/Breakdowns/LineFilterInput.tsx
```diff
@@ -1,41 +1,111 @@
...
+      setInvalid(false);
+      setErrorMessage('');
+      return;
+    }
+
+    if (!regex) {
+      setInvalid(false);
+      setErrorMessage('');
+    }
+  }, [regex, value, initValidation]);
```
matyax, line 79: Because the dependencies, this still gets executed on every keystroke.

https://github.com/user-attachments/assets/5cbb42f6-505c-4d0a-972c-e530ebb94198

If I may suggest, something like:

```
const load = async () => {
  re2JS = null;
  re2JS = (await import('re2js')).RE2JS;
};

export const LineFilterInput = ({ value, onChange, placeholder, onClear, suffix, width, regex, ...rest }: Props) => {
  const styles = useStyles2(getStyles);
  const [invalid, setInvalid] = useState(false);
  const [errorMessage, setErrorMessage] = useState('');

  const validate = useCallback(
    (value: string) => {
      if (!value || !regex) {
        setErrorMessage('');
        setInvalid(false);
        return;
      }
      if (re2JS === undefined) {
        load().then(() => validate(value));
        return;
      }
      try {
        re2JS?.compile(value);
        setInvalid(false);
        setErrorMessage('');
      } catch (e) {
        const msg = narrowErrorMessage(e);
        setInvalid(true);
        if (msg) {
          setErrorMessage(msg);
        }
      }
    },
    [regex]
  );

  useEffect(() => {
    validate(value);
  }, [validate, value])

  return (...);
};
```
Would simplify things a bit. Also it seems that you can move load outside of the component.

gtk-grafana: Good call, implemented in https://github.com/grafana/logs-drilldown/pull/1190/commits/e27fd154d1bbbcfc1bbe4f8e1142364dfbe0e1cb.

Thanks for taking another look at this!

matyax (approved):
Love this!

gtk-grafana:
> Not related with this PR, but it seems that the types don't reflect what we're getting, since it can be undefined.

Yeah I noticed that last night, I tweaked it so this component doesn't throw the switching from uncontrolled to controlled error: https://github.com/grafana/logs-drilldown/pull/1190/commits/797a27fe9814a797034c2068959f31461edb586f#diff-324e04327e8dcb29229c883c12c4cc854844346a9ffccb74e547178c7d332a28R76, but we should probably track that down and make sure it defaults to empty string

PR #1194, opened 2025-04-16
Service Selection: Conditionally display "show logs" buttons

matyax:
Potential follow-up to https://github.com/grafana/logs-drilldown/pull/1188, where we keep the option of going directly to the details of a single value.

Additionally, "Exclude" has been renamed as "Remove", to better match the clicking outcome.

https://github.com/user-attachments/assets/95881f70-c75d-4ab6-b7d2-bd81049bd39e


gtk-grafana (commented):
Code looks good, I'll just want to test the aggregated metrics.
Looks like some e2e tests broke with the new button copy change as well

gtk-grafana (approved):
Looks good, we might want to differentiate the button styles and make the show logs stand out a bit more, but we can handle that in another PR

matyax:
Good call. Let's go with primary, which matches the "show logs" in the sticky bar. We can always revisit otherwise, but it definitely makes it a bit clearer.

PR #1204, opened 2025-04-21
Logs Panel: Use new Log Controls component

matyax:
This PR enables the controls exposed through https://github.com/grafana/grafana/pull/104237 for the Logs visualization.

For the Table and Json visualizations, a custom controls component with limited options have been created.

Requires https://github.com/grafana/grafana/pull/104237

Demo:

https://github.com/user-attachments/assets/2c4da97c-dda1-4d87-aca4-709c02373c4e


matyax:
Opening for early feedback from @grafana/observability-logs

src/services/panel.ts
```diff
@@ -333,3 +336,5 @@ export function getQueryRunnerFromProvider(provider: SceneDataProvider): SceneQu
 
   throw new Error('SceneDataProvider is missing SceneQueryRunner');
 }
+
+export const logsControlsSupported = config.buildInfo.version > '12' || config.buildInfo.version.includes('12.1');
```
gtk-grafana, line 340: I was just about to ask if we need to wait until 12.1 is released, but it looks like we'll support both with/without for a bit. Should we create an issue when <12.1 is deprecated to remove this and clean up the old code paths?

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -310,8 +314,38 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
...
+      .setOption('sortOrder', this.state.sortOrder)
+      .setOption('wrapLogMessage', this.state.wrapLogMessage)
+      .setOption('prettifyLogMessage', this.state.prettifyLogMessage);
+
+    if (!logsControlsSupported) {
+      panel.setOption('showTime', true);
+    } else {
+      panel
+        .setOption('showTime', getBooleanLogOption('showTime', true))
+        // @ts-expect-error Requires Grafana 12.1
```
gtk-grafana, line 326: lets create an issue to remove these when our min grafana version is >=12.1

src/services/panel.ts
gtk-grafana: And should this be moved to guards, or somewhere else outside of the panel.ts?

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -310,8 +314,38 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
...
+        .setOption('showControls', true)
+        // @ts-expect-error Requires Grafana 12.1
+        .setOption('controlsStorageKey', LOG_OPTIONS_LOCALSTORAGE_KEY)
```
gtk-grafana, line 329: oh this is neat, but do we want to use the same key as used in Explore so folks get the same options set in both places?

gtk-grafana (commented):
It works well locally, can verify that we see the old options and they work when users are running Grafana < 12.1 that doesn't have the new options in https://github.com/grafana/grafana/pull/104237.

We're taking on a bit of debt with this, I don't think there's much we can do about that for now, but we should document somewhere to why we're not exporting the logs options from core in the table/json panel (logs model vs dataframe).

gtk-grafana:
Oh, don't forget to look into adding the log text/labels toggle to the table, and adding to the JSON viz as well (which will probably just have sort order, and maybe scroll to top/bottom (although I'm not sure if that will require an update in the base package)).

src/Components/ServiceScene/LogsPanelScene.tsx
matyax: Yeah, let's do that. I can't find a good argument to keep them separated and it should help with going back and forth between modules.

src/Components/ServiceScene/LogsPanelScene.tsx
matyax: Done.

src/services/panel.ts
matyax: Issue created. About placement, I couldn't find a great fit for this. The rationale here was that this is related with the underlying panel version :/

src/services/store.ts
```diff
@@ -189,11 +189,11 @@ function getExplorationPrefix(sceneRef: SceneObject) {
   return `${ds}.${serviceName}`;
 }
 
-export function getDisplayedFields(sceneRef: SceneObject) {
+export function getDisplayedFields(sceneRef: SceneObject): string[] {
   const PREFIX = getExplorationPrefix(sceneRef);
   const storedFields = localStorage.getItem(`${pluginJson.id}.${PREFIX}.logs.fields`);
   if (storedFields) {
-    return JSON.parse(storedFields);
+    return narrowStringsArray(JSON.parse(storedFields)) ?? [];
```
matyax, line 196: @L2D2Grafana Reused the narrowing here 💪

matyax:
> Oh, don't forget to look into adding the log text/labels toggle to the table

Done.

>  and adding to the JSON viz as well 

Done.

src/services/narrowing.ts
```diff
@@ -131,19 +131,12 @@ export function narrowFilterOperator(op: string): LabelFilterOp | NumericFilterO
...
-    return null;
-  } catch (e) {
-    // If parsing fails, it's not a valid JSON string
-    return null;
+export function narrowStringsArray(array: unknown): string[] | null {
+  // Verify the parsed result is an array of strings
+  if (Array.isArray(array) && array.every((item) => typeof item === 'string')) {
+    return array;
   }
+  return null;
```
L2D2Grafana, line 139: Ah yes this is now like `unknownToStrings`. I will have to create some kind of function to parse url params which are just strings (not arrays) which is why I couldn't use `unknownToStrings`  or I can put the parsing logic back in the component logic `const parsedParam = JSON.parse(decodeURIComponent(param as string));`

gtk-grafana:
This is odd: 
image

But otherwise things look good. 
There is a small change in functionality, which is that we're splitting up line wrap lines and expand/collapse json in logs drilldown to match Explore, not sure if that's intended or not, but not a blocker.

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -665,6 +760,9 @@ const getStyles = (theme: GrafanaTheme2) => ({
     --json-tree-arrow-color: ${theme.colors.secondary.contrastText};
     --json-tree-ul-root-padding: 0 0 ${theme.spacing(2)} 0;
 
+    overflow: auto;
+    height: 100%;
+
```
gtk-grafana, line 765:
```suggestion
width: 100%;
```

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -682,10 +780,14 @@ const getStyles = (theme: GrafanaTheme2) => ({
       z-index: 2;
       padding-left: ${theme.spacing(1)};
       align-items: center;
-      overflow-x: scroll;
+      overflow-x: auto;
       overflow-y: hidden;
     }
 
+    > ul > li > span > span > span {
```
gtk-grafana, line 787: I think this is the same selector as `drillUpWrapperStyle` in `/explore-logs/src/services/JSONViz.ts`

src/services/panel.ts
gtk-grafana: Works for me.

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -360,104 +438,114 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
-
-                // Don't render time node
-                if (keyPath[0] === DataFrameTimeName) {
-                  return null;
-                }
-
-                return <strong>{keyPath[0]}:</strong>;
-              }}
+        <div className={styles.container}>
+          {logsControlsSupported && (
```
gtk-grafana, line 442: do we also want to check the `logsPanelControls` ff?

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -154,29 +233,44 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
...
-            />
-          )}
+          <div className={styles.container}>
+            {logsControlsSupported && (
+              <LogListControls
+                sortOrder={sortOrder}
+                onSortOrderChange={model.handleSortChange}
+                onManageColumnsClick={model.onManageColumnsClick}
+                onLineStateClick={model.onLineStateClick}
+                // "Auto" defaults to display "show text"
```
gtk-grafana, line 252: from what I remember auto will show text if the cardinality of the labels is 1, otherwise will show labels

src/services/narrowing.ts
matyax: Ah, I will refactor if there's redundancy there.

> which is why I couldn't use unknownToStrings or I can put the parsing logic back in the component logic const parsedParam = JSON.parse(decodeURIComponent(param as string));

Yeah, I already made that change. The retrieval (searchParams.get(), localStorage.getItem()) or pasing (JSON.parse(), etc) is not a concern for narrowing. The consumer part (let's say a Scene) gets the `unknown` value from somewhere, and the narrowing part is in charge of determining that this unknown is of the expected type.

src/Components/ServiceScene/LogsJsonScene.tsx
matyax: It certainly is.

src/Components/ServiceScene/LogsJsonScene.tsx
matyax: Ohhh, good call!

matyax: Checking for the featue flag now.

src/Components/ServiceScene/LogsTableScene.tsx
matyax: What I did was forcing both states (with labels and with no labels) in the component and check that both the Table and the Controls show the same icon.

https://github.com/grafana/logs-drilldown/blob/main/src/Components/Table/LogLineCellComponent.tsx#L136

Hopefully this was a correct interpreation of it.

matyax:
> There is a small change in functionality, which is that we're splitting up line wrap lines and expand/collapse json in logs drilldown to match Explore, not sure if that's intended or not, but not a blocker.

The exposed component ships with both options, and maintaining the same behavior would require a snowflake just for Drilldown that I don't think is worth the effort. In any case, this will change and deprecate with the new panel.

matyax:
Small follow up here: https://github.com/grafana/grafana/pull/104477

gtk-grafana (approved):
Looks good, works well locally. 
Can confirm feature flag works as expected, and the new menu is not supported for Grafana <12.1.
Great work on this as always!

PR #1215, opened 2025-04-23
chore: (plugin-ci-workflows) - init vanilla plugin-ci-workflows

L2D2Grafana:
This PR is officially ready for merging 🥳 

# pr-ci.yml
- Here's a successful [job](https://github.com/grafana/logs-drilldown/actions/runs/14886934571?pr=1215) running in 15min. 
- On playwright test failures playwright reports are showing up [example](https://github.com/grafana/logs-drilldown/actions/runs/14886416107)

# publish.yml
- Mostly the same as pr-ci.yml
- We still need to create a branch, npm run version, generate changelog
- Trigger is pushing a git tag
- Right now the plugin will be uploaded to GCS for backwards compatibility. Eventually it will just be uploaded to the plugin catalogue and deployment_tools should be updated to provision from the catalog rather than GCS.

# Future
- Merge this PR
- Remove old release.yml, ci.yml, drone 
- Delete all old jobs from those pipelines to full delete the files
- Create a gha for creating a branch, npm run version, generate changelog

PR #1217, opened 2025-04-24
Chore: CI: Test Playwright

xnyo:


L2D2Grafana:
🙇 TY! Closing and continuing on with [1215](https://github.com/grafana/logs-drilldown/pull/1215)
TL:DR 
- Namespacing of gh actions matters: having 2 prs with the exact same job names I believe is what cause the hour long ci builds initially.
- If you want to delete a gh action you have to delete all the workflow runs before it magically goes away on it's own 🪄, don't ask how I figured that out 🤐
- Needed to run the exact version grafana/grafana-enterprise:11.6.1. Locally we run latest which was `grafana/grafana-enterprise:11.6.1.` but in the matrix ci that ended up being `grafana/grafana-dev:12.0.0-239779` a bit of a red herring for me 🦩

PR #1221, opened 2025-04-28
chore(zizmor): add gh action for zizmor

L2D2Grafana:
Add zizmore gh-action


L2D2Grafana:
Trufflehog I think is okay. I've been running it since last week and it says verified secrets 0 and unverified secrets 0
Screenshot 2025-04-28 at 9 01 16 AM

Issue #1147, opened 2025-03-27
Docker: Update to 3.5.0 when released


matyax:
Needs to be updated to 3.5.0 when released

_Originally posted by @gtk-grafana in https://github.com/grafana/logs-drilldown/pull/1122#discussion_r2012964086_

PR #1235, opened 2025-05-01
chore(docker-compose): update loki to 3.5.0 

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1147


dist/plugin.json
```diff
@@ -0,0 +1,99 @@
+{
```
matyax, line 1: Extra change here.

gtk-grafana: yeah weird, this should be ignored

matyax (approved):

PR #1238, opened 2025-05-01
fix(line-filters): Expand no longer working.

gtk-grafana:
Looks like with latest Grafana onFocus of the `<Input />` component is no longer firing on focus. 
`onFocusCapture` seems to work


L2D2Grafana (approved):
👍

PR #1240, opened 2025-05-01
chore: add option to output to syslog from the generator

trevorwhitney:
Increasing the surface area of Loki use cases we can test, this PR add a `-syslog=true` option to the generator, allowing it to send logs to specified syslog sink. it also adds some scaffolding for easily testing this use case, which will be particularly useful for me since I've been asked to demo how to get syslog logs into Loki, and using our generator will allow me to show how we can achieve feature parity with other ingestion methods.


generator/flog/random_test.go
```diff
@@ -1,12 +0,0 @@
-package flog
```
trevorwhitney, line 1: this file had no tests, and was producing errors when I tried to run the tests locally, so I'm not sure what the goal of it is or how I'm supposed to run it?

project-words.txt
```diff
@@ -518,3 +518,9 @@ grafanacon
 upsert
 Dedup
 dedup
+PROCID
+MSGID
+SSSZ
+NILVALUE
+PROCID
+MSGID
```
L2D2Grafana, line 526: Why do we add these?

trevorwhitney: The comment I added to document the RFC5424 spec tripped the spelling czar

```
// Format the RFC5424 message - strictly adhering to RFC5424 format
// <PRI>VERSION TIMESTAMP HOSTNAME APP-NAME PROCID MSGID STRUCTURED-DATA MSG
```

docker-compose-syslog.dev.yaml
```diff
@@ -0,0 +1,54 @@
+version: '3.0'
+# Assumes you are running loki and the generator locally
+services:
```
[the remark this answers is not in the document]
trevorwhitney: so, I think we 1) move the generator out of this repo, and then 2) create a mega Dockerimage with the generator, alloy, and otel-collector, and refactor the generator to allow as many outputs as you want. Then we hopefully only need one compose that takes some runtime args?

PR #1247, opened 2025-05-06
chore(e2e): clean up & flake prevention

gtk-grafana:
Quarterly e2e flake cleaning

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -40,7 +40,7 @@ export default defineConfig<PluginOptions>({
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: 'html',
   /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
+  retries: process.env.CI ? 1 : 0,
   testDir: './tests',
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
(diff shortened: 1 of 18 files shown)


playwright.config.ts
```diff
@@ -40,7 +40,7 @@ export default defineConfig<PluginOptions>({
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: 'html',
   /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
+  retries: process.env.CI ? 1 : 0,
```
gtk-grafana, line 43: Increasing the amount of retries doesn't seem to help reduce flakey CI executions, but it does make the suite take a good bit longer to fail

L2D2Grafana:
This does seem like an improvement!
Locally I see three tests are flaky and only traces for the flakes

in CI I see a different flaky test 
`mock volume API calls › changing datasource will trigger new queries `
Screenshot 2025-05-06 at 3 45 12 PM

gtk-grafana:
@L2D2Grafana I'm not able to reproduce those local fails besides the JSON one, which I pushed a "fix" for.
Where are you seeing the failure in CI?

gtk-grafana:
> Dang you got the build time down to 10min!
It was failing after 10m, a bunch of tests hadn't run when it gave up.
There are two "bytes" tests, I had only updated one and the other was flaking, hopefully the latest commit cleans that up.

L2D2Grafana (approved):
ty 👍

Issue #1165, opened 2025-04-02
[Chore]: Add CP bundle stats action and remove bundlewatch


L2D2Grafana:
# Description
Once all the kinks are worked out with the create plugin bundle-stats action add that and remove bundlewatch.
* Note that bundle watch only compares the diff from last time.
[demo](https://github.com/jackw/heywesty-bundlestats-panel/pull/1)

Leverage the create plugin bundle-stats action and remove the bundlewatch npm package and possibly other attempts like 
```
// const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```

PR #1252, opened 2025-05-07
chore(bundle-stats): swap bundlewatch for cp gha bundle-stats

L2D2Grafana:
- #1165 
Successful run, it should show up on the PR once it's in main
Screenshot 2025-05-07 at 9 57 29 AM

### DIFF
--- a/.github/workflows/bundle-stats.yml
+++ b/.github/workflows/bundle-stats.yml
@@ -1,0 +1,25 @@
+name: Bundle Stats
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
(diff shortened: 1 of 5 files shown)


gtk-grafana (approved):
Looks good, is there no way to configure limits like with bundlewatch?

L2D2Grafana:
> Looks good, is there no way to configure limits like with bundlewatch?

There is a threshold: The minimum entrypoints total percentage difference required to post comments in PRs (defaults to 5) and a few other options https://github.com/grafana/plugin-actions/tree/main/bundle-size

PR #1253, opened 2025-05-07
fix(JSON): logs and table panel adding JSON parsers making fields disappear 

gtk-grafana:
Some fallout from https://github.com/grafana/logs-drilldown/pull/1122

The bug:
When adding field filters in the logs or table panel, this adds the field as a prop to the logQL JSON parser, which has the consequence of removing any non-specified fields.

In the JSON panel this doesn't impact much, as the fields are parsed directly from the resultant JSON, but logs and table panel. But in the logs/table panel this has the consequence of making fields (other then those in active filters) disappear when adding filters.

To reproduce:
In the table panel, [visualize two fields as columns](https://play.grafana.org/a/grafana-lokiexplore-app/explore/service/grafanacon-json-otel/logs?patterns=%5B%5D&from=now-15m&to=now&var-lineFormat=&var-ds=ddhr3fttaw8aod&var-filters=service_name%7C%3D%7Cgrafanacon-json-otel&var-fields=&var-levels=&var-metadata=&var-jsonFields=&var-patterns=&var-lineFilterV2=&var-lineFilters=&timezone=utc&var-all-fields=&urlColumns=%5B%22Time%22,%22Line%22,%22nested_object_deeplyNestedObject_extraDeeplyNestedObject_url%22,%22nested_object_deeplyNestedObject_url%22%5D&visualizationType=%22table%22&displayedFields=%5B%22nested_object_deeplyNestedObject_extraDeeplyNestedObject_url%22,%22nested_object_deeplyNestedObject_url%22%5D&sortOrder=%22Descending%22)

Then click on a cell in either column and include or exclude. You'll notice that the other column is removed. 

This fix:
This PR simply removes the assumption that every JSON field added anywhere in the application needs a JSON parser, and instead adds them when the user switches to the JSON viz. However since filters added in the JSON viz may not be valid queries without the json parser props (e.g. parent node filters), we cannot remove JSON prop filters when switching back to other parts of the application, so the behavior in this bug will continue to exhibit if users switch to JSON and then back to another viz or the fields tab.

The long-term fix:
I think we'll need to either change the behavior of the json parser in Loki, or add a wildcard parser argument so we return the default set of fields in addition to whatever has been specified in the json parser props. This way the logs and table panels will continue to see a consistent set of fields in addition to whatever fields have been "created" via the JSON panel.

### DIFF
--- a/src/Components/ServiceScene/LogsJsonScene.tsx
+++ b/src/Components/ServiceScene/LogsJsonScene.tsx
@@ -29,15 +29,20 @@ import {
 import { Alert, Badge, Button, Icon, PanelChrome, useStyles2 } from '@grafana/ui';
 
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from '../../services/analytics';
+import { areArraysEqual } from '../../services/comparison';
 import {
   clearJsonParserFields,
+  extractParserFromString,
   getDetectedFieldsJsonPathField,
   getDetectedFieldsParserField,
+  getJsonPathArraySyntax,
+  getParserAndPathForField,
   isLogLineField,
 } from '../../services/fields';
 import {
   addJsonParserFieldValue,
   EMPTY_AD_HOC_FILTER_VALUE,
+  filterUnusedJSONFilters,
   getJsonKey,
   removeLineFormatFilters,
 } from '../../services/filters';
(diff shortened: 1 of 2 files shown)


L2D2Grafana:
- On the table tab I add two json fields and included one field in the filter both columns stay
Screenshot 2025-05-07 at 12 18 54 PM

- JSON viz if I add a field only one field is retained
Screenshot 2025-05-07 at 12 22 59 PM
Screenshot 2025-05-07 at 12 22 47 PM  

imho this seems like a nice quick fix.

L2D2Grafana (approved):
💨

gtk-grafana:
> For a long term fix (i.e. before the Logs Drilldown JSON visualization can be released GA) We'll need to either change the behavior of the json parser in Loki, or add a wildcard parser argument so we return the default set of fields in addition to whatever has been specified in the json parser props. This way the logs and table panels will continue to see a consistent set of fields in addition to whatever fields have been "created" via the JSON panel.

So it turns out this is incorrect, we can simply add a [second json parser stage](https://play.grafana.org/explore?schemaVersion=1&panes=%7B%22l5k%22:%7B%22datasource%22:%22ddhr3fttaw8aod%22,%22queries%22:%5B%7B%22refId%22:%22logs%22,%22expr%22:%22%7Bservice_name%3D%5C%22grafanacon-json-otel%5C%22%7D%20%7C%20json%20user_identifier%3D%5C%22%5B%5C%5C%5C%22user-identifier%5C%5C%5C%22%5D%5C%22%20%7C%20json%20%7C%20logfmt%20%7C%20user_identifier%3D%60-%60%20%7C%20drop%20__error__,%20__error_details__%20%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22ddhr3fttaw8aod%22%7D,%22editorMode%22:%22code%22,%22queryType%22:%22range%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%22now-15m%22,%22to%22:%22now%22%7D,%22panelsState%22:%7B%22logs%22:%7B%22visualisationType%22:%22logs%22%7D%7D%7D%7D&orgId=1), so we don't need to wait for any Loki side changes!

PR #1263, opened 2025-05-08
chore(plugin-ci-workflows): publish skip playwright

L2D2Grafana:
In case we need this.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,9 +29,4 @@ jobs:
       environment: ${{ github.event.inputs.environment }}
       docs-only: ${{ fromJSON(github.event.inputs.docs-only) }}
       attestation: true
-      # Same settings as PR-CI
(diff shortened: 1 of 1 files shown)


gtk-grafana (approved):
Yeah we probably want to get rid of this, e2e will already get run on merge into main, and this flaking could slow us down getting urgent fixes out. 

I would love to run a non-blocking suite in dev/ops/prod after the release (but I think we'd want a different suite of tests to run then what we have), but that's a discussion for another issue.

L2D2Grafana:
> Yeah we probably want to get rid of this, e2e will already get run on merge into main, and this flaking could slow us down getting urgent fixes out.
> 
> I would love to run a non-blocking suite in dev/ops/prod after the release (but I think we'd want a different suite of tests to run then what we have), but that's a discussion for another issue.

I do agrees it's excessive to run the playwright tests 3 times for every deploy after they've already run on the PR!

PR #1276, opened 2025-05-15
fix(investigations): interpolate jsonFields

gtk-grafana:
Fix jsonFields variable not interpolating in investigation queries.


svennergr (approved):
Thank you!

PR #1284, opened 2025-05-19
fix(table): move sorting and remove initial sorting from table

### LINKED ISSUE #1282: "View log line" (aka Insepct) from Table View shows incorrect log data
Clicking the "View log line" in the table view shows incorrect log data for a log line that is much further down in the view. This causes a lot of confusion and frustration for users that are using this functionality to drill down in the log data.

Image

matyax:
There was a discrepancy between:
- The parsed logsFrame, which contained the original unsorted dataframe in "raw"
- The sorted dataFrame with the current LogsSortOrder method
- The "initial sort by" pointing to the time field.

This caused an incorrect "rowId" to reach the LineActionIcons component. To fix this:

- Sorting has been moved to TableProvider
- logsFrame is now generated after sorting
- Removed setting an "initialSortBy", as the default order is the one that corresponds to the selected sort order in the controls. Then the user can further sort by any of the columns if they want to.

Fixes #1282

### DIFF
--- a/src/Components/Table/Table.tsx
+++ b/src/Components/Table/Table.tsx
@@ -16,7 +16,6 @@ import {
   FieldWithIndex,
   Labels,
   MappingType,
-  sortDataFrame,
   transformDataFrame,
   ValueMap,
 } from '@grafana/data';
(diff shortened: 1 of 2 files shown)


L2D2Grafana (approved):
lgtm 👍 

Clicking on the first log row after switching the sorting in the logs view I see the right info in the overlay. I did not check this with the new logs panel controls. It would be nice to have an e2e test for this. 
Screenshot 2025-05-19 at 12 07 58 PM

PR #1294, opened 2025-05-21
feat(embedding): Embedded Service Scene Component

gtk-grafana:
Adds a new `grafana-lokiexplore-app/embedded-logs-exploration/v1` extension point which embeds the ServiceScene component for use in external applications.

When embedded, Logs Drilldown will not use routes to navigate, and instead use url parameters. By default the previously used datasource that is saved in local storage will be used, and the embedded view does not have the ability to change datasources. 
To change the datasource: manually set the DS ID in the url parameters: `&var-ds=PDDA8E780A17E7EF1`.
To set the initial labels: manually set the filter query in the url parameters: `&var-filters=service_name|=|tempo-distributor`
* The initial labels are locked, and cannot be removed in the UI

These changes are exposed on the `/embed` route of the application for testing purposes (e.g. http://localhost:3000/a/grafana-lokiexplore-app/embed)

**There should be no way to navigate to any other route besides `/embed` when testing the embedded route.**

Note: After checking out this branch and building the frontend, core Grafana will need to be restarted to add the new extension point to the registry, otherwise the `/embed` route will not work, and it will just load forever.


src/services/navigate.ts
```diff
@@ -25,6 +25,13 @@ function buildValueBreakdownUrl(label: string, newPath: ValueSlugs, labelValue:
   }
 }
 
+function buildEmbedValueBreakdownUrl(label: string, newPath: ValueSlugs, queryPrams: UrlQueryMap): string {
+  queryPrams['pageSlug'] = newPath;
```
gtk-grafana, line 29: constants?

src/services/variableHelpers.ts
```diff
@@ -85,3 +87,15 @@ export const operatorFunction = function (variable: AdHocFiltersVariable) {
 
   return operators;
 };
+
+/**
+ * For embedded contexts, return the first label as primary label
+ */
+export function getPrimaryLabelFromScene(scene: ServiceScene) {
```
gtk-grafana, line 94: This should only be called when embedded, maybe we should gate on the service scene embedded state?

gtk-grafana:
@adrapereira 

> Can you explain why in LD we need to do so much manual URL manipulation and loading? 

Logs Drilldown uses href links which set state on nav instead of setting state that changes href, this allows users to open anything in a new tab or window, and supports browser history for state and route changes.

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -566,7 +591,7 @@ function getContentScene(drillDownLabel?: string) {
   });
 }
 
-function getVariableSet(initialDatasourceUid: string, initialFilters?: AdHocVariableFilter[]) {
+function getVariableSet(initialDatasourceUid: string, initialFilters?: AdHocVariableFilter[], embedded?: boolean) {
```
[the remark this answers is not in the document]
gtk-grafana: Thanks for the reminder, should be added in https://github.com/grafana/logs-drilldown/pull/1294/commits/913989a5970e845dd42992d677cfd1de1a8c95e9

src/Components/EmbeddedLogsExploration/EmbeddedLinkScene.tsx
```diff
@@ -0,0 +1,45 @@
...
+    // @todo how do we keep this up to date if new variables are added?
+    labelsVar.useState();
+    getFieldsVariable(model).useState();
+    getLevelsVariable(model).useState();
+    getMetadataVariable(model).useState();
+    getLineFiltersVariable(model).useState();
+    getPatternsVariable(model).useState();
+    timeRange.useState();
+    const dataSourceVariable = getDataSourceVariable(model);
+    const queryExpr = sceneGraph.interpolate(model, LOG_STREAM_SELECTOR_EXPR);
```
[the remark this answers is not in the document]
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1294/commits/eeccccddeaa3e3a16aeb12c300395af5ac16dd80

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -278,7 +330,53 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     }
   }
 
+  getUrlState() {
```
L2D2Grafana, line 353: Naive question do we need all this code, if users can't access the beginning service scene?

tests/embed.spec.ts
```diff
@@ -0,0 +1,218 @@
+import { Page } from '@playwright/test';
```
L2D2Grafana, line 1: very nice!

L2D2Grafana (commented):
I believe I finally got this PR running 🏃™️, but prolly need a session to go over it. Nice work 🥳
Screenshot 2025-06-09 at 1 36 42 PM

src/Components/EmbeddedLogsExploration/EmbeddedLogs.tsx
```diff
@@ -0,0 +1,65 @@
...
+}: EmbeddedLogsExplorationProps) {
+  const $timeRange = new SceneTimeRange(timeRangeState);
+  $timeRange.subscribeToState((state) => {
+    if (onTimeRangeChange) {
+      onTimeRangeChange(state.value);
+    }
+  });
+
+  if (!query) {
+    return null;
```
matyax, line 26: Could at least probably use a console.error() for development friendlyness.

src/Components/EmbeddedLogsExploration/EmbeddedLogs.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export default function EmbeddedLogsExploration(props: EmbeddedLogsExplorationProps) {
+  const [exploration, setExploration] = useState<IndexScene | null>(null);
+
+  useEffect(() => {
+    if (!exploration) {
+      initializeMetadataService();
+      setExploration(buildLogsExplorationFromState(props));
+    }
+  }, [exploration, props]);
```
matyax, line 55: If props change, we would not generate a new exploration. My suggestion would be to destructure it into individual props and pass them, memoizing whatever we need to memoize to prevent unnecessary regenerations.

src/Components/EmbeddedLogsExploration/types.ts
```diff
@@ -0,0 +1,22 @@
+import { AbstractLabelMatcher, TimeRange } from '@grafana/data';
+import { SceneTimeRangeState } from '@grafana/scenes';
+
+import { IndexSceneState } from 'Components/IndexScene/IndexScene';
+
+interface EmbeddedLogsExplorationFromQuery extends IndexSceneState {
```
matyax: Pet peve of mine that in these scenes app we define everything as possibly undefined, even required attributes. Not only it's incorrect TS, but primarily we can't use it to reliably document to external developers how this is supposed to be used.

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -566,12 +591,12 @@ function getContentScene(drillDownLabel?: string) {
...
 }
 
-function getVariableSet(initialDatasourceUid: string, initialFilters?: AdHocVariableFilter[]) {
+function getVariableSet(initialDatasourceUid: string, initialLabels?: AdHocVariableFilter[], embedded?: boolean) {
   const labelVariable = new AdHocFiltersVariable({
     allowCustomValue: true,
     datasource: EXPLORATION_DS,
     expressionBuilder: renderLogQLLabelFilters,
-    filters: initialFilters ?? [],
+    filters: (initialLabels ?? []).map((f) => ({ ...f, readOnly: true })),
```
matyax: Missing an embedded check, right?

src/Components/LogExplorationPage.tsx
```diff
@@ -6,11 +6,11 @@ import { config } from '@grafana/runtime';
 import { SceneApp, useSceneApp } from '@grafana/scenes';
 
 import { initializeMetadataService } from '../services/metadata';
-import { makeIndexPage, makeRedirectPage } from './Pages';
+import { makeEmbedPage, makeIndexPage, makeRedirectPage } from './Pages';
```
matyax: Nit.

```suggestion
import { makeEmbeddedPage, makeIndexPage, makeRedirectPage } from './Pages';
```

src/Components/Pages.tsx
```diff
@@ -42,6 +48,70 @@ function getServicesScene(routeMatch: OptionalRouteMatch) {
...
 }
 
+function EmbeddedSceneWrapper(props: EmbeddedLogsExplorationProps) {
+  // Component is always null, doesn't look like we can embed something from the same app?
+  const { component: LogsDrilldownComponent, isLoading } = usePluginComponent<EmbeddedLogsExplorationProps>(
+    'grafana-lokiexplore-app/embedded-logs-exploration/v1'
+  );
+
+  // We don't want to re-render the entire app every time the props change, only once when the plugin component is done loading
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
matyax, line 58: This code is a bit odd, becasue when loading is true, you would be returning LogsDrilldownComponent, only to be saved by the condition below.

src/Components/Pages.tsx
```diff
@@ -42,6 +48,70 @@ function getServicesScene(routeMatch: OptionalRouteMatch) {
...
+  const Component = useMemo(() => LogsDrilldownComponent, [isLoading]);
+  return isLoading || !Component ? <LoadingPlaceholder text={'Loading...'} /> : <Component {...props} />;
+}
+
+function getEmbedScene() {
```
matyax: Nit.

```suggestion
function getEmbeddedScene() {
```

src/Components/Pages.tsx
```diff
@@ -71,10 +141,6 @@ export function makeIndexPage() {
         getPage: (routeMatch: RouteMatch, parent) => makeBreakdownValuePage(routeMatch, parent, ValueSlugs.field),
         routePath: CHILD_ROUTE_DEFINITIONS.field,
       },
-      {
-        getPage: () => makeRedirectPage(),
-        routePath: '*',
-      },
```
matyax, line 77: Did we implement a manual version of this route or we don't need it?

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
```diff
@@ -141,8 +142,18 @@ export class FieldsBreakdownScene extends SceneObjectBase<FieldsBreakdownSceneSt
     if (detectedFieldsFrame) {
       this.updateOptions(detectedFieldsFrame);
     }
+  }
+
+  private getPrimaryLabel(): RouteProps {
+    let { breakdownLabel, labelName, labelValue } = getRouteParams(this);
```
matyax, line 148: We could probably check for a label value here. Probably if labelValue is missing or empty it would generate an error state.

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -217,9 +225,19 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
+                throw new Error(`Invalid page slug ${slug}`);
+              }
             } else {
-              navigateToValueBreakdown(getDrilldownValueSlug(), breakdownLabel, this);
+              const valueSlug = this.getDrilldownValueSlug();
+              if (valueSlug) {
+                navigateToValueBreakdown(valueSlug, breakdownLabel, this);
+              } else {
+                throw new Error(`Invalid value slug ${valueSlug}`);
+              }
```
matyax: I wonder if we could gracefully handle these situations, in case the user mistakenly lands here.

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -234,26 +252,60 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
+      getMetadataService().setServiceSceneState(this.state);
+      this._subs.unsubscribe();
 
-    // Redirect to root with updated params, which will trigger history push back to index route, preventing empty page or empty service query bugs
-    navigateToIndex();
+      // Redirect to root with updated params, which will trigger history push back to index route, preventing empty page or empty service query bugs
+      navigateToIndex();
+    } else {
+      // @todo set initial labels?
+      console.warn('Cannot redirect when embedded');
```
matyax: Back to embedded plugin landing page?

src/services/datasource.ts
```diff
@@ -512,7 +512,12 @@ export class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
   }
 }
 
+let initialized = false;
 function init() {
+  if (initialized) {
```
matyax, line 517: console.error(...) for debugging.

src/services/extensions/exposedComponents.tsx
```diff
@@ -19,11 +21,10 @@ function SuspendedOpenInLogsDrilldownButton(props: OpenInLogsDrilldownButtonProp
...
-  {
-    component: SuspendedOpenInLogsDrilldownButton,
-    description: 'A button that opens a logs view in the Logs Drilldown app.',
-    id: `grafana-lokiexplore-app/open-in-explore-logs-button/v1`,
-    title: 'Open in Logs Drilldown button',
-  },
-];
+export function SuspendedEmbeddedLogsExploration(props: EmbeddedLogsExplorationProps) {
+  return (
+    <Suspense fallback={<div>Loading...</div>}>
```
matyax:
```suggestion
    <Suspense fallback={<div>Loading Logs application</div>}>
```

IDk. Sounds a bit less dry.

src/services/navigate.ts
```diff
@@ -81,22 +90,33 @@ export function navigateToValueBreakdown(newPath: ValueSlugs, label: string, ser
...
+    // URL not defined, use url params
+    if (extraQueryParams === undefined) {
+      extraQueryParams = urlUtil.getUrlSearchParams();
+    }
+    extraQueryParams[pageSlugUrlKey] = path;
+    extraQueryParams[drilldownLabelUrlKey] = undefined;
+
+    return buildDrilldownPageUrl(location.pathname, extraQueryParams);
+  } else {
+    throw new Error('Unable to build drilldown tab link!');
```
matyax, line 118: Again, I would be afraid to inadvertedly take the user to a broken app state.

tests/embed.spec.ts
matyax: +1

matyax:
Amazing work! Only minor comments and suggestions. I'll give this a try now.

matyax:
Not sure if I'm missing anything, but I see this in Asserts:

src/Components/ServiceScene/ServiceScene.tsx
gtk-grafana: I'm not sure I understand, are you referring to the `getUrlState` and `updateFromUrl` methods?

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: Hmmm, good point, although I don't think we're currently passing in any initial labels as props, maybe rename `initialLabels` to `readOnlyLabelFilters`?

src/Components/Pages.tsx
gtk-grafana: Cleaned up in https://github.com/grafana/logs-drilldown/pull/1294/commits/3f7e9d4320aa148a414dd55ee6d2cf9ad31f760e, but I'm not sure if I addressed what you were asking?

src/Components/Pages.tsx
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1294/commits/0b898f1781e63ae26a09ea2b3a60ff7c355c11af

src/Components/Pages.tsx
gtk-grafana: We don't need it, we're already redirecting any non-matching routes with the `makeRedirectPage` called in `getSceneApp`.

src/Components/IndexScene/IndexScene.tsx
matyax: Ah yeah, good point. That explains why it wasn't broken when I tried the standalone app. Either double checking embedded or renaming would give a lot of context to any reader.

src/Components/ServiceScene/ServiceScene.tsx
gtk-grafana: Shouldn't be possible outside of a dev environment where we're messing with the routes. I refactored this method in https://github.com/grafana/logs-drilldown/pull/1294/commits/394be0dee14c654f2d3b97bb178d175e2f332c5d to hopefully make it a bit clearer, but let me know if I'm missing something on this?

src/Components/Pages.tsx
matyax: An example:

```tsx
  const Component = useMemo(() => LogsDrilldownComponent, []); // will return two LogsDrilldownComponent, one when isLoading is true, one when isLoading is false
  // Also I don't think we're really memoizing anything
  // Component here holds a LogsDrilldownComponent reference
  return isLoading || !Component ? <LoadingPlaceholder text={'Loading...'} /> : <Component {...props} />;
```

Shouldn't it just be either

```tsx
  const Component = useMemo(() => !isLoading ? <Component {...props} /> : undefined, [isLoading]); 
  return !Component ? <LoadingPlaceholder text={'Loading...'} /> : Component;
```

Or

```tsx
  return isLoading ? <LoadingPlaceholder text={'Loading...'} /> : <Component {...props} />;
```

I find the current implementation very confusing, although it doesn't produce any error.

matyax: Additionally, it's very likely that the re-renders are related with having all the props in `props` and then using the spread operator.

src/Components/ServiceScene/ServiceScene.tsx
gtk-grafana: We'd need an additional prop passed in from embedded plugin to know where to redirect to, and I'd have to understand how a user could get into this state to know if redirecting would be the correct UX. This todo was from before we added the readonly filters, now the initial label set is "locked" I don't think this is necessary.

Cleaned up in https://github.com/grafana/logs-drilldown/pull/1294/commits/97626b2134bce3fa37c18d469ec1910107cb1c68

src/Components/EmbeddedLogsExploration/EmbeddedLogs.tsx
gtk-grafana: Good catch, fixed in https://github.com/grafana/logs-drilldown/pull/1294/commits/31f1c5f776259f0423d2acf21cf023cfa0aaed05, and I also cleaned up an old type that was implying this could be used with an AbstractLabelMatcher, which is not currently supported.

src/Components/ServiceScene/ServiceScene.tsx
gtk-grafana: I would hazard a guess you are conflating the `ServiceScene` and `ServiceSelectionScene`, which to be fair, is confusing now that you can select more labels then just service

src/Components/EmbeddedLogsExploration/types.ts
gtk-grafana: ok it took me a minute but yes I agree, I refactored things here to hopefully make it a bit clearer: https://github.com/grafana/logs-drilldown/pull/1294/commits/3cb7b6c94ee6f6c6082129555afe8c322414cb96

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: Cleaned up in https://github.com/grafana/logs-drilldown/pull/1294/commits/11f3e347ce37ce3427c3d539e9d23383aad71657

src/Components/LogExplorationPage.tsx
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1294/commits/93efb14e7ccfdab5e2e69d2f1e59d142f65fef7c

src/Components/Pages.tsx
gtk-grafana: I refactored this ternary in https://github.com/grafana/logs-drilldown/commit/3f7e9d4320aa148a414dd55ee6d2cf9ad31f760e.
Take a look at https://github.com/grafana/logs-drilldown/pull/1294/commits/068b4554cd753e9b7f028f9f299c998cf6df547f as well which should make it a bit more apparent the case I'm providing for. 

As for the prop spread causing re-renders, isn't that why we're memoizing the exposed component? Let's chat if we're still unaligned here, my jetlag brain could be missing something here.

src/Components/ServiceScene/Breakdowns/FieldsBreakdownScene.tsx
gtk-grafana: Good catch: although we wouldn't be be on this route if we didn't have a full label/value pair, but we could have a mismatch if someone copied partial url params or something weird like that? Fixed in https://github.com/grafana/logs-drilldown/pull/1294/commits/ea826f73b8d13eb07e1edfcb4b52a2cd2a841117

gtk-grafana:
Found a bug when embedded and navigating to the unembedded logs drilldown via the button link, embedded state is still set which is causing errors. Will add test coverage and fix
✅

src/services/datasource.ts
gtk-grafana: I'll take a closer look in a bit but I think this currently fires when not embedded

src/services/extensions/exposedComponents.tsx
gtk-grafana: Updated in https://github.com/grafana/logs-drilldown/pull/1294/commits/3e25817ba566d4cff6f3c22dfc3dabf104760aac

src/services/navigate.ts
gtk-grafana: If this case is hit none of the navigation will work and the app is already in an unrecoverable/useless state, I would argue it's better for it to be obviously broken as we're more likely to catch this in local/dev/ops before being released to actual users.

src/Components/Pages.tsx
matyax: 👌

matyax:
The count resets back from 16k to 350 when going to Patterns.

https://github.com/user-attachments/assets/ad9d39d9-69e6-4ce7-b194-30473bffae0c

gtk-grafana:
> The count resets back from 16k to 350 when going to Patterns.

maybe fixed by the latest push? I'll keep an eye out for this locally, we weren't able to repro on the call.

matyax:
I can no longe repro.

Issue #1300, opened 2025-05-28
[JSON panel]: After adding filters in JSON panel, fields are missing in logs/table panels


gtk-grafana:
**Describe the bug**
If field filters are added in the JSON panel, the other fields disappear from logs/table panels.

[All fields](https://play.grafana.org/a/grafana-lokiexplore-app/explore/service/grafanacon-json-otel/logs?patterns=%5B%5D&from=now-15m&to=now&var-lineFormat=&var-ds=ddhr3fttaw8aod&var-filters=service_name%7C%3D%7Cgrafanacon-json-otel&var-fields=&var-levels=&var-metadata=&var-jsonFields=&var-patterns=&var-lineFilterV2=&var-lineFilters=&timezone=utc&var-all-fields=&displayedFields=%5B%5D&urlColumns=%5B%5D&visualizationType=%22logs%22&prettifyLogMessage=false&sortOrder=%22Descending%22&wrapLogMessage=false):

<img width="820" alt="Image" src="https://github.com/user-attachments/assets/6dd47956-7b33-48da-a4dc-414db535d260" />

After adding a [single filter](https://play.grafana.org/a/grafana-lokiexplore-app/explore/service/grafanacon-json-otel/logs?patterns=%5B%5D&from=now-15m&to=now&var-lineFormat=&var-ds=ddhr3fttaw8aod&var-filters=service_name%7C%3D%7Cgrafanacon-json-otel&var-fields=nested_object%7C%21%3D%7C%7B%22parser%22:%22json%22__gfc__%22value%22:%22%5C%22%5C%22%22%7D,%22%22&var-levels=&var-metadata=&var-jsonFields=nested_object%7C%3D%7C%5B%5C%22nested_object%5C%22%5D&var-patterns=&var-lineFilterV2=&var-lineFilters=&timezone=utc&var-all-fields=nested_object%7C%21%3D%7C%7B%22parser%22:%22json%22__gfc__%22value%22:%22%5C%22%5C%22%22%7D,%22%22&displayedFields=%5B%5D&urlColumns=%5B%5D&visualizationType=%22logs%22&prettifyLogMessage=false&sortOrder=%22Descending%22&wrapLogMessage=false) in the JSON panel:

<img width="846" alt="Image" src="https://github.com/user-attachments/assets/05e09904-8ac8-43db-989e-d03ef1d2d704" />

`msg`, and every other parsed field besides the one that was filtered is now gone!

**To Reproduce**
[...]

PR #1301, opened 2025-05-28
fix(JSON): add second json parser stage

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1300

In Loki, adding arguments to a json parser will tell the parser to only parse the specified fields, this has the unfortunate side-effect of removing parsed fields from the result that are not specified in the parser. However, by adding an additional  json parser stage with no arguments, we get all of the "default" fields parsed in addition to the fields specified in the json parser arguments. 

Note:
This also enables us to remove all the hoop-jumping to sync json parser args and every field filter, which greatly simplifies the most complex part of the JSON viz. Huge win unless I'm missing something else. Let's walk through it:

Before we realized we could double parse we needed a json parser argument for any filter used by the json panel because we didn't know if a filter was added by the json panel or something else. If we didn't have a json parser argument for a filter on a valid field before any parser args were added, the filter would be invalid when we add parser args in the json panel, breaking any query that added filters from outside the json viz. As a workaround we added complex synchronization to add json parser args for every field that was currently filtered, which worked as long as you stayed in the JSON viz for adding fields (as those fields are parsed directly from the json returned by Loki instead of using the fields/labels parsed by Loki), but if you switched to another viz or to the breakdown view the total set of fields was reduced to only the ones that the user had filters defined for (or parent nodes of a nested node set as root).

But now we can assume that every filter targets a field that will continue to exist after adding json parser args, removing that whole ball of yarn from the code, and fixing the bug.


L2D2Grafana (approved):
Thanks for removing this!

Issue #1298, opened 2025-05-23
Bug: Infinite scrolling request contains uninterpolated statements


matyax:
Since some recent changes, the request that the logs panel makes for infinite scrolling contains uninterpolated strings, failing to be processed. An example is: `{service_name=\"grafana/grafana\"}     | json ${jsonFields} | logfmt | drop __error__, __error_details__ `, being `| json ${jsonFields} ` the uninterpolated variable.

PR #1302, opened 2025-05-28
fix(logs-panel): infinite scroll broken

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1298

The logs panel query is interpolated by scenes, and then again by the datasource, but infinite scroll queries is only interpolated once in the datasource.
Since the addition of the `${jsonFields}` variable to the logs panel expression contains a nested variable, which is not interpolated twice in infinite scroll queries, resulting in broken queries.

This PR:
Simply replaces the `${logsFormat}` variable in the `LOG_STREAM_SELECTOR_EXPR` with its interpolated output, as `logsFormat` is not a dynamic variable we should be safe making this fix without any risk of regression.
The jsonFields variable is still being interpolated, but now that interpolation is happening at the first interpolation step with all of the other variables.

Note:
This might make https://github.com/grafana/logs-drilldown/pull/1276 unnecessary


L2D2Grafana (approved):
😎
Uploading Screen Recording 2025-05-29 at 4.55.24 PM 4.59.24 PM.mov…

PR #1315, opened 2025-06-10
feat(embedding): Embedded readonly filters

gtk-grafana:
Fixes bug with readonly filters becoming editable after adding any label and refreshing the page.
Updates scenes to latest.
Adds e2e coverage for readonly filters.
Adds new `embedderName` required prop for embedding application.

image

Parent branch: https://github.com/grafana/logs-drilldown/pull/1294


matyax (approved):

tests/embed.spec.ts
```diff
@@ -237,4 +236,18 @@ test.describe('embed', () => {
     // Assert something from the service selection is visible
     await expect(page.getByTestId(testIds.index.addNewLabelTab)).toHaveCount(1);
   });
+
+  test('can add label and refresh page without losing readonly', async ({ page }) => {
+    async function assertFiltersAreReadOnly() {
+      await expect(page.getByLabel('Edit filter with key service_name')).toHaveCount(1);
+      await expect(page.getByLabel('Edit filter with key cluster')).toHaveCount(1);
+      await expect(page.getByLabel('Remove filter with key service_name')).toHaveCount(0);
+      await expect(page.getByLabel('Remove filter with key cluster')).toHaveCount(1);
```
L2D2Grafana, line 245: Awesome! I also refreshed a bunch and all the filters were still there.

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -203,13 +205,20 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
...
+    } else {
+      this.showLogsError('No logs match your search. Please try a different time range.');
+    }
   }
 
   showLogsError(error: string) {
     const logsVolumeCollapsedByError = this.state.logsVolumeCollapsedByError ?? !getLogsVolumeOption('collapsed');
 
-    this.setState({ error, logsVolumeCollapsedByError });
+    // @here
```
L2D2Grafana: here, here

L2D2Grafana (approved):
lgtm!
Perhaps we hide the welcome message.
Screenshot 2025-06-11 at 10 11 38 AM
Do we need to use some different ids? This error is more of a warning as everything loads
Screenshot 2025-06-11 at 10 23 46 AM
Anyone else getting max series errors running this in only drilldown, embedded seems fine?
Screenshot 2025-06-11 at 10 17 12 AM

src/Components/ServiceScene/LogsPanelScene.tsx
gtk-grafana: Whoopsie. This was a reminder to fix the bug where you could clear readonly filters with the filter button that shows up when there's no data. Fixed in https://github.com/grafana/logs-drilldown/pull/1315/commits/ab2d12e1bfac50b12439f6a8a57f74f2e6705c0b

gtk-grafana:
> Perhaps we hide the welcome message.

Interesting 🤔 this should only show if you haven't seen logs drilldown in this particular Grafana instance (in the current browser), so it might still have some value orienting new users, and once dismissed shouldn't bother folks any more. 

> Do we need to use some different ids? This error is more of a warning as everything loads

Also interesting 🤔 🤔 . I seem to get that on the `/embed` route and not in embedded plugin, so maybe it's thrown because we are embedding something in a plugin that also exposes it? I also seem to get loads of that error from other plugins (investigations, etc) in both contexts so maybe Grafana is calling the thing that attempts to register extensions a bit too liberally. 🤷 Hopefully it shows up when this is in dev/ops and we can ask plugins platform about it.

> Anyone else getting max series errors running this in only drilldown, embedded seems fine?

Yeah that is expected whenever Loki attempts to query something with a higher cardinality then the series limit.

gtk-grafana:
@matyax I just found another bug 😢.
You can remove the readonly filters in the logs/table panels (and in the label drilldown UI), and if you remove all of them it breaks the app.
Let's follow up in another PR

matyax:
Yeah! Let's work on top of this. This is a great initial approach in any case and complex enough in a single pr.

matyax:
Yeah! Let's work on top of this. This is a great initial approach in any case and complex enough in a single pr.

Issue #1318, opened 2025-06-11
[BUG]: Readonly filters in embedded app can be removed in logs/table/drilldown UI


gtk-grafana:
**Describe the bug**
When embedded the embedding context can define readonly filters which should not be removed by the user.
But the user can remove them, just not in the filter variable (combobox).

**Expected behavior**

**Screenshots**

gtk-grafana:
I guess we already extended the `AdHocFiltersVariable` in the `LabelFiltersVariable`, it shouldn't be too hard to just add the filters back in whenever they aren't present on setstate.

But we should also try to sync the UI between the variable filter pills and anywhere we have UI that conflicts with that (make disabled, show the cute little lock icon, etc)

gtk-grafana:
Also @matyax I was wrong, we should not throw a breaking error for `Cannot redirect to start when embedded`, but should redirect back to the original prop state and show a toast congratulating the user for breaking the internet

PR #1323, opened 2025-06-11
fix(EmbeddedLogs): Prevent readonly filter removal

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1318

Also a few other bug fixes:
* Clear filters wasn't working if the label name of the filter that should be removed matched the "primary" label
* If a readonly filter contained a `/` then the primary label value comparison was failing which was triggering an uncessary setState which was triggering queries on a noop (triggered by focusing in/out of the label filter variable without making changes

Also refactors the AdHocVariable extension and cleans stuff up a bit


src/services/ReadOnlyAdHocFiltersVariable.tsx
```diff
@@ -0,0 +1,98 @@
...
+          const filterKey = getFilterSetKey(filter);
+          if (filterSet.has(filterKey)) {
+            return false;
+          } else {
+            filterSet.add(filterKey);
+            return true;
+          }
+        });
+
+        console.warn('Cannot remove readonly filters!');
```
gtk-grafana: I'm not sure we care to warn on this until we've locked down the UI

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -613,7 +613,7 @@ function getVariableSet(
   embedded?: boolean,
   embedderName?: string
 ) {
-  const labelVariable = new LabelFiltersVariable({
+  const labelVariable = new ReadOnlyAdHocFiltersVariable({
```
matyax, line 616: We should conditionally use one or the other, right? I'm probably missing some context.

matyax: I think I get it now. My only nit is that this variable will hold both readonly and not readonly filters.

src/services/variableHelpers.ts
```diff
@@ -1,23 +1,23 @@
...
 
-type ClearableVariable = AdHocFiltersVariable | CustomConstantVariable | LabelFiltersVariable;
+type ClearableVariable = AdHocFiltersVariable | CustomConstantVariable | ReadOnlyAdHocFiltersVariable;
 export function getVariablesThatCanBeCleared(indexScene: IndexScene): ClearableVariable[] {
   const variables = sceneGraph.getVariables(indexScene);
   let variablesToClear: ClearableVariable[] = [];
 
   for (const variable of variables.state.variables) {
-    if (variable.state.name === VAR_LABELS && variable instanceof LabelFiltersVariable) {
+    if (variable instanceof ReadOnlyAdHocFiltersVariable) {
```
matyax, line 20: 👌

matyax (approved):

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: Do you like the previous name better? I renamed it before I made changes that were specific to our usage, but we are assuming that it's only for the labels in other contexts so maybe naming it generically was folly

matyax: Nah, it's fine 👍 . I can't think of a suggestion either.

Issue #1344, opened 2025-06-16
[FEAT]: Field values breakdown - show better error messages


gtk-grafana:
**Description of the new feature**
When users navigate to a field value breakdown, the error messages are a bit cryptic, especially with common errors like exceeding the max series limit in Loki.

**Acceptance Criteria**
A concise description of requirements.

gtk-grafana:
<img width="846" alt="Image" src="https://github.com/user-attachments/assets/6702aae1-5994-40f8-9682-744de7211482" />

PR #1345, opened 2025-06-16
feat(FieldValues): add better max series limit error message

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1344

image

This PR:
* Doesn't remove the field value selection dropdown when an error is thrown
* Adds specific error messages for max series error
* Cleans up error UI a bit

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/QueryErrorContent.tsx
+++ b/src/Components/ServiceScene/Breakdowns/QueryErrorContent.tsx
@@ -1,0 +1,86 @@
+import React from 'react';
+
+import { css } from '@emotion/css';
+
+import { DataQueryError, GrafanaTheme2 } from '@grafana/data';
+import { Alert, useStyles2 } from '@grafana/ui';
+
+import { GrotError } from '../../GrotError';
+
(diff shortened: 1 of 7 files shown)


gtk-grafana:
Do we have other errors that are commonly thrown that we want to gussy up?

src/Components/ServiceScene/Breakdowns/QueryErrorContent.tsx
```diff
@@ -0,0 +1,86 @@
...
+                href="https://grafana.com/docs/loki/latest/configure/#limits_config"
+                className="external-link"
+                rel="noreferrer"
+              >
+                max_query_series
+              </a>{' '}
+              in your Loki configuration.
+            </p>
+            <p>
+              <strong>Workaround:</strong> Reduce the time range, or add additional filters to reduce the number of
```
matyax: Maybe

```suggestion
              <strong>Tip:</strong> Reduce the time range, or add additional filters to reduce the number of
```

gtk-grafana:
@matyax it was a bit bugged when sharding was enabled: updated to dedupe on error message and list out all traceIds. 

image

Issue #1343, opened 2025-06-16
[FEAT]: Aggregated fields breakdown - don't remove errored panels


gtk-grafana:
**Description of the new feature**
Instead of hiding panels when Loki throws an error, let's leave the errored panels visible, and show the error message so users know why they cannot be displayed.

**Acceptance Criteria**
A concise description of requirements.

PR #1346, opened 2025-06-16
feat(FieldsBreakdown): Show panels with errors

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1343

https://github.com/user-attachments/assets/06cfe42e-299b-47dc-8996-6791a5e7d0a2

For reviewers:
Sharded results should show partial results in the panel.
When panels with errors are displayed the user can still get to the value breakdown, which will always show an error, however it is still useful for users that have partial sharded results.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/LayoutSwitcher.tsx
+++ b/src/Components/ServiceScene/Breakdowns/LayoutSwitcher.tsx
@@ -16,10 +16,9 @@ export interface LayoutSwitcherState extends SceneObjectState {
   options: Array<SelectableValue<LayoutType>>;
 }
 
-export type LayoutType = 'grid' | 'rows' | 'single';
+export type LayoutType = 'grid' | 'rows';
 
 export enum LayoutTypeEnum {
-  single = 'single',
   grid = 'grid',
   rows = 'rows',
 }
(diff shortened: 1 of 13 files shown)


matyax:
This is a great improvement overall, and takes the app one step closer to addressing this, but I don't think it fixes #1343, or it does it in a way that is still not fully clear to the user. Maybe we can get to a middle ground by:
- Changing the control to be slightly more verbose. Even with the tooltip, I think a "Show errors" control feels a bit try. An example:
  imagen
- Only show either version of "Show errors" when there are actually panels with errors, not all the time. When there are no errors it will be a noop and will use UI space without providing any service.
- Disable or gray out the "select" button on failed panels?
- We could catch the "max series" error in the querier and replace it with a better message. I can take this part.
- The field breakdown looks currently broken. One thing that we're missing there is a button to return to the fields/label breakdown.
  Captura de pantalla 2025-06-17 a la(s) 10 52 52

With this I think we would have left this in a good enough state to focus on other stuff.

gtk-grafana:
Agreed all around, additionally the value detail panel should only show panel if we have partial/shared data, otherwise just the error splash. Probably remove error splash for partial data?

matyax:
Agree!

gtk-grafana:
> Changing the control to be slightly more verbose.
> Only show either version of "Show errors" when there are actually panels with errors

Added in https://github.com/grafana/logs-drilldown/pull/1346/commits/4561b0b69dccc3424ab1c033bb4418d1b4b98249

gtk-grafana:
> Disable or gray out the "select" button on failed panels?
> The field breakdown looks currently broken. One thing that we're missing there is a button to return to the fields/label breakdown.

Added in https://github.com/grafana/logs-drilldown/pull/1346/commits/f244fc01b1df8e2db72ae8b7798e1ced8bd59693

@matyax if you want to address the max series error message in the panel in another branch I think this one is ready for another look. Sorry about f244fc01b1df8e2db72ae8b7798e1ced8bd59693, it was a big-un.

Oh, I also removed the "single" layout because it's already buggy/broken in main, if we want to support the single view we should support it, but it being in a half supported state isn't good for anyone, plus with the new summary panel that is actually filterable on interaction I don't really see the purpose of the single view in values any more.

src/services/store.ts
```diff
@@ -277,6 +277,15 @@ export function setLogsVisualizationType(type: string) {
   localStorage.setItem(VISUALIZATION_TYPE_LOCALSTORAGE_KEY, type);
 }
 
+const SHOW_ERROR_PANELS_KEY = `${pluginJson.id}.panelOptions.showErrors`;
+export function getShowErrorPanels(): boolean {
+  return !!localStorage.getItem(SHOW_ERROR_PANELS_KEY);
```
matyax, line 282: 💯

src/Components/ServiceScene/Breakdowns/LayoutSwitcher.tsx
```diff
@@ -16,10 +16,9 @@ export interface LayoutSwitcherState extends SceneObjectState {
   options: Array<SelectableValue<LayoutType>>;
 }
 
-export type LayoutType = 'grid' | 'rows' | 'single';
```
matyax, line 19: 👌

Issue #1308, opened 2025-06-04
[FEAT]:Add confiugurable option for the default selected datasource


alexwulf:
**Description of the new feature**
What problem does this feature solve? What are the pain points or missing capabilities?
Currently, the default selected datasource for logs drilldown is set by backend logic of the first datasource it can find, unless the user has set the datasource before and then it will pull from local storage.
Having a configurable default for (each) drilldown page would be useful if a DS is unused for example.

**Acceptance Criteria**
A concise description of requirements.
Have a UI configurable option to choose a default DS for each drilldown page.

PR #1347, opened 2025-06-17
fix(datasources): pull default loki datasource

gtk-grafana:
Addresses: https://github.com/grafana/logs-drilldown/issues/1308

This might not technically classify as a "fix" for #1308, as you can only select a single datasource in Grafana as default, so if folks have a prometheus datasource as default, they can't also select a default Loki datasource for Loki specific applications like Logs Drilldown.

We might want to follow up on adding a default picker to the config...


gtk-grafana:
Let's go with https://github.com/grafana/logs-drilldown/pull/1348 instead

PR #1349, opened 2025-06-17
L2d2/1267 argo auto update ops

L2D2Grafana:
Updating dev with latest main and the argo workflows seems to be going well. I think it is safe to try this in `ops`. I'm hoping we benefit from faster update with hashed named version changes and that the webpack errors will disappear https://raintank-corp.slack.com/archives/C075BDBTX96/p1750103816318619.

- @xnyo let's remove the pin in ops and let me know if you think this is the correct way to not re-run playwright on merge to main `run-playwright: ${{ (github.event_name == 'push' && github.ref_name == 'main') && 'false' || 'true' }}`


matyax (approved):

.github/workflows/pr-ci.yml
```diff
@@ -20,8 +20,8 @@ jobs:
       # Checkout/build PR or main branch, depending on event
       branch: ${{ github.event_name == 'push' && github.ref_name || github.ref }}
 
-      # When pushing to "main", publish and deploy to "dev" (CD). For PRs, skip publishing and deploying (run CI only)
-      environment: ${{ (github.event_name == 'push' && github.ref_name == 'main') && 'dev' || 'none' }}
+      # When pushing to "main", publish and deploy to "dev" and "ops" (CD). For PRs, skip publishing and deploying (run CI only)
+      environment: ${{ (github.event_name == 'push' && github.ref_name == 'main') && 'dev' || 'ops' || 'none' }}
```
xnyo: This should be:

```suggestion
      environment: ${{ (github.event_name == 'push' && github.ref_name == 'main') && 'ops' || 'none' }}
```

If pushing to main -> "ops"

Otherwise (PR) -> "none" (don't trigger Argo CD)

Setting the environment to `ops` will deploy to ops and all the previous environments (so dev + ops).

Same for prod (dev + ops + prod).

With this change it will not re-run Playwright tests and should fix the issue with "Only 'dev' environment is allowed for non-main branches."

xnyo (approved):
LGTM! :tada: However, you may want to hold on merging this for now. The reason is that this will open PRs for each push to main for dev+ops, but ops PRs won't be auto-merged, so it can create noise/confusion in deployment_tools.

We have https://github.com/grafana/plugin-ci-workflows/issues/123 to allow auto-merging ops PRs (opt-in), but we had to revert it due to a bug. I'm currently testing a fix for it: https://github.com/grafana/deployment_tools/pull/281671

L2D2Grafana:
Holding off until the bug fixes go through.

PR #1351, opened 2025-06-17
chore(webpack): use chunk hash as chunk name

gtk-grafana:
Will this break stuff? 

Using the chunkhash as chunk filename so users can cache assets that don't change across deploys.
Also hoping this will make it more apparent when HA Grafana nodes are sending stale assets.

module.js will remain unchanged, but all chunks will now in the the `ld.[chunkhash].js` format. Where [chunkhash](https://webpack.js.org/configuration/output/) is the hash of the content of the hash.


webpack.config.ts
```diff
@@ -3,6 +3,7 @@ import LiveReloadPlugin from 'webpack-livereload-plugin';
 import { merge } from 'webpack-merge';
 
 import grafanaConfig from './.config/webpack/webpack.config';
+// const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
matyax, line 6: We keeping these? I wouldn't mind, just checking.

L2D2Grafana: I deleted bundle analyzer before because I thought we were switching to the plugins-platform bundle analyzer github action. It seems like you use this locally though @gtk-grafana to monitor our package sizes? I'm fine leaving it in, apologies.

L2D2Grafana (approved):
Seeing the built files prefaced with `ld.` (my initials 😂) I'm not sure how much utility we'll get out of this, but I do believe the change is harmless 🤞I think only module.js is sacred. Not sure if special caching or anything is done for vendors or the other files Screenshot 2025-06-18 at 7 27 23 AM
Screenshot 2025-06-18 at 7 26 34 AM

gtk-grafana:
TODO: Discuss with plugins platform before releasing, maybe there's a reason they aren't doing this already, or maybe this is  something we want to change in core.

webpack.config.ts
gtk-grafana: Yeah I use this locally from time to time, good to check bundle splitting/sizes without waiting for CI

webpack.config.ts
```diff
@@ -11,6 +12,9 @@ const config = async (env: any): Promise<Configuration> => {
       // Required to load WASM modules.
       asyncWebAssembly: true,
     },
+    output: {
+      chunkFilename: 'ld.[chunkhash].js',
```
jackw, line 16: Is this attempting to cache bust? We already do that with query strings in the default configs for prod.

https://github.com/grafana/logs-drilldown/blob/main/.config/webpack/webpack.config.ts#L182

tolzhabayev (changes requested):
the solution to this issue is moving to assets being served from Plugins CDN, not changing the hash names - so I would avoid changing the hash names we already have a pattern set in the [webpack.config](https://github.com/grafana/logs-drilldown/blob/main/.config/webpack/webpack.config.ts#L182) in the `.config` folder. Which we also use for CDN caching, so I would avoid changing it.

webpack.config.ts
gtk-grafana: not attempting to cache bust but to allow users to cache assets across deploys when the contents of the chunks haven't changed

webpack.config.ts
jackw: I tend to keep it and use an env to push it into the config...

```ts
const config = {
//  custom configs
}

if (env.stats) {
  config.stats = 'normal';
  config.plugins?.push(new BundleAnalyzerPlugin());
}
```

That way it doesn't cause bother and can be used locally simply enough.

webpack.config.ts
jackw: The query param already provides a way for users to cache each individual chunk across deploys when the contents of the chunk hasn't changed by using `_cache=[contenthash]`. 

image

Using `chunkhash` doesn't have the same effect. That causes the chunk and any of its dependencies to change hash which can result in unnecessary cache invalidation. Also the change here will change the filename which can cause other bother for on prem users if the frontend is stale. Plugins platform tried to weigh these caching options up as best we could for on prem and CDN. Changing this will have an impact for one or the other and I'd suggest the same as @tolzhabayev mentions and invest in getting this plugin on the plugins CDN instead of doing this.

PR #1353, opened 2025-06-19
feat(EmbeddedLogs): embedded logs url parameter namespace

gtk-grafana:
Implementation of https://github.com/grafana/scenes/pull/1156

Adds `ld-` prefix to url parameters (besides `from`, `to`, and `timezone`) to prevent conflicts between embedded plugins.

Note: the embed testing route (`http://localhost:3000/a/grafana-lokiexplore-app/embed`) uses a `SceneAppPage` which is not how embedded components are supposed to be used, this might cause some differences in our e2e env then in a real embedded application.


matyax (approved):

matyax:
Looks good to me! Great job finding this limitation and driving the solution!

I only recommend syncing namespaces with the other apps, so it looks more homogeneous.

Issue #1286, opened 2025-05-19
chore: core/plugin table unification


L2D2Grafana:
- Clean up tech debt, and unify the implementation through the visualization.
- Otel

![Image](https://github.com/user-attachments/assets/5abc8da4-7320-437c-9db5-02b7c82fa0a1)

# Acceptance Criteria 
The Table from Grafana Core is updated, and plugins access the table through the visualization.

L2D2Grafana:
Follow issues have been created.

PR #1354, opened 2025-06-20
feat(table): core table unification

L2D2Grafana:
#1286 

- Removed drawer
- Added resizable left sidebar, matching core grafana
- Added a collapse button for small

https://github.com/user-attachments/assets/831d712c-1bfd-4941-b701-fef2d5e6a2e1

### DIFF
--- a/src/Components/Table/Table.tsx
+++ b/src/Components/Table/Table.tsx
@@ -52,7 +52,7 @@ interface Props {
 }
 
 const getStyles = (theme: GrafanaTheme2, height: number, sideBarWidth: number) => ({
-  // Sidear resize styles matching https://github.com/grafana/grafana/blob/main/public/app/features/explore/Logs/LogsTableWrap.tsx#L561
+  // Sidebar resize styles matching https://github.com/grafana/grafana/blob/main/public/app/features/explore/Logs/LogsTableWrap.tsx#L561
   collapsedTableSidebar: css({
     alignItems: 'center',
     display: 'flex',


gtk-grafana:
@L2D2Grafana the code looks good at a glance, I'll dig in more later this week, but I'm noticing that there are some conflicts with the sidebar from the new logs panel:
image
image

Issues on main/ don't fix in this PR:
The pill heights don't match in the line/body column vs the others any more, but I see that in main as well. 
image

Also if you put a numeric column first the buttons get all messed up
image

L2D2Grafana:
Put back row-reverse to account for the panel controls
Screenshot 2025-06-26 at 2 10 40 PM
Updated Pill heights to be the same
Screenshot 2025-06-27 at 11 14 26 AM
Fix for 1st column numbers
Screenshot 2025-06-27 at 11 07 16 AM

gtk-grafana:
Looking better, a few nits:
When resizing a column the reset custom resize button takes up a bit too much space:
image
vs main:
image


Maybe we should replace those buttons with an IconButton, or just remove it since we added the column reset in the column headers as well.

Also the context menus are missing the "add column" button that's currently present in the line filter pills (but not in standard columns):
image

vs main:
image

Feature request for another PR:
image
Show labels/show raw log line button might need a disabled state if the Log Line column is not visualized.

Feature request for another PR:
Looks like my hacky column width estimator isn't taking the new icons into effect, so we're seeing some columns with titles that are longer then the values push the icons out of the viewport:
image
Although maybe we want to address that after ng-table switch, I think wrapping text in the header was just added (although maybe the icons should overlap or push out the title, but those custom headers are pretty tough to work with)

gtk-grafana:
Also when long labels get `overflow: ellipsis` the sampled cardinality (`n values`) wrap on two lines which looks kinda funky
image

L2D2Grafana:
- Restored add column 😶‍🌫️ 
- Removed reset width buttons
- No wrap on sampled cardinality 
Screenshot 2025-07-08 at 4 51 09 PM

Todo:
Show labels/show raw log line button might need a disabled state if the Log Line column is not visualized.
Column title width with the new controls bar

src/Components/Table/Table.tsx
```diff
@@ -49,16 +51,66 @@ interface Props {
...
+    cursor: 'pointer',
+    padding: theme.spacing(0.5),
+    position: 'absolute',
+    right: theme.spacing(1),
+    top: theme.spacing(1),
+    transition: 'all 0.2s ease-in-out',
+    zIndex: 10,
+  }),
+  rzHandle: css({
+    [theme.transitions.handleMotion('no-preference', 'reduce')]: {
```
gtk-grafana, line 82: I think for background transitions we don't need to respect the reduced motion unless something is actively moving, color/opacity transitions should be fine?

src/Components/Table/Table.tsx
```diff
@@ -49,16 +51,66 @@ interface Props {
...
+      transition: '0.3s background ease-in-out',
+    },
+    ['&:hover']: {
+      background: theme.colors.secondary.shade,
+    },
+    background: theme.colors.secondary.main,
+    borderRadius: theme.shape.radius.pill,
+    cursor: 'grab',
+    height: '50% !important',
```
gtk-grafana, line 91: If there's an !important I'd like to see a comment explaining why there is no other solution available

L2D2Grafana: Added a comment that I'm trying to match the core styles.

L2D2Grafana:
Okay we have some PR fallout, next issues to tackle
🐛 https://github.com/grafana/logs-drilldown/issues/1400
✨ https://github.com/grafana/logs-drilldown/issues/1399

PR #1359, opened 2025-06-25
chore: add more playwright workers

shelldandy:
speed things up!

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -55,5 +55,5 @@ export default defineConfig<PluginOptions>({
   },
 
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 2 : 4,
 });


playwright.config.ts
```diff
@@ -55,5 +55,5 @@ export default defineConfig<PluginOptions>({
   },
 
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 2 : 4,
```
L2D2Grafana, line 58: Totally want to do this! So we run a logs instances in our docker with fake data to simulate realistic api calls. Unfortunately our logs instance can't handle multiple parallel calls 😢, so we are forced to basically run sequentially.  We've talked about mocking the calls or beefing up our logs instance. I'd be interested to know how metrics generates fake data.

gtk-grafana: yeah we've tried this before and we learned that most of our flake comes from the Loki docker image being resource constrained in CI, and running more users in parallel increases that flake rate, which increases the amount of attempts we run tests, which increases the overall time in CI. But this works great locally on beefy hardware

shelldandy: So the way we're doing it is that we run this command from pyroscope for say 1-2 hours.

https://github.com/grafana/pyroscope/blob/main/examples/language-sdk-instrumentation/golang-push/README.md

Which generated a bunch of prometheus data which we then add to our prometheus instance in here:

https://github.com/grafana/metrics-drilldown/blob/main/e2e/docker/Dockerfile.prometheus-static-data

And then we set our e2e fixtures to point to that time range of the static data, that way we can have predictable tests, do screenshots, etc

https://github.com/grafana/metrics-drilldown/blob/main/e2e/config/constants.ts#L8-L13

@grafakus has more details on the setup but i think thats the gist of it

Issue #1357, opened 2025-06-24
[FEAT]: Disable Patterns


gtk-grafana:
**Description of the new feature**
Allow users to disable usage of the Loki patterns API endpoint.

**Acceptance Criteria**
Plugin config which allow users to disable the patterns feature in Logs Drilldown.

PR #1361, opened 2025-06-26
feat(patterns): allow disabling patterns in Logs Drilldown

gtk-grafana:
Adds new option to plugin config (`/plugins/grafana-lokiexplore-app`):
image

When patterns is disabled we don't query patterns, and the tab is removed from the UI:
image

Fixes: https://github.com/grafana/logs-drilldown/issues/1357

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/SelectLabelActionScene.tsx
@@ -263,20 +263,29 @@ export class SelectLabelActionScene extends SceneObjectBase<SelectLabelActionSce
   public onActivate() {
     const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
 
-    if (serviceScene.state.$data?.state.data?.state === LoadingState.Done) {
+    if (
+      serviceScene.state.$data?.state.data?.state === LoadingState.Done ||
+      serviceScene.state.$data?.state.data?.state === LoadingState.Error
+    ) {
       this.calculateSparsity();
     }
 
     this._subs.add(
       sceneGraph.getData(this).subscribeToState((newState) => {
         if (newState.data?.state === LoadingState.Done) {
-          if (serviceScene.state.$data?.state.data?.state === LoadingState.Done) {
+          if (
+            serviceScene.state.$data?.state.data?.state === LoadingState.Done ||
+            serviceScene.state.$data?.state.data?.state === LoadingState.Error
+          ) {
             this.calculateSparsity();
           }
 
           this._subs.add(
             serviceScene.state.$data?.subscribeToState((newLogsPanelState) => {
-              if (newLogsPanelState.data?.state === LoadingState.Done) {
+              if (
+                serviceScene.state.$data?.state.data?.state === LoadingState.Done ||
+                serviceScene.state.$data?.state.data?.state === LoadingState.Error
+              ) {
                 this.calculateSparsity();
               }
             })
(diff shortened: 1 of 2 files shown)


src/Components/AppConfig/AppConfig.tsx
```diff
@@ -105,6 +116,34 @@ const AppConfig = ({ plugin }: Props) => {
...
+            <span>
+              Disables Logs Drilldown&apos;s usage of the{' '}
+              <a
+                className="external-link"
+                href="https://grafana.com/docs/loki/latest/reference/loki-http-api/#patterns-detection"
+                target="_blank"
+                rel="noreferrer"
+              >
+                Loki Patterns API
+              </a>{' '}
```
L2D2Grafana, line 131: Does this actually turn it off in the loki config? https://grafana.com/docs/loki/latest/reference/loki-http-api/#patterns-detection. I thought this was just shutting it down on the ui side.

src/Components/ServiceScene/ActionBarScene.tsx
```diff
@@ -95,36 +95,41 @@ export class ActionBarScene extends SceneObjectBase<ActionBarSceneState> {
...
+                <Tab
+                  data-testid={tab.testId}
+                  key={index}
+                  label={tab.displayName}
+                  active={currentBreakdownViewSlug === tab.value}
+                  counter={loadingStates[tab.displayName] ? undefined : getCounter(tab, state)}
+                  suffix={
+                    tab.displayName === TabNames.logs
+                      ? ({ className }) => LogsCount(className, totalLogsCount, logsCount, maxLines ?? LINE_LIMIT)
+                      : undefined
```
L2D2Grafana, line 114: I do have to refresh the browser for the patterns tab to become hidden.

L2D2Grafana (approved):
It does work once I refresh the page 👍

src/Components/AppConfig/AppConfig.tsx
gtk-grafana: no we can't (and shouldn't) alter the Loki config from Grafana, it just stops us from hitting the API, so if folks don't have it configured they won't see the warning toast on every route change

gtk-grafana:
I'll wait for @matyax to take a look before merge, ideally we could know if the Loki config was set and users could disable things that way. Plus this might only be helpful for OSS folks, cloud folks might need a support ticket to change the plugin config?

src/Components/AppConfig/AppConfig.tsx
```diff
@@ -105,6 +116,34 @@ const AppConfig = ({ plugin }: Props) => {
...
+              endpoint
```
matyax: "and disables the Patterns tab" or something like that, because both things happen.

matyax: Also something about requiring a page refresh/reload.

gtk-grafana: The refresh bit might be too verbose, and isn't specific to patterns.

matyax: Not a blocker, but I'd love to avoid questions like "I disabled patterns but I still see them".

gtk-grafana: image

matyax: Deal! 🤝

gtk-grafana: @matyax ^ what you think?

gtk-grafana: Where's github's warning that comments may not appear until you refresh :P

matyax: They don't know what's good 😎

Issue #1259, opened 2025-05-08
Hide header


gtk-grafana:
Allow users to collapse the header section to dedicate more of the viewport to viewing logs or panels.

gtk-grafana:
I don't think we want to hide variables from view, but I do think it would be nice if we could collapse them when empty and generally make better use of the space in the header.

PR #1366, opened 2025-07-01
feat(header): add collapsing header

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1259

Pros:
* Save vertical space, especially on smaller viewports
* Expand/collapse button is small and will go unnoticed

Cons:
* People are going to forget this is set and have no idea what field/level filters have been set
* Expand/collapse button is small and will go unnoticed
* Removes level filters from view - making it more difficult to quickly change level filters.

I'm not sure the pros outweigh the cons, while there might be some cases when you have many filters that are taking up lots of space, if you are actively refining a search you need to be able to quickly add/remove filters in any context, hiding things will inevitably lead some users to try and click back into fields breakdowns to remove filters instead of using the variables.

Tl;DR - While there might be some use-cases this is helpful, I don't think this is something we'd want to expose by default.

## Tablet widths

image

image

## Desktop widths

image

image

## Desktop widths - many filters

image

image

### DIFF
--- a/src/Components/IndexScene/GiveFeedbackButton.tsx
+++ b/src/Components/IndexScene/GiveFeedbackButton.tsx
@@ -35,9 +35,7 @@ const getStyles = (theme: GrafanaTheme2) => {
     wrapper: css({
       display: 'flex',
       gap: theme.spacing(1),
-      marginLeft: 'auto',
       position: 'relative',
-      top: theme.spacing(-1),
     }),
   };
 };
--- a/src/Components/IndexScene/VariableLayoutScene.tsx
+++ b/src/Components/IndexScene/VariableLayoutScene.tsx
@@ -1,12 +1,12 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import { css, cx } from '@emotion/css';
 
 import { GrafanaTheme2 } from '@grafana/data';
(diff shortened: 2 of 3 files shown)


gtk-grafana:
@matyax seeing this was enough to illustrate how terrible of an idea this is, if we want to support the ability to use the entire viewport to read logs we should just bump up the height of the logs/table/json panel to 100vh and users can scroll to get a full viewport experience.

gtk-grafana:
Or a full viewport modal like mode for panels, like panelMenu > view in dashboards, but even that view keeps all of the variables in view, so I'm not sure this is a pattern we want to support.

Making the variables smaller or wrap better might be a better direction to take this.

matyax:
Yeah. I don't really remember the context for https://github.com/grafana/logs-drilldown/issues/1259 but I don't think it's something we should do, specially after seeing these examples. 

> Making the variables smaller or wrap better might be a better direction to take this.

Absolutely agree. The tall header is a consequence of the controls we can use; not really a problem of the header itself. If we can further improve how we use that space, without drifting too much from the other Drilldown apps, it would be better. I wonder if we can try omitting all the variable labels (Labels, Log Levels, Fields, etc), and rely on placeholders and tooltips.

matyax:
Traces uses inline labels, but I'm not 100% sold on that yet.

imagen

Update: notice the general "Filters" label and then the other two controls without label.

Issue #1303, opened 2025-05-28
[JSON panel]: Filter icons not showing up in logs panel when navigating from the "fields" tab


gtk-grafana:
**Describe the bug**
Filter icons not showing they should.

<img width="930" alt="Image" src="https://github.com/user-attachments/assets/050b7b1b-98fd-4710-9a12-58b20a3e34c2" />

Looks like we avoid running detected_fields on the logs panel when it's already been run in another scene, we probably need to kick it off in the json panel in this situation.

**To Reproduce**
1. Go to https://play.grafana.org/a/grafana-lokiexplore-app/explore/service/grafanacon-json-otel/fields?patterns=%5B%5D&from=now-15m&to=now&var-lineFormat=&var-ds=ddhr3fttaw8aod&var-filters=service_name%7C%3D%7Cgrafanacon-json-otel&var-fields=&var-levels=&var-metadata=&var-jsonFields=&var-patterns=&var-lineFilterV2=&var-lineFilters=&displayedFields=%5B%5D&urlColumns=%5B%5D&visualizationType=%22json%22&var-fieldBy=$__all&timezone=utc&var-all-fields=
2. Click on Logs tab
3. Set panel type to JSON
4. See missing icons

Or:
1. Reloading the logs tab while the JSON viz is active triggers a race condition of the scene activation and the detected_fields response.

**Workaround**
Refresh the query.

**Expected behavior**
Filter icons should show

PR #1368, opened 2025-07-02
fix(JSON): detected fields not always getting called on activation

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1303

Prevents:
* Race condition when loading json viz and detected_fields call is in flight
* Detected_fields not getting called when cached while navigating from the fields tab to the json viz.

Adding e2e coverage for both of these cases as well.

### DIFF
--- a/src/Components/ServiceScene/LogsJsonScene.tsx
+++ b/src/Components/ServiceScene/LogsJsonScene.tsx
@@ -181,7 +181,16 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
       } else {
         this.setVizFlags(detectedFieldFrame);
       }
+    } else if (serviceScene.state?.$detectedFieldsData?.state.data?.state === undefined) {
+      serviceScene.state?.$detectedFieldsData?.runQueries();
     }
+
+    // Subscribe to detected fields
+    serviceScene.state?.$detectedFieldsData?.subscribeToState((newState) => {
+      if (newState.data?.state === LoadingState.Done && newState.data?.series.length) {
+        this.setVizFlags(newState.data.series[0]);
+      }
+    });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -181,7 +181,16 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
       } else {
         this.setVizFlags(detectedFieldFrame);
       }
+    } else if (serviceScene.state?.$detectedFieldsData?.state.data?.state === undefined) {
```
gtk-grafana, line 184: I guess if we wanted an abundance of caution we should throw a console warning if the service scene doesn't have a $detectedFieldsData query runner

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -181,7 +181,16 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
+      serviceScene.state?.$detectedFieldsData?.runQueries();
     }
+
+    // Subscribe to detected fields
```
matyax, line 188: Does this guarantee that you would have something to subscribe to? What happens if this is still undefined?

My guess is that it would probably be handled by the original code, but just making sure.

gtk-grafana: Yeah if the query is undefined on activate we'll manually run the query, if the $detectedFieldsData is undefined then we're including this component in a scene tree that doesn't have a serviceScene parent with a detected_fields query runner, so then nothing is gonna work

gtk-grafana:
I forgot to clean up the subscription 🤦 , luckily caught it before merge

matyax:
Yeah, it's easy to miss. GJ.

Issue #1367, opened 2025-07-01
[FEAT]: JSON panel - support structured metadata


gtk-grafana:
**Description of the new feature**
Allow users to filter on indexed labels and structured metadata directly in the json viz.

**Acceptance Criteria**
Metadata key/values show in JSON panel view.

**Questions**
Toggle to show parsed fields or always on by default?

PR #1370, opened 2025-07-02
feat(JSON): support labels/metadata filtering in JSON viz

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1367

Allows users to toggle metadata/labels as nodes in the JSON viewer.
Supports filtering values in/out just like table/logs panel.

image
image
image

### DIFF
--- a/src/Components/ServiceScene/LogsJsonScene.tsx
+++ b/src/Components/ServiceScene/LogsJsonScene.tsx
@@ -81,7 +81,14 @@ import { getDetectedFieldsFrameFromQueryRunnerState, getLogsPanelFrame, ServiceS
 import { JSONTree, KeyPath } from '@gtk-grafana/react-json-tree';
 import { logger } from 'services/logger';
 import { logsControlsSupported } from 'services/panel';
-import { getLogOption, setLogOption } from 'services/store';
+import {
+  getJsonLabelsVisibility,
+  getJsonMetadataVisibility,
+  getLogOption,
+  setJsonLabelsVisibility,
+  setJsonMetadataVisibility,
+  setLogOption,
+} from 'services/store';
 
 interface LogsJsonSceneState extends SceneObjectState {
   data?: PanelData;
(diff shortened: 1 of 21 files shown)


matyax:
This is great! Just tested it and works as expected. I like that the labels and fields are each in their own category.

What do you think, as icons, for labels:

imagen

And alternative for metadata, although the key could work:

imagen

gtk-grafana:
So I took a look at expanding/collapsing metadata on toggle instead of showing/hiding nodes so users could manually open nodes instead of just expanding/collapsing all label/metadata, but it's not going to be easy without forcing a full render of the viz (and resetting the state of what nodes are currently expanded/collapsed) or changing some of the internals of the JSON viz library to allow us to update the react keys on nested nodes to selectively reset state of specific nodes without resetting everything. 

Not anything we need to address right now, but we might need to revisit this if/when we want to support "expand/collapse all nodes that match this keyPath" functionality.

gtk-grafana:
@matyax Fixed the metadata bug, updated the icons, and switched to the monospace font to match the logs panel.
Gonna open it up for review 🎉

gtk-grafana:
Although I found a bug, without the new logs panel enabled we don't have the options menu: 
image

I'll add some e2e coverage for the new options as well

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -87,14 +105,26 @@ interface LogsJsonSceneState extends SceneObjectState {
...
 export type NodeTypeLoc = 'Array' | 'Boolean' | 'Custom' | 'Number' | 'Object' | 'String';
 export type AddJSONFilter = (keyPath: KeyPath, key: string, value: string, filterType: FilterType) => void;
+export type AddMetadataFilter = (
+  key: string,
+  value: string,
+  filterType: FilterType,
+  variableType: InterpolatedFilterType
+) => void;
 
 const DataFrameTimeName = 'Time';
```
matyax, line 122: Minor, but we generally capitalize all these constants.

matyax (approved):
Works as expected and looks much better!

matyax:
Currently, the main challenge of this component is related with performance. It would be positive to spend some time optimizing re-renders and any recalculation.

Issue #1373, opened 2025-07-02
[FEAT]: Fields breakdown - alternative views


gtk-grafana:
**Description of the new feature**
Instead of only showing volume breakdown for each parsed field, users should be able to show the cardinality of each field as a gauge, or just the name of the field

**Why tho?**
Huge Loki instances or queries run over long time ranges can be slow, especially in the fields view. 
Users should be able to find specific fields to filter and decide when the logs space is narrow enough to support showing the full time series for every field in the UI. 

**Ideas**
* Name only view
* Sampled cardinality view (gauge - detected_fields response)
* Exact cardinality view (gague - instant query)
* Full time series view (what we currently display)

gtk-grafana:
WIP PoC:

Default volume UI (what we have now):
<img width="1727" alt="Image" src="https://github.com/user-attachments/assets/313c5fb3-9f92-466a-b211-68170948ba59" />

Sampled cardinality (no queries)
<img width="1726" alt="Image" src="https://github.com/user-attachments/assets/917c0c35-729d-41c9-9a06-1b86353c1799" />

Cardinality (instant query):

<img width="1726" alt="Image" src="https://github.com/user-attachments/assets/763c5c9f-cdbe-4bad-a2d2-a2d58b6b0caf" />

gtk-grafana:
I think there's loads of improvement here, just playing around for now.
It is really nice that the estimated cardinality/freq returns results instantly without waiting for any queries.

Any other ideas are appreciated, I'll play around with this a bit more tomorrow and hopefully get something ready for review.

gtk-grafana:
Closed by https://github.com/grafana/logs-drilldown/pull/1374, but we might want to follow up with more views as suggested in this issue.

PR #1374, opened 2025-07-02
feat(fields): field name view

gtk-grafana:
MVP of https://github.com/grafana/logs-drilldown/issues/1373.

Loads a no-query view of text panels

image


matyax:
I like this a lot! Also how it's integrated with the tab.

imagen

What I do see is that the sampled cardinality is both innacurate and potentialy confusing. What about just showing it without any metric?

It would also be interesting to see how the instant queries look like for the values breakdown. Great work!

matyax:
@trevorwhitney Do you think having an alternative view that users an instant query for metrics instead of a full range query would improve responsiveness/decrease work load? For the fields breakdown, btw.

gtk-grafana:
> What I do see is that the sampled cardinality is both innacurate and potentialy confusing. What about just showing it without any metric?

It's not inaccurate, just a sample, the tooltip should call this out, and why I wanted to include "sampled" in the toggle, but yeah it could cause confusion and we could certainly just show the names, I wanted to go big for PoC to help spark ideation, but we can dial it back for MVP if we decide to move forward on something like this.

> It would also be interesting to see how the instant queries look like for the values breakdown. Great work!

We typically show instant query results in a table, which is probably the best, but not as sexy as a barchart (but barchart would be unreadable for fields with cardinality greater then 10-20). I wonder if we could get a stat and a table working, either as a panel menu toggle or just by shoving them both in the panel?

src/services/fields.ts
```diff
@@ -142,6 +143,58 @@ export function getDetectedFieldsJsonPathField(detectedFieldsFrame?: DataFrame)
...
+  return undefined;
+}
+
+export interface SparsityCalculation {
+  cardinality: number | undefined;
+  description: string | undefined;
+  sparsity: number | undefined;
+}
+
+export function calculateSparsity(sceneRef: SceneObject, labelName: string): SparsityCalculation {
```
L2D2Grafana, line 169: Too me this function is calculating label prevalence or label coverage (names are hard), sparsity usually refers to zero or null datasets.

L2D2Grafana (commented):
This is neat-o 👏 ! 

- Many bar gauges can become bar gauges for 🐜 
Screenshot 2025-07-11 at 10 07 44 AM
- I hit this error a few times and my computer tanked, no sure how real a problem that would be in prod
Screenshot 2025-07-11 at 10 07 15 AM

gtk-grafana:
@L2D2Grafana whoa I never saw more then 2, one of the transformations must be buggy, should only ever show the cardinality and frequency

src/Components/ServiceScene/Breakdowns/FieldSelector.tsx
```diff
@@ -139,7 +139,7 @@ function getStyles(theme: GrafanaTheme2) {
     selectWrapper: css({
       label: 'field-selector-select-wrapper',
       marginBottom: 0,
-      marginRight: theme.spacing.x1,
+      // marginRight: theme.spacing.x1,
```
matyax:
```suggestion
```

src/Components/ServiceScene/Breakdowns/FieldsAggregatedBreakdownScene.tsx
```diff
@@ -46,17 +57,23 @@ import {
 import { FIELDS_BREAKDOWN_GRID_TEMPLATE_COLUMNS, FieldsBreakdownScene } from './FieldsBreakdownScene';
 import { LayoutSwitcher } from './LayoutSwitcher';
 import { SelectLabelActionScene } from './SelectLabelActionScene';
+import { ShowErrorPanelToggle } from './ShowErrorPanelToggle';
+import { ShowFieldDisplayToggle } from './ShowFieldDisplayToggle';
 import { MAX_NUMBER_OF_TIME_SERIES } from './TimeSeriesLimit';
 
+export type FieldsPanelsType = 'text' | 'volume';
```
matyax: Minor, but the visible name used was "names" and "volume".

src/Components/ServiceScene/Breakdowns/FieldsAggregatedBreakdownScene.tsx
```diff
@@ -457,45 +537,26 @@ export class FieldsAggregatedBreakdownScene extends SceneObjectBase<FieldsAggreg
...
+
+export function estimatedCardinality(ctx: DataTransformContext, sparsity: SparsityCalculation) {
+  return (source: Observable<DataFrame[]>) => {
+    return source.pipe(
+      map((frames) => {
+        const resultFrames = [
+          toDataFrame({
+            fields: [
+              {
+                name: GAUGE_CARDINALITY_FIELD_NAME,
```
matyax: This isn't used, right?

matyax (commented):
This is great! The difference in speed, when you don't care about volume, is substantial.

src/Components/ServiceScene/Breakdowns/FieldsAggregatedBreakdownScene.tsx
gtk-grafana: Renamed to match the panel type that is created instead of the label name in the UI in https://github.com/grafana/logs-drilldown/pull/1374/commits/5edef4fff5ee302d4bfef5823292647da11939d9

matyax (approved):
👌

src/Components/ServiceScene/Breakdowns/FieldsAggregatedBreakdownScene.tsx
gtk-grafana: No, cleaned up in https://github.com/grafana/logs-drilldown/pull/1374/commits/dcdac39ae1ca8c053ea194ffb14663d02172bb24

PR #1376, opened 2025-07-03
feat(Embbedding): open embedding API for default line filters

gtk-grafana:
Allow embedded Logs Drilldown parse default line filters from query string prop.


matyax (approved):
Can't test it right now but the changes look good!

matyax (approved):

PR #1382, opened 2025-07-03
JSON - Highlighting 

gtk-grafana:
Adds syntax highlighting adapted from the logs panel
Highlights line filter matches

Note: I doubt we'll be able to avoid prismjs or some other external package, but I wanted to give it a shot.

### DIFF
--- a/src/Components/ServiceScene/LogListControls.tsx
+++ b/src/Components/ServiceScene/LogListControls.tsx
@@ -64,7 +64,7 @@ export const LogListControls = ({
           name="document-info"
           className={showMetadata ? styles.controlButtonActive : styles.controlButton}
           onClick={() => onToggleStructuredMetadataClick(!showMetadata)}
-          tooltip={'Show structured metadata'}
+          tooltip={showMetadata ? 'Hide structured metadata' : 'Show structured metadata'}
           size="lg"
         />
       )}
@@ -73,7 +73,7 @@ export const LogListControls = ({
           name="key-skeleton-alt"
           className={showLabels ? styles.controlButtonActive : styles.controlButton}
           onClick={() => onToggleLabelsClick(!showLabels)}
-          tooltip={'Show labels'}
+          tooltip={showLabels ? 'Hide Labels' : 'Show labels'}
           size="lg"
         />
       )}
--- a/src/Components/ServiceScene/LogListControls.tsx
+++ b/src/Components/ServiceScene/LogListControls.tsx
@@ -14,8 +14,10 @@ interface Props {
   onScrollToBottomClick?(): void;
   onScrollToTopClick?(): void;
   onSortOrderChange(newOrder: LogsSortOrder): void;
+  onToggleHighlightClick?(visible: boolean): void;
   onToggleLabelsClick?(visible: boolean): void;
   onToggleStructuredMetadataClick?(visible: boolean): void;
+  showHighlight?: boolean;
   showLabels?: boolean;
   showMetadata?: boolean;
   sortOrder: LogsSortOrder;
@@ -28,8 +30,10 @@ export const LogListControls = ({
   onScrollToBottomClick,
   onScrollToTopClick,
   onSortOrderChange,
+  onToggleHighlightClick,
   onToggleLabelsClick,
   onToggleStructuredMetadataClick,
+  showHighlight,
   showLabels,
   showMetadata,
   sortOrder,
@@ -62,6 +66,7 @@ export const LogListControls = ({
       {showMetadata !== undefined && onToggleStructuredMetadataClick && (
         <IconButton
           name="document-info"
+          aria-pressed={showMetadata}
           className={showMetadata ? styles.controlButtonActive : styles.controlButton}
           onClick={() => onToggleStructuredMetadataClick(!showMetadata)}
           tooltip={showMetadata ? 'Hide structured metadata' : 'Show structured metadata'}
@@ -71,9 +76,21 @@ export const LogListControls = ({
       {showLabels !== undefined && onToggleLabelsClick && (
         <IconButton
           name="key-skeleton-alt"
+          aria-pressed={showLabels}
           className={showLabels ? styles.controlButtonActive : styles.controlButton}
           onClick={() => onToggleLabelsClick(!showLabels)}
           tooltip={showLabels ? 'Hide Labels' : 'Show labels'}
+          size="lg"
+        />
+      )}
+
+      {showHighlight !== undefined && onToggleHighlightClick && (
+        <IconButton
+          name="brackets-curly"
+          aria-pressed={showHighlight}
+          className={showHighlight ? styles.controlButtonActive : styles.controlButton}
+          onClick={() => onToggleHighlightClick(!showHighlight)}
+          tooltip={showHighlight ? 'Disable highlighting' : 'Enable highlighting'}
           size="lg"
         />
       )}
--- a/src/Components/ServiceScene/LogListControls.tsx
+++ b/src/Components/ServiceScene/LogListControls.tsx
@@ -14,6 +14,8 @@ interface Props {
   onScrollToBottomClick?(): void;
   onScrollToTopClick?(): void;
   onSortOrderChange(newOrder: LogsSortOrder): void;
+  onToggleLabelsClick?(): void;
+  onToggleStructuredMetadataClick?(): void;
   sortOrder: LogsSortOrder;
 }
 
@@ -24,6 +26,8 @@ export const LogListControls = ({
   onScrollToBottomClick,
   onScrollToTopClick,
   onSortOrderChange,
+  onToggleLabelsClick,
+  onToggleStructuredMetadataClick,
   sortOrder,
 }: Props) => {
   const styles = useStyles2(getStyles);
@@ -51,6 +55,24 @@ export const LogListControls = ({
         tooltip={sortOrder === LogsSortOrder.Descending ? 'Newest logs first' : 'Oldest logs first'}
         size="lg"
       />
+      {onToggleStructuredMetadataClick && (
+        <IconButton
+          name="document-info"
+          className={styles.controlButton}
+          onClick={onToggleStructuredMetadataClick}
+          tooltip={'Show structured metadata'}
+          size="lg"
+        />
+      )}
+      {onToggleLabelsClick && (
+        <IconButton
+          name="key-skeleton-alt"
+          className={styles.controlButton}
+          onClick={onToggleLabelsClick}
+          tooltip={'Show labels'}
+          size="lg"
+        />
+      )}
       {onManageColumnsClick && (
         <IconButton
           name="columns"
--- a/src/Components/ServiceScene/LogListControls.tsx
+++ b/src/Components/ServiceScene/LogListControls.tsx
@@ -14,8 +14,10 @@ interface Props {
   onScrollToBottomClick?(): void;
   onScrollToTopClick?(): void;
   onSortOrderChange(newOrder: LogsSortOrder): void;
-  onToggleLabelsClick?(): void;
-  onToggleStructuredMetadataClick?(): void;
+  onToggleLabelsClick?(visible: boolean): void;
+  onToggleStructuredMetadataClick?(visible: boolean): void;
+  showLabels?: boolean;
+  showMetadata?: boolean;
   sortOrder: LogsSortOrder;
 }
 
@@ -28,6 +30,8 @@ export const LogListControls = ({
   onSortOrderChange,
   onToggleLabelsClick,
   onToggleStructuredMetadataClick,
+  showLabels,
+  showMetadata,
   sortOrder,
 }: Props) => {
   const styles = useStyles2(getStyles);
@@ -55,20 +59,20 @@ export const LogListControls = ({
         tooltip={sortOrder === LogsSortOrder.Descending ? 'Newest logs first' : 'Oldest logs first'}
         size="lg"
       />
-      {onToggleStructuredMetadataClick && (
+      {showMetadata !== undefined && onToggleStructuredMetadataClick && (
         <IconButton
           name="document-info"
-          className={styles.controlButton}
-          onClick={onToggleStructuredMetadataClick}
+          className={showMetadata ? styles.controlButtonActive : styles.controlButton}
+          onClick={() => onToggleStructuredMetadataClick(!showMetadata)}
           tooltip={'Show structured metadata'}
           size="lg"
         />
       )}
-      {onToggleLabelsClick && (
+      {showLabels !== undefined && onToggleLabelsClick && (
         <IconButton
           name="key-skeleton-alt"
-          className={styles.controlButton}
-          onClick={onToggleLabelsClick}
+          className={showLabels ? styles.controlButtonActive : styles.controlButton}
+          onClick={() => onToggleLabelsClick(!showLabels)}
           tooltip={'Show labels'}
           size="lg"
         />
@@ -109,6 +113,22 @@ export const LogListControls = ({
 const getStyles = (theme: GrafanaTheme2) => {
   return {
     controlButton: css({
+      color: theme.colors.text.secondary,
+      height: theme.spacing(2),
+      margin: 0,
+    }),
+    controlButtonActive: css({
+      '&:after': {
+        backgroundImage: theme.colors.gradients.brandHorizontal,
+        borderRadius: theme.shape.radius.default,
+        bottom: theme.spacing(-1),
+        content: '" "',
+        display: 'block',
+        height: 2,
+        opacity: 1,
+        position: 'absolute',
+        width: '95%',
+      },
       color: theme.colors.text.secondary,
       height: theme.spacing(2),
       margin: 0,
--- a/src/Components/ServiceScene/LogsJsonScene.tsx
+++ b/src/Components/ServiceScene/LogsJsonScene.tsx
@@ -81,7 +81,14 @@ import { getDetectedFieldsFrameFromQueryRunnerState, getLogsPanelFrame, ServiceS
 import { JSONTree, KeyPath } from '@gtk-grafana/react-json-tree';
 import { logger } from 'services/logger';
 import { logsControlsSupported } from 'services/panel';
-import { getLogOption, setLogOption } from 'services/store';
+import {
+  getJsonLabelsVisibility,
+  getJsonMetadataVisibility,
+  getLogOption,
+  setJsonLabelsVisibility,
+  setJsonMetadataVisibility,
+  setLogOption,
+} from 'services/store';
 
 interface LogsJsonSceneState extends SceneObjectState {
   data?: PanelData;
@@ -91,6 +98,8 @@ interface LogsJsonSceneState extends SceneObjectState {
   // If undefined, we haven't detected the loki version yet; if false, jsonPath (loki 3.5.0) is not supported
   jsonFiltersSupported?: boolean;
   menu?: PanelMenu;
+  showLabels: boolean;
+  showMetadata: boolean;
   sortOrder: LogsSortOrder;
 }
 
@@ -105,6 +114,8 @@ export type AddMetadataFilter = (
 
 const DataFrameTimeName = 'Time';
 const DataFrameLineName = 'Line';
+const StructuredMetadataDisplayName = 'Metadata';
+const LabelsDisplayName = 'Labels';
 const DataFrameStructuredMetadataName = '__Metadata';
 const DataFrameLabelsName = '__Labels';
 const VizRootName = 'root';
@@ -117,6 +128,8 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
   constructor(state: Partial<LogsJsonSceneState>) {
     super({
       ...state,
+      showLabels: getJsonLabelsVisibility(),
+      showMetadata: getJsonMetadataVisibility(),
       sortOrder: getLogOption<LogsSortOrder>('sortOrder', LogsSortOrder.Descending),
     });
 
@@ -175,6 +188,7 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
         }
       })
     );
+
     clearJsonParserFields(this);
 
     const detectedFieldFrame = getDetectedFieldsFrameFromQueryRunnerState(
@@ -203,6 +217,15 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
         this.setVizFlags(newState.data.series[0]);
       }
     });
+
+    // Subscribe to options state changes
+    this._subs.add(
+      this.subscribeToState((newState, prevState) => {
+        if (newState.showMetadata !== prevState.showMetadata || newState.showLabels !== prevState.showLabels) {
+          this.transformDataFrame($data.state);
+        }
+      })
+    );
   }
 
   /**
@@ -421,7 +444,8 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
 
   public static Component = ({ model }: SceneComponentProps<LogsJsonScene>) => {
     // const styles = getStyles(grafanaTheme)
-    const { data, emptyScene, hasJsonFields, jsonFiltersSupported, menu, sortOrder } = model.useState();
+    const { data, emptyScene, hasJsonFields, jsonFiltersSupported, menu, showLabels, showMetadata, sortOrder } =
+      model.useState();
     const $data = sceneGraph.getData(model);
     // Rerender on data change
     $data.useState();
@@ -435,16 +459,8 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
     // If we have a line format variable, we are drilled down into a nested node
     const dataFrame = getLogsPanelFrame(data);
     const lineField = dataFrame?.fields.find((field) => field.type === FieldType.string && isLogLineField(field.name));
-    const labelsField = dataFrame?.fields.find((field) => field.type === FieldType.other && isLabelsField(field.name));
-    const labelTypesField = dataFrame?.fields.find(
-      (field) => field.type === FieldType.other && isLabelTypesField(field.name)
-    );
-
-    // const data = { ...lineField.values };
-
-    console.log('dataframe', { dataFrame, labelsField, labelTypesField, lineField });
-
     const jsonParserPropsMap = new Map<string, AdHocFilterWithLabels>();
+
     jsonVar.state.filters.forEach((filter) => {
       // @todo this should probably be set in the AdHocFilterWithLabels valueLabels array
       // all json props are wrapped with [\" ... "\], strip those chars out so we have the actual key used in the json
@@ -464,9 +480,21 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
       scrollRef.current?.scrollTo(0, 0);
     }, []);
 
-    const onToggleStructuredMetadataClick = useCallback(() => {}, []);
+    const onToggleStructuredMetadataClick = useCallback(
+      (visible: boolean) => {
+        model.setState({ showMetadata: visible });
+        setJsonMetadataVisibility(visible);
+      },
+      [model]
+    );
 
-    const onToggleLabelsClick = useCallback(() => {}, []);
+    const onToggleLabelsClick = useCallback(
+      (visible: boolean) => {
+        model.setState({ showLabels: visible });
+        setJsonLabelsVisibility(visible);
+      },
+      [model]
+    );
 
     return (
       // @ts-expect-error todo: fix this when https://github.com/grafana/grafana/issues/103486 is done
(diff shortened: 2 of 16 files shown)


src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -143,13 +149,254 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
+                    This view is built for JSON log lines, but none were detected. Switch to the Logs or Table view for
+                    a better experience.
+                  </Alert>
+                </>
+              )}
+              <JSONTree
+                data={lineField.values}
+                hideRootExpand={true}
+                valueWrap={''}
+                getItemString={(_, data, itemType, itemString, keyPath) => {
```
matyax: Extracting these functions would be great and potentially aid with performance. Every inline function will be a potential performance regression.

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -105,19 +140,263 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
+                  }
+                  const value = valueAsString?.toString();
+                  if (!value) {
+                    return null;
+                  }
+
+                  // I don't typically like clever crap like this, but this is considerably smaller then prism, Matyax will let me know why this won't work
+                  // Ideally we would use the re2 package that is already being used in Logs Drilldown instead of evaluating js RegExp, but this is how it works in core as well AFAIK
+                  const highlights = logsHighlightMatches(lineFilterVar.state.filters.map((f) => f.value));
+                  const splitMatches = Object.keys(highlights).map((key) => value.split(highlights[key]));
```
gtk-grafana: split doesn't catch matches that overlap, so this won't work for all cases.

src/Components/ServiceScene/LogsJsonScene.tsx
gtk-grafana: They've been extracted into new components in new files, should be a bit easier to work with now.

.eslintrc
```diff
@@ -37,7 +37,8 @@
       "files": ["src/**/*.test.{ts,tsx}"],
       // Allow location to be used in test files
       "rules": {
-        "no-restricted-globals": 0
+        "no-restricted-globals": 0,
+        "react/jsx-key": [1, { "checkFragmentShorthand": true }]
```
L2D2Grafana, line 41: This appears unused.

src/Components/ServiceScene/LogListControls.tsx
```diff
@@ -71,12 +76,24 @@ export const LogListControls = ({
...
+      {showHighlight !== undefined && onToggleHighlightClick && (
+        <IconButton
+          name="brackets-curly"
+          aria-pressed={showHighlight}
+          className={showHighlight ? styles.controlButtonActive : styles.controlButton}
+          onClick={() => onToggleHighlightClick(!showHighlight)}
+          tooltip={showHighlight ? 'Disable highlighting' : 'Enable highlighting'}
+          size="lg"
+        />
+      )}
```
L2D2Grafana, line 96: Seems like we're overloading the button icons here with double meanings. In core the brackets are to collapse json, and a tag icon is used for labels. 
Screenshot 2025-07-07 at 4 12 39 PM

src/services/highlight.tsx
```diff
@@ -0,0 +1,228 @@
...
+  'log-token-method': /\b(GET|POST|PUT|DELETE|PATCH|HEAD|OPTIONS|TRACE|CONNECT)\b/g,
+  'log-token-size': /(?:\b|")\d+\.{0,1}\d*\s*[kKmMGgtTPp]*[bB]{1}(?:"|\b)/g,
+  'log-token-trace': /(\b)(TRACE)($|\s)/gi,
+  'log-token-uuid': /[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}/g,
+  'log-token-warning': /(\b)(WARNING|WARN)($|\s)/gi,
+
+  // JSON values should not contain these
+  // 'log-token-string': /"(?!:)([^'"])*?"(?!:)/g,
+  // 'log-token-json-key': /"(\b|\B)[\w-]+"(?=\s*:)/gi,
+};
```
L2D2Grafana: Would be cool to share these somehow, maybe like `@gtk-grafana/react-json-tree`

src/services/highlight.tsx
```diff
@@ -0,0 +1,228 @@
...
+  // @todo find way to sync/pull from core?
+  const colors = {
+    critical: '#B877D9',
+    debug: '#6E9FFF',
+    error: theme.colors.error.text,
+    info: '#6CCF8E',
+    metadata: theme.colors.text.primary,
+    parsedField: theme.colors.text.primary,
+    trace: '#6ed0e0',
+    warning: '#FBAD37',
```
L2D2Grafana: I believe warning is `theme.colors.warning.text` but yeah not sure about critical etc

L2D2Grafana (commented):
This is really cool, I can already see everyone asking for this in the table!  Colors and fonts look really nice 🤩 . I'd love to find a way to make them reusable throughout all the different views.  

Some of my highlighting failed but I think it doesn't work in the logs view either
1. Cannot copy key value pairs like `msg: The flag is green` 
Screenshot 2025-07-07 at 3 45 04 PM
2. Cannot search for labels like eu-west-1 or trace id
Screenshot 2025-07-07 at 4 05 27 PM

src/Components/ServiceScene/LogListControls.tsx
matyax: Good call noticing this! In the new panel this button is used for highlighting, so I think it's okay 👍

src/Components/ServiceScene/JSONPanel/GetItemString.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { KeyPath } from '@gtk-grafana/react-json-tree/dist/types';
+
+interface GetItemStringProps {
+  data: unknown;
+  itemString: string;
+  itemType: React.ReactNode;
+  keyPath: KeyPath;
+  nodeType: string;
+}
+export default function GetItemString({ data, itemString, itemType, keyPath }: GetItemStringProps) {
```
matyax: Minor, but what's the purpose of the "Get" prefix? If it's just to render an ItemString, `<ItemString />` should be enough.

matyax:
There's a missing key somewhere

imagen

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
```diff
@@ -0,0 +1,251 @@
...
+  $data.useState();
+  const logsListScene = sceneGraph.getAncestor(model, LogsListScene);
+  const { visualizationType } = logsListScene.useState();
+  const styles = useStyles2((t) => getStyles(t, showHighlight));
+
+  const fieldsVar = getFieldsVariable(model);
+  const jsonVar = getJsonFieldsVariable(model);
+
+  // If we have a line format variable, we are drilled down into a nested node
+  const dataFrame = getLogsPanelFrame(data);
```
matyax, line 51: One big limitation that I see from this approach of highlighting based in variables is that you have zero certainty of what Loki is going to match; we're guessing in the frontend. The second is the added complexity of parsing line filters for highlighting.

To have matching behaviors between visualizations and highlight what Loki effectively matches, I suggest to use `searchWords` from the data frame, like logs model does. That, in combination with this function https://github.com/grafana/grafana/blob/main/packages/grafana-data/src/text/text.ts#L73 should give you the regex to match.

You can find the search words as meta information in the frame coming from the data source:

imagen

matyax:
I can't repro in play, so it seems that it comes from this branch. I killed the app by putting "[0m" in the line filter field. There seems to be a missing escapeRegex() call somewhere.

imagen

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
```diff
@@ -0,0 +1,251 @@
...
+    showLabels,
+    showMetadata,
+    sortOrder,
+  } = model.useState();
+  const $data = sceneGraph.getData(model);
+  // Rerender on data change
+  $data.useState();
+  const logsListScene = sceneGraph.getAncestor(model, LogsListScene);
+  const { visualizationType } = logsListScene.useState();
+  const styles = useStyles2((t) => getStyles(t, showHighlight));
```
matyax:
```suggestion
  const styles = useStyles2(getStyles, showHighlight);
```

https://github.com/grafana/grafana/blob/d22842b52b23d76f3a7da2d58ddb4884d42ae715/packages/grafana-ui/src/themes/ThemeContext.tsx#L117

.eslintrc
gtk-grafana: It's only used in the new unit tests in highlight.ts, this override should only target test files so I think this should be ok

src/services/highlight.tsx
gtk-grafana: Added warning in https://github.com/grafana/logs-drilldown/pull/1382/commits/efab1ac0ef7bfc3570dfe1b3cb3316c6810295cd

src/Components/ServiceScene/JSONPanel/GetItemString.tsx
gtk-grafana: Yeah I was just matching the name of the callback from the `JSONTree` package with the component name, but updated since the component isn't a getter

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
gtk-grafana: I'm not always seeing this present on the backend response. I'll dig in further in a minute, I would guess we're adding this to the frontend dataframe somehow?

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
gtk-grafana: Oh this is nice, I've got to remember this one! Fixed in https://github.com/grafana/logs-drilldown/pull/1382/commits/f84ba0ba5d8eb2f1788614a8e8a87283c425f8c8

src/services/highlight.tsx
gtk-grafana: Yeah we should schedule some time to discuss adding another package that can be consumed in core and in plugins. Adding a new package to core has advantages, but slow, adding a new disparate package is fast, but can't use it to externalize stuff in logs and explore.

gtk-grafana:
@matyax
* Unexpected error should be fixed in https://github.com/grafana/logs-drilldown/pull/1382/commits/af0946f6146a5a89bf2d6ef626a8d349e7288b56
* `There's a missing key somewhere`: I'm not seeing this locally, but I was yesterday. Not sure if it got fixed or if we need to do something special to repro?


@L2D2Grafana 

> Cannot copy key value pairs like msg: The flag is green

This is expected, you'd need to include all of the raw JSON formatting to match the actual log line, like `"msg":"The flag is green"`, although this won't highlight results in the JSON panel like it will the logs panel since we're executing regex against labels and values in isolation after parsing the JSON.

> Cannot search for labels like eu-west-1 or trace id

This is also expected, line filters only check the log line itself, not labels or metadata.

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
gtk-grafana: So yeah, this is being added to the frontend dataframe in backendResultTransformer by parsing the Loki expression with lezer from the frontend query expression (and not the executed result passed back to us from Loki)

matyax: Unfortunate. That's really no different from evaluating the filter variables. All good then, thanks for checking!

src/services/highlight.tsx
```diff
@@ -104,19 +110,22 @@ export const highlightValueStringMatches = (
...
     do {
       try {
-        valueMatch = regex.exec(value);
+        valueMatch = regex?.exec(value);
         if (valueMatch) {
           valueMatches.push(valueMatch);
         }
       } catch (e) {
-        logger.error(e, { msg: 'Error executing match expression', regex: regex.source });
+        logger.error(e, { msg: 'Error executing match expression', regex: regex?.source ?? '' });
```
gtk-grafana: these might be too noisy, maybe a warning or an info would be better?

matyax: Maybe only enable by a debug flag.

src/services/highlight.tsx
```diff
@@ -104,19 +110,22 @@ export const highlightValueStringMatches = (
 };
 
 // @todo cache results by regex/value?
```
gtk-grafana, line 98: There would be some advantages to caching the matches by value, especially now that we've added matching on the labels we could see the same regex getting executed against the same value for every node.

But we aren't checking against super long strings since we only execute against each label or value, so even worst case regex performance is going to be significantly faster then when testing against the whole log line, and cacheing could have a noticeable memory impact, if we need to store every unique label/value that is visualized (although labels shouldn't have too many uniques, also if we hashed the values that would save us from using long values as keys), and the resulting array of intervals (worst case behavior will contain a number for each char in the value (e.g. 'abcd': [[0,1], [2,3]), and floor(value.length /2) arrays).

But since RE2 runtime should scale linearly with the size of the input, as long as users are entering valid line filter regex (and we are validating user regex against re2 spec in the frontend, although currently users can still enter invalid regex that could explode the browser (although that wouldn't return results from Loki)), the runtime bonus might not be worth the memory trade-offs and garbage collection of keys every time the line filters change and we need to invalidate the cache.

I also imagine that horribly unoptimized regex that have a bunch of matches will also be inefficient on the Loki side, so even though we are potentially adding logLine.length/2 DOM nodes in the worst case that isn't a use-case we should worry about too much.

So let's leave it for now and iterate later if we start to notice performance problems? I'll do some more performance testing against some huge json to assert that this is not a concern.

src/services/highlight.tsx
gtk-grafana: Plugin feature flags would be lovely so we can have different logging levels for each env. I downgraded to an info for now.

gtk-grafana:
Found a bug with syntax highlighting, if a value contains a UUID the entire value is highlighted, not just the UUID

image

Also we should be skipping syntax highlighting for logs that are not JSON parsable.

matyax:
I saw that and assumed you wanted the whole line colored 😆

gtk-grafana:
@matyax fixed in https://github.com/grafana/logs-drilldown/pull/1382/commits/87f288073d6509aa8aae404ec815a617e7e8433b, I'm trying to match the results in the logs panel as close as I can, but I still don't match multiple syntax highlights like we match multiple line filter results.

I already spent ~45 minutes on it, it's gonna take a very clear head and some time, maybe follow up in another PR?

src/services/highlight.tsx
L2D2Grafana: How are you hitting large datasets, 005?

L2D2Grafana (approved):
Lgtm for an experimental feature 👏 Thanks for all the comments. Not sure I understand the real different, nuances between loki datasource search words vs using the filters variables.

src/Components/ServiceScene/JSONPanel/ItemString.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import { rootNodeItemString } from '../../../services/JSONViz';
+import { hasProp } from '../../../services/narrowing';
+import { JsonDataFrameTimeName, JsonVizRootName } from '../LogsJsonScene';
+import { KeyPath } from '@gtk-grafana/react-json-tree/dist/types';
+
+interface GetItemStringProps {
```
matyax:
```suggestion
interface ItemStringProps {
```

src/Components/ServiceScene/JSONPanel/ItemString.tsx
```diff
@@ -0,0 +1,28 @@
...
+  data: unknown;
+  itemString: string;
+  itemType: React.ReactNode;
+  keyPath: KeyPath;
+  nodeType: string;
+}
+export default function ItemString({ data, itemString, itemType, keyPath }: GetItemStringProps) {
```
matyax:
```suggestion
export default function ItemString({ data, itemString, itemType, keyPath }: ItemStringProps) {
```

src/Components/ServiceScene/JSONPanel/LabelRenderer.tsx
```diff
@@ -0,0 +1,97 @@
...
+    return <strong className={jsonLabelWrapStyles}>{LabelsDisplayName}</strong>;
+  }
+
+  if (keyPath[0] === JsonVizRootName) {
+    return model.renderNestedNodeButtons(keyPath, jsonFiltersSupported);
+  }
+
+  // Value nodes
+  if (
+    nodeTypeLoc !== 'Object' &&
```
matyax: This moved to a function would look beautiful (and thus easier to read). I found this unappealing on the first pass but failed to put words to it. No need to do it now, minor thing.

Same below.

src/Components/ServiceScene/JSONPanel/ValueRenderer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+  const value = valueAsString?.toString();
+  if (!value) {
+    return null;
+  }
+
+  const keyPathParent = keyPath[1];
+
+  if (
+    keyPathParent !== undefined &&
```
matyax: Same. Would be nice to put a name to this condition under a function.

src/Components/ServiceScene/JSONPanel/ValueRenderer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  keyPath: KeyPath;
+  lineFilters: AdHocFilterWithLabels[];
+  valueAsString: unknown;
+}
+
+export default function ValueRenderer({ keyPath, lineFilters, valueAsString }: ValueRendererProps) {
+  if (keyPath[0] === JsonDataFrameTimeName) {
+    return null;
+  }
+  const value = valueAsString?.toString();
```
matyax, line 21: Also minor. `valueAsString` and with type unknown seems like an oxymoron.

matyax: Alternatively:

```suggestion
interface ValueRendererProps {
  keyPath: KeyPath;
  lineFilters: AdHocFilterWithLabels[];
  value: unknown;
}

export default function ValueRenderer({ keyPath, lineFilters, value }: ValueRendererProps) {
  if (keyPath[0] === JsonDataFrameTimeName) {
    return null;
  }
  const valueAsString = value?.toString();
```

src/Components/ServiceScene/JSONPanel/ValueRenderer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  if (
+    keyPathParent !== undefined &&
+    keyPathParent !== JsonDataFrameStructuredMetadataName &&
+    keyPathParent !== JsonDataFrameLabelsName
+  ) {
+    let valueArray = highlightLineFilterMatches(lineFilters, value);
+
+    // If we have highlight matches we won't show syntax highlighting
+    if (valueArray.length) {
+      return valueArray.filter((e) => e);
```
matyax: Is this actually a:
```suggestion
      return valueArray.slice();
```
?

src/Components/ServiceScene/JSONPanel/highlightLineFilterMatches.ts
```diff
@@ -0,0 +1,33 @@
...
+  getMatchingIntervals,
+  highlightValueStringMatches,
+  mergeOverlapping,
+} from '../../../services/highlight';
+
+export function highlightLineFilterMatches(lineFilters: AdHocFilterWithLabels[], value: string, className?: string) {
+  const matchExpressions = getLineFilterRegExps(lineFilters);
+  const lineFilterMatches = getMatchingIntervals(matchExpressions, value);
+  const size = mergeOverlapping(lineFilterMatches);
+  let valueArray: Array<React.JSX.Element | string> = [];
```
matyax: I see a couple more of usages of this, maybe worth adding a type for `React.JSX.Element | string`

let highlightedResults: Array<string | React.JSX.Element> = [];
let value: string | Array<string | React.JSX.Element> = keyPath[0].toString();
let valueArray: Array<React.JSX.Element | string> = [];
export type HighlightedValue = Array<React.JSX.Element | string>;

src/Components/ServiceScene/LogListControls.tsx
```diff
@@ -71,12 +76,24 @@ export const LogListControls = ({
       {showLabels !== undefined && onToggleLabelsClick && (
         <IconButton
           name="key-skeleton-alt"
+          aria-pressed={showLabels}
```
matyax, line 79: <3

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -733,26 +532,38 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
     jsonFiltersSupported?: boolean
   ) => {
     const styles = useStyles2(getJSONVizValueLabelStyles);
     const value = this.getValue(keyPath, lineField.values)?.toString();
     const label = keyPath[0];
     const existingVariableType = this.getFilterVariableTypeFromPath(keyPath);
 
+    let highlightedValue: string | Array<string | React.JSX.Element> = [];
+    if (
+      keyPath[1] !== undefined &&
```
matyax: Another candidate for a function.

src/services/highlight.tsx
```diff
@@ -0,0 +1,228 @@
...
+    .filter(
+      (search) => (search.operator === LineFilterOp.match || search.operator === LineFilterOp.regex) && search.value
+    )
+    .map((search) => {
+      try {
+        if (search.key === 'caseSensitive') {
+          return new RegExp(search.value, 'g');
+        } else {
+          return new RegExp(search.value, 'gi');
+        }
```
matyax:
```suggestion
        return new RegExp(search.value, search.key === 'caseSensitive' ? 'g' : 'gi');
```

matyax:
> @matyax thoughts?

For simplicity, as a rule of thumb or acceptance criteria, I would say that if whatever this panel highlights matches what the current, old, Logs Panel highlights, that is acceptable. Even with inaccuracies, it's been doing this job for a long time and we rarely (not to say never) see complaints about it.

src/services/highlight.tsx
```diff
@@ -0,0 +1,228 @@
...
+  matchingIntervals: Array<[number, number]>,
+  value: string,
+  size: number,
+  className?: string
+) => {
+  let valueArray: Array<{ value: string } | string> = [];
+  let lineFilterMatchIndex = 0;
+  let matchInterval = matchingIntervals[lineFilterMatchIndex];
+
+  // @todo debug why it's only grabbing the first one
```
matyax: Why not delegate this to Prism? Probably treeshake could spare a few extra bytes and we don't need to maintain/test/extend it, and we get the same behavior as the new panel.

src/services/highlight.tsx
matyax: > But we aren't checking against super long strings since we only execute against each label or value

Famous last words.

In any case, we can fix when we encounter a bad example and not prematurely optimize without a good reason.

src/services/highlight.tsx
```diff
@@ -0,0 +1,228 @@
...
+  }
+  return 0;
+};
+
+export const getLogsHighlightStyles = (theme: GrafanaTheme2, showHighlight: boolean) => {
+  if (!showHighlight) {
+    return {};
+  }
+
+  // @todo find way to sync/pull from core?
```
matyax, line 173: We could have a chat at some point with any of the platform teams to see how we can approach this without introducing extra repos, at least for now.

src/services/logsSyntaxMatches.ts
```diff
@@ -0,0 +1,18 @@
+// Synced with grafana/grafana/public/app/features/logs/components/panel/grammar.ts - July 3, 2025
```
matyax: Pretty cool. You can also put a permalink to the current version from the "get permalink" button https://github.com/grafana/grafana/blob/ca730935733d86339177ed5d014b9343831df98b/public/app/features/logs/components/panel/grammar.ts

matyax:
One thing that we would have to eventually address is this, which I don't think it's doing anything particularly useful for anyone other than requiring extra clicks:

imagen

src/Components/ServiceScene/JSONPanel/ItemString.tsx
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1382/commits/68bbce5bb8873df41ec36c3e94682290a2fe16e9

src/Components/ServiceScene/JSONPanel/LabelRenderer.tsx
gtk-grafana: Cleaned up in https://github.com/grafana/logs-drilldown/pull/1382/commits/f1cf2985a61038140482666565b2ddd654d75532

src/Components/ServiceScene/JSONPanel/ValueRenderer.tsx
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1382/commits/c745633c5b3bb5958cb617bceca60de5e9f70cf3

src/Components/ServiceScene/JSONPanel/ValueRenderer.tsx
gtk-grafana: Good point, this might warrant some refactoring and type clean up in the base package, the `valueRenderer` callback returns a `valueAsString` and a `value` which are both unknown types, but the valueAsString is the value passed through a valueGetter method which attempts to return a string type, e.g. if the node type is "Date", the valueGetter calls `toISOString`.

TL;DR the type probably shouldn't be `unknown` in the package for valueAsString

gtk-grafana: See the [above](https://github.com/grafana/logs-drilldown/pull/1382#discussion_r2194861827), there's already a `value` returned from the `valueRenderer` callback, this should get cleaned up in the base package first or this might cause more confusion

gtk-grafana: commented in https://github.com/grafana/logs-drilldown/pull/1382/commits/4a7f0b64706f4cb027a86f0a5cbab803250ba346, let's figure out why this isn't just a string in the package first and then revisit this.

matyax: Sure thing.

src/Components/ServiceScene/JSONPanel/ValueRenderer.tsx
gtk-grafana: I'm not sure I understand, but I don't think this is necessary any more, I think this was a carry over from a previous implementation that could contain null or undefined values in the array we didn't want to render. Fixed in https://github.com/grafana/logs-drilldown/pull/1382/commits/c3d81cd52e20ec05400a25e6dd1cc1eb4c275ed9

src/Components/ServiceScene/JSONPanel/highlightLineFilterMatches.ts
gtk-grafana: https://github.com/grafana/logs-drilldown/pull/1382/commits/fe9417cec671427a8b953c407f0011333f88dac3

src/Components/ServiceScene/LogsJsonScene.tsx
gtk-grafana: We added one above, moved to JSONVizNodes service here:

https://github.com/grafana/logs-drilldown/pull/1382/commits/7d7361e300b3e547e09c644925b269e141167029
https://github.com/grafana/logs-drilldown/pull/1382/commits/f090f5ccb60d9f99e4df5d080c05cf38bbeda3c1

src/services/highlight.tsx
gtk-grafana: https://github.com/grafana/logs-drilldown/pull/1382/commits/41a72e297d6bd3bd2d0bf3ecf38ee6e16f8c03b7

src/services/highlight.tsx
gtk-grafana: Yeah we might want to, I'll follow up with a prism PoC and maybe timebox updating the current implementation to work with line filter matches and syntax highlighting in tandem and then we can make a more informed decision. 

although this comment is stale. Removed in https://github.com/grafana/logs-drilldown/pull/1382/commits/a1f2a41eb2ef8273eae85a69678abb6c1d1ebbfe

src/services/logsSyntaxMatches.ts
gtk-grafana: https://github.com/grafana/logs-drilldown/pull/1382/commits/8bd2c48887e8849ffd21f049a2a08558c29c86aa

gtk-grafana:
@matyax I think everything should be addressed now.

> One thing that we would have to eventually address is this, which I don't think it's doing anything particularly useful for anyone other than requiring extra clicks:

Not rendering every node is one of the reasons why this library is more performant and works well for arbitrary datasets, but we should discuss the tradeoffs of rendering fewer top-level nodes more deeply, or more shallow top-level nodes.

Virtualization might be better then this "pagination", but that's another discussion.

Issue #1387, opened 2025-07-08
[FEAT]: JSON - Toggle text wrap


gtk-grafana:
**Description of the new feature**
Allow users to toggle text wrap on/off like in the logs panel.

Issue #1385, opened 2025-07-08
[BUG]: JSON - time header stops being sticky after expanding additional section of logs


gtk-grafana:
**Describe the bug**
Time header is sticky until the user scrolls past the first collapsed section of log lines.

**To Reproduce**
Scroll down in JSON panel and open an additional page of log results.

<img width="1681" height="623" alt="Image" src="https://github.com/user-attachments/assets/178dd39f-b937-4c65-85a6-22e40ef343c6" />

After scrolling below these new logs after expanding, the time value is no longer sticky.

**Expected behavior**
Time header should remain sticky.

**Screenshots**
Sticky before opening additional section:

<img width="1727" height="278" alt="Image" src="https://github.com/user-attachments/assets/2f0781d9-53b7-48c2-8910-9e381bc60466" />

Not sticky after:

<img width="1728" height="222" alt="Image" src="https://github.com/user-attachments/assets/c049dac4-6d02-4269-ba8d-ceb2b668caf7" />

Issue #1380, opened 2025-07-03
[EPIC]: JSON panel - GA


gtk-grafana:
- [ ] https://github.com/grafana/logs-drilldown/issues/1377
- [ ] https://github.com/grafana/logs-drilldown/issues/1367
- [ ] https://github.com/grafana/logs-drilldown/issues/1378
- [ ] https://github.com/grafana/logs-drilldown/issues/1379
- [ ] https://github.com/grafana/logs-drilldown/issues/1369

Let's discuss with the squad next week and see if we need all of these for GA release.
Also we should discuss moving this panel outside of Logs Drilldown so it can be used in classic explore/dashboards.

PR #1388, opened 2025-07-08
Feat(JSON): Line wrap

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1387
Epic: https://github.com/grafana/logs-drilldown/issues/1380


src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
```diff
@@ -220,21 +243,22 @@ const getStyles = (theme: GrafanaTheme2, showHighlight: boolean) => {
...
         padding-left: ${theme.spacing(1)};
         align-items: center;
-        overflow-x: auto;
-        overflow-y: hidden;
       }
 
-      > ul > li > ul > li > span {
+      // sticky time header
+      > ul > li > ul > li > span,
+      > ul > li > ul > div > li > span {
```
gtk-grafana, line 275: fixes https://github.com/grafana/logs-drilldown/issues/1385

matyax:
Expanded a log line and got this:

imagen

matyax:
It's an interesting concept:

imagen

gtk-grafana:
@matyax talk about better use of horizontal screen space :P.

Looks like I forgot to add inline-grid style to those divs wrapping the nodes, should be an easy fix but the variable we set should already be targeting that node as well in the library. I'll take a look at making that update and bumping our version

gtk-grafana:
@matyax should be fixed in https://github.com/grafana/logs-drilldown/pull/1388/commits/dd664e12c3d59598362b21b9a871f6fdef04e3c6 (https://github.com/gtk-grafana/react-json-tree/pull/20)

matyax:
👌

src/Components/ServiceScene/LogListControls.tsx
```diff
@@ -63,6 +68,20 @@ export const LogListControls = ({
         tooltip={sortOrder === LogsSortOrder.Descending ? 'Newest logs first' : 'Oldest logs first'}
         size="lg"
       />
+      {wrapLogMessage !== undefined && onWrapLogMessageClick && (
+        <IconButton
+          name="wrap-text"
+          className={wrapLogMessage ? styles.controlButtonActive : styles.controlButton}
+          aria-pressed={wrapLogMessage}
+          onClick={() => onWrapLogMessageClick(!wrapLogMessage)}
```
L2D2Grafana, line 76: Am I trippin' is this backwards?
This looks wrapped to me, but the button is not highlighted and the text says 'wrap'
Screenshot 2025-07-09 at 12 29 50 PM
This looks unwrapped but the button is highlighted and the text say 'unwrap'
Screenshot 2025-07-09 at 12 30 03 PM

src/Components/ServiceScene/LogListControls.tsx
```diff
@@ -3,6 +3,7 @@ import React, { useCallback } from 'react';
 import { css } from '@emotion/css';
 
 import { GrafanaTheme2, LogsSortOrder } from '@grafana/data';
+import { t } from '@grafana/i18n';
```
L2D2Grafana, line 6: Ah we doing this now for all our text?

gtk-grafana: We should as a best effort for new strings, maybe we want to spike to get everything else translated and/or add eslint rules to require translation for copy that is modified/changed over time

src/Components/ServiceScene/LogListControls.tsx
gtk-grafana: Yup it's backwards. Also I noticed that while we are using the same local storage key as logs panel/explore, the logs panel in logs drilldown saves the URL state, so updates in JSON don't update the logs panel unless the user comes into the app fresh.

gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1388/commits/59c7231687e811ae6a093c961f0c4f63e95c0d8f

Url sync fixed in https://github.com/grafana/logs-drilldown/pull/1388/commits/f92d5e178e2574ee9a59879dd917831e4138eabb.

Also there was a bug where the visualization type was not url synced either (or working with JSON) which should be fixed now as well

gtk-grafana:
Fixed a few bugs, the docked grafana menu was causing the json panel to push the content outside of the page wrapper when text wrap was turned off, and the tab urls weren't updating when the viz type was set which was breaking an e2e test.

L2D2Grafana (approved):

PR #1401, opened 2025-07-10
JSON - Add reroot component to logs panel title items

gtk-grafana:
This is a few hacks inside of a trench coat, but every other solution I could come up with required moving the json breadcrumbs outside of the panels, which I liked even less.

JSON root navigation component inside the JSON viz:
image

JSON root navigation component inside the logs panel:
image

We should probably add it within the table as well, although setting a different root only impacts the logs table if the log text toggle is active and the line field is visible, otherwise the normal column/fields display is not impacted when a different root is selected.

### DIFF
--- a/src/Components/ServiceScene/LogsPanelScene.tsx
+++ b/src/Components/ServiceScene/LogsPanelScene.tsx
@@ -42,6 +42,7 @@ import { IndexScene } from '../IndexScene/IndexScene';
 import { getPanelWrapperStyles, PanelMenu } from '../Panels/PanelMenu';
 import { addToFilters, FilterType } from './Breakdowns/AddToFiltersButton';
 import { CopyLinkButton } from './CopyLinkButton';
+import JsonRootNodeNavigation from './JSONPanel/JsonRootNodeNavigation';
 import { LogOptionsScene } from './LogOptionsScene';
 import { LogsListScene } from './LogsListScene';
 import { LogsPanelError } from './LogsPanelError';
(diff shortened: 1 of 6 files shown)


matyax:
This is very interesting, in particular as it can be combined with displayed fields. 

imagen

The main challenge that I see is that, even though I like how it looks in the header, that will not scale with longer field names and is not responsive at all.

imagen

gtk-grafana:
@matyax I forgot that button could be there, we can workaround this a bit by increasing the width reserved for that button, which will at least keep this from pushing out of the container, but narrow viewports might not have any space left over to render this component :/

gtk-grafana:
I'm gonna put this on the back burner for now and come back to it

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -371,7 +372,20 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
...
+          sceneRef={this}
+          showEndingSeparator={false}
+          hideIfEmpty={true}
+          forceRender={true}
+          maxWidth={config.featureToggles.logsPanelControls ? `calc(100vw - 320px)` : `calc(100vw - 610px)`}
+        />
+      ),
+    });
+
+    return vizPanel;
```
L2D2Grafana, line 388: Do we want the navigation in the Logs Panel? It doesn't change the log line viewing. Would be cool to have it below the controls `Show Original Log Line`, `Newest First` etc.

Image

gtk-grafana: It would be cool, but PanelChrome doesn't support multiple rows of content, so it has to be inline 😢

gtk-grafana:
I'm gonna close this PR and re-implement after the refactor PR is merged.
Also I'm not sure if this isn't something we need for GA.

Issue #1396, opened 2025-07-10
[BUG]: JSON - Changing labelset after setting a nested node as root is not reversible


gtk-grafana:
**Describe the bug**
When setting a nested node as root in the JSON panel, and then changing the label to something that doesn't have that structure returns empty log lines, and there's no way to undo it in the UI without starting over fresh.

<img width="1719" height="581" alt="Image" src="https://github.com/user-attachments/assets/5aa22777-fadc-4935-b806-8c7f029e133c" />

**To Reproduce**
1. Set custom root node in JSON panel
2. Change labels to include logs with different JSON structure, or no JSON at all
3. Try to unset the custom root node set in 1.

**Expected behavior**
After changing labels the root node would reset.

PR #1402, opened 2025-07-10
fix(JSON): Allow changing root if no JSON logs present

gtk-grafana:
Partially fixes: https://github.com/grafana/logs-drilldown/issues/1396
This PR will at least prevent the bug that requires json logs to change the root node, but doesn't solve this problem if the user is not in the JSON panel when the labels are changed, that is being addressed in https://github.com/grafana/logs-drilldown/pull/1401.

This PR also refactors this component out for potential use in other contexts, and helps pay down some debt in the JSON panel.

### DIFF
--- a/src/Components/ServiceScene/JSONPanel/JsonRootNodeNavigation.tsx
+++ b/src/Components/ServiceScene/JSONPanel/JsonRootNodeNavigation.tsx
@@ -1,0 +1,175 @@
+import React from 'react';
+
+import { AdHocFilterWithLabels, SceneObject } from '@grafana/scenes';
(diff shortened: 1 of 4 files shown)


L2D2Grafana (approved):
I believe I successfully added filters and was able to remove the root node, in the JSON view 👍 



https://github.com/user-attachments/assets/6b63a236-d87a-4e87-9234-95eaf71b53bb

gtk-grafana:
@L2D2Grafana the bug is when you change the query to no longer return json (or return JSON that doesn't contain the new root node) when after setting a different JSON root, that ^ should work in main as well.

Issue #1394, opened 2025-07-09
[FEAT]: JSON - Share link to JSON log


gtk-grafana:
**Description of the new feature**
Provide way for users to share links to a specific json log.

PR #1406, opened 2025-07-11
feat(JSON): share link to line

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1394
Requires library updates: https://github.com/gtk-grafana/react-json-tree/pull/21, https://github.com/gtk-grafana/react-json-tree/pull/22

image

Working as expected, opening for review, but I'm going to add e2e coverage in the morning.

Need to add selected/hover styles, but I'll tackle that in another branch.

### DIFF
--- a/src/Components/Buttons/CopyToClipboardButton.tsx
+++ b/src/Components/Buttons/CopyToClipboardButton.tsx
@@ -7,17 +7,23 @@ import { t } from '@grafana/i18n';
 import { IconButton, InlineToast, useStyles2 } from '@grafana/ui';
 
 const SHOW_SUCCESS_DURATION = 2 * 1000;
+const COPY_TO_CLIPBOARD_TEXT = t('logs.log-line-details.copy-to-clipboard', 'Copy to clipboard');
+const COPY_LINK_TO_LINE_TEXT = t('logs.log-line-menu.copy-link', 'Copy link to log line');
+const COPY_LINK_ERROR_TEXT = t('logs.log-line-details.copy-to-clipboard-error', 'Error copying link!');
+const COPY_SUCCESS = t('clipboard-button.inline-toast.success', 'Copied');
 
 export default function CopyToClipboardButton({
   onClick,
   stopPropagation = true,
+  type = 'copy',
 }: {
   onClick: () => void;
   stopPropagation?: boolean;
+  type?: 'copy' | 'share-alt';
 }) {
+  const defaultText = type === 'copy' ? COPY_TO_CLIPBOARD_TEXT : COPY_LINK_TO_LINE_TEXT;
   const [copied, setCopied] = React.useState(false);
-  const copiedText = t('clipboard-button.inline-toast.success', 'Copied');
-  const defaultText = t('logs.log-line-details.copy-to-clipboard', 'Copy to clipboard');
+  const [copiedText, setCopiedText] = React.useState(COPY_SUCCESS);
   const buttonRef = useRef<null | HTMLButtonElement>(null);
   const styles = useStyles2(getStyles);
 
@@ -48,14 +54,19 @@ export default function CopyToClipboardButton({
         tooltip={copied ? '' : defaultText}
         tooltipPlacement="top"
         size="md"
-        name="copy"
+        name={type}
         ref={buttonRef}
         onClick={(e) => {
           if (stopPropagation) {
             // If the user clicked on the button, don't trigger the node to expand/collapse
             e.stopPropagation();
           }
-          onClick();
+          try {
+            onClick();
+            setCopiedText(COPY_SUCCESS);
+          } catch (e) {
+            setCopiedText(COPY_LINK_ERROR_TEXT);
+          }
           setCopied(true);
         }}
         tabIndex={0}
--- a/src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
+++ b/src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
@@ -6,7 +6,7 @@ import { FieldType, GrafanaTheme2 } from '@grafana/data';
 import { AdHocFilterWithLabels, SceneComponentProps, sceneGraph } from '@grafana/scenes';
 import { Alert, Badge, PanelChrome, useStyles2 } from '@grafana/ui';
 
-import { isLogLineField } from '../../../services/fields';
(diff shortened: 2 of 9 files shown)


src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -522,8 +523,30 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
   };
 
-  public renderCopyToClipboardButton(keyPath: KeyPath) {
-    return <CopyToClipboardButton onClick={() => copyLogLine(keyPath, sceneGraph.getData(this))} />;
+  private getLinkToLog(keyPath: KeyPath) {
+    const $data = sceneGraph.getData(this);
+    const timeRange = sceneGraph.getTimeRange(this).state.value;
+    const dataFrame = getLogsPanelFrame($data.state.data);
+    const idField: Field<string> | undefined = dataFrame?.fields.find((f) => isLogsIdField(f.name));
+    const logLineIndex = keyPath[0];
```
gtk-grafana, line 407: The json index might not necessarily by the index of the dataframe before sorting!

gtk-grafana: Should be fixed in main now

src/Components/Buttons/CopyToClipboardButton.tsx
```diff
@@ -7,17 +7,23 @@ import { t } from '@grafana/i18n';
 import { IconButton, InlineToast, useStyles2 } from '@grafana/ui';
 
 const SHOW_SUCCESS_DURATION = 2 * 1000;
+const COPY_TO_CLIPBOARD_TEXT = t('logs.log-line-details.copy-to-clipboard', 'Copy to clipboard');
```
matyax, line 10: <3 If we do this in every file that we edit, we should organically i18n the app relatively quickly.

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
```diff
@@ -62,6 +64,12 @@ export default function LogsJsonComponent({ model }: SceneComponentProps<LogsJso
   const jsonParserPropsMap = new Map<string, AdHocFilterWithLabels>();
   const lineFilterVar = getLineFiltersVariable(model);
 
+  const idField = rawFrame?.fields.find((field) => isLogsIdField(field.name));
+  const lineIndex = idField?.values.findIndex((v) => v === selectedLine?.id);
+  const cleanLineIndex = lineIndex !== undefined && lineIndex !== -1 ? lineIndex : undefined;
+  const scrollToPath: ScrollToPath | undefined =
+    cleanLineIndex !== undefined ? [cleanLineIndex, JsonVizRootName] : undefined;
```
matyax:
```suggestion
  const scrollToPath: ScrollToPath | undefined = useMemo(() => {
    if (selectedLine === undefined) {
      return undefined;
    }
    const idField = rawFrame?.fields.find((field) => isLogsIdField(field.name));
    const lineIndex = idField?.values.findIndex((v) => v === selectedLine?.id);
    const cleanLineIndex = lineIndex !== undefined && lineIndex !== -1 ? lineIndex : undefined;
    return cleanLineIndex !== undefined ? [cleanLineIndex, JsonVizRootName] : undefined;
  }, [selectedLine, rawFrame, ...etc]);
```

If you do this you would do less work on every render and skip NOOPs when there is no selectedLine

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
```diff
@@ -243,11 +252,16 @@ const getStyles = (theme: GrafanaTheme2, wrapLogMessage: boolean) => {
       --json-tree-ul-root-padding: ${theme.spacing(3)} 0 ${theme.spacing(2)} 0;
       --json-tree-arrow-left-offset: -${theme.spacing(2)};
       --json-tree-inline: inline-grid;
+      // Scroll offset for sticky header
+      --json-tree-scroll-margin: 26px;
+      // Scroll behavior @todo set "auto" instead of "smooth" for users with prefers-reduced-motion
+      scroll-behavior: smooth;
```
matyax:
```suggestion
      scroll-behavior: ${window.matchMedia('(prefers-reduced-motion: reduce)').matches ? 'auto' : 'smooth'};
```

I'm not a huge fan of @todo comments, as they usually seem to be forgotten.

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -399,8 +400,28 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
+    const dataFrame = this.state.rawFrame;
+    const idField: Field<string> | undefined = dataFrame?.fields.find((f) => isLogsIdField(f.name));
+    const logLineIndex = keyPath[0];
+    if (!isNumber(logLineIndex)) {
+      const error = Error('Invalid line index');
+      logger.error(error, { msg: 'Error getting log line index' });
+      throw error;
+    }
+    const logId = idField?.values[logLineIndex];
+    const logLineLink = generateLogShortlink('selectedLine', { id: logId, row: logLineIndex }, timeRange);
```
matyax, line 414: A reminder that this matches the behavior that the table introduced but not what is used by the visualization, supported in Dashboards and Explore, using `panelState`

matyax (approved):
A few suggestions, but looks great otherwise!

src/Components/ServiceScene/JSONPanel/LogsJsonComponent.tsx
gtk-grafana: Good catch, fixed in https://github.com/grafana/logs-drilldown/pull/1406/commits/f216a80da6e225ff0ff9b21336d236101ec10622

src/Components/ServiceScene/LogsJsonScene.tsx
gtk-grafana: Still waking up here, is that a good thing or no? 😄

matyax: No, I think all the viz should use panelState, no only to be consistent between them but also DD/Dashboards/Explore. I thought we had an issue for that, but I couldn't find it with a quick search.

PR #1407, opened 2025-07-11
feat(JSON): Add detected_level button

gtk-grafana:
Add log level to line ItemString.
Also fixes some bugs introduced by the dataframe sort.

image

### DIFF
--- a/src/services/extensions/links.test.ts
+++ b/src/services/extensions/links.test.ts
@@ -9,6 +9,22 @@ import { addAdHocFilterUserInputPrefix, EMPTY_VARIABLE_VALUE } from '../variable
 import { LinkConfigs, linkConfigs } from './links';
 import { addCustomInputPrefixAndValueLabels, encodeFilter, getPath } from './utils';
 
+// Mocking templateSrv is such a pain, if you are fighting variable interpolation start here.
+jest.mock('@grafana/runtime', () => ({
+  ...jest.requireActual('@grafana/runtime'),
+  getTemplateSrv: jest.fn(() => ({
+    replace: jest.fn((a: string, ...rest: unknown[]) => {
+      if (a === '${ds}') {
+        return '123abc';
+      }
+      if (a.includes('$adhoc')) {
+        return a.replace('$adhoc', 'cluster="eu-west-1"');
+      }
+      return a;
+    }),
+  })),
+}));
+
 function getTestConfig(
   links: LinkConfigs,
   target: Partial<LokiQuery> & { refId: string },
@@ -18,8 +34,9 @@ function getTestConfig(
     dashboard: {
       tags: [],
       title: 'test',
-      uid: 'test',
+      uid: '${ds}',
     },
+    scopedVars: { ds: { value: 'test', text: 'test' } },
     id: 0,
     pluginId: 'grafana-lokiexplore-app',
     timeRange: {
@@ -37,9 +54,9 @@ function getTestTarget(lokiQuery?: Partial<LokiQuery>): Partial<LokiQuery> & { r
   return {
     datasource: {
       type: 'loki',
-      uid: '123abc',
+      uid: '${ds}',
     },
-    expr: '{cluster="eu-west-1"} |= "\\\\n" ',
+    expr: '{$adhoc} |= "\\\\n" ',
     ...lokiQuery,
     refId: lokiQuery?.refId ?? 'A', // Ensure refId is defined
   };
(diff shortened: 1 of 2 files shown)


src/Components/ServiceScene/JSONPanel/ItemString.tsx
```diff
@@ -26,5 +36,43 @@ export default function ItemString({ data, itemString, itemType, keyPath, model
     );
   }
 
+  if (keyPath[0] === JsonDataFrameLineName) {
+    const labelsField: Field<Labels> | undefined = model.state.rawFrame?.fields.find(
```
gtk-grafana: refactor into component

matyax:
I'm still seeing 
imagen

matyax:
I love how it looks, but regarding the filtering behavior, but I don't think the behavior is straighforward to understand and discover as it is, besides the cursor: pointer. Maybe add a hover state and a tooltip?

In parallel, this looks the same as the other, but is not interactable, which I understand, but a random user maybe wouldn't.

imagen

gtk-grafana:
@matyax I'm still not seeing the unique key issue locally, is there something specific that you can do to trigger this error or is it always showing for you? 

I added a basic hover state and tooltips to that button.

> In parallel, this looks the same as the other, but is not interactable, which I understand, but a random user maybe wouldn't.

Ahh, yeah this makes it look like all syntax highlighted text should be filterable.
Should we add a border and/or differentiate the style here a bit? I was trying to have parity with the Logs Panel, but the levels in the logs panel aren't filterable like this.
image

matyax:
> @matyax I'm still not seeing the unique key issue locally, is there something specific that you can do to trigger this error or is it always showing for you?

No, by just going to the panel. It reproduces consistently when you go to JSON for the first time in a session or if you refresh when in JSON mode.

https://github.com/user-attachments/assets/43818e49-01c9-4160-9e9a-937a071eaf7c

> I added a basic hover state and tooltips to that button.

Cool. I think it should be enough with this for now, until we receive further feedback.

gtk-grafana:
Oh I already added the border, but I'm still not seeing that key warning...

matyax:
In this or another PR we could use at least 2 more px of margin below the header:

imagen

src/Components/ServiceScene/JSONPanel/JsonLineItemType.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { t } from '@grafana/i18n';
+import { AdHocFiltersVariable, SceneObject } from '@grafana/scenes';
+import { Tooltip, useStyles2 } from '@grafana/ui';
+
+import { FilterOp } from '../../../services/filterTypes';
+import { logsLabelLevelsMatches } from '../../../services/panel';
+import { LEVEL_VARIABLE_VALUE, VAR_LEVELS } from '../../../services/variables';
+import { addToFilters } from '../Breakdowns/AddToFiltersButton';
+
+export function JsonLineItemType({
```
matyax: Can you memo this component? Tooltips tend to add a lot of extra re-renders.

matyax (approved):
Looks great to me!

src/Components/ServiceScene/JSONPanel/JsonLineItemType.tsx
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1407/commits/f6fe977d668edc75c35f876c10e100455595d4c5

src/services/panel.ts
```diff
@@ -41,6 +41,15 @@ export const ERROR_LEVEL_FIELD_NAME_REGEX = /^error$/i;
...
 
+export const logsLabelLevelsMatches: Record<string, RegExp> = {
+  'log-token-info': INFO_LEVEL_FIELD_NAME_REGEX,
+  'log-token-debug': DEBUG_LEVEL_FIELD_NAME_REGEX,
+  'log-token-warning': WARNING_LEVEL_FIELD_NAME_REGEX,
+  'log-token-error': ERROR_LEVEL_FIELD_NAME_REGEX,
+  'log-token-critical': CRITICAL_LEVEL_FIELD_NAME_REGEX,
+  'log-token-unknown': UNKNOWN_LEVEL_FIELD_NAME_REGEX,
+};
+
```
L2D2Grafana, line 52: These look like they could potentially be shared with core.

tests/exploreServicesJsonBreakDown.spec.ts
```diff
@@ -205,6 +205,7 @@ test.describe('explore nginx-json breakdown pages ', () => {
     test('can filter nested props and drill back to root without removing json parser prop for active filters', async ({
       page,
     }) => {
+      //also
```
L2D2Grafana: also 🤔

gtk-grafana:
@matyax 

> In this or another PR we could use at least 2 more px of margin below the header:

In another PR, right now all of the nodes/lines have the same height, things look a bit funny if we change just this one.

L2D2Grafana (approved):
All the colors 🖌️ 👏 . I agree that the tooltip is a bit strange to add to filter, should we put the canonical +/- 🔍 in front of it?

Image

gtk-grafana:
@L2D2Grafana 

> All the colors 🖌️ 👏 . I agree that the tooltip is a bit strange to add to filter, should we put the canonical +/- 🔍 in front of it?

I'd rather not as this isn't the detected_level field, we're just cramming something into the line header and as such the icons won't line up with the others. +/- is available in the level if metadata is showing.
If this gets to be too confusing we can just pull the filtering functionality and have this text be for display purposes only.

gtk-grafana:
I'm gonna merge so we get the e2e/sorting fixes, we can make further tweaks in a follow-up PR.

PR #1410, opened 2025-07-14
fix(LogsPanel): visible range and panel improvements

matyax:
- No longer passing `displayMode` so it uses the default value or the stored preference (requires https://github.com/grafana/grafana/pull/108065)
- Fixed visible range when Logs Volume was collapsed
- Removed internal data override when new logs are received from infinite scrolling, which was causing a scroll jump in the new panel

Besides these changes, the current behavior should match the behavior in main minus the bugs.

### Before

https://github.com/user-attachments/assets/a7eb9501-8f47-4f96-90f0-f0bc70d3df1d

### After

https://github.com/user-attachments/assets/cd1a0273-f50d-4a81-8f4e-c2dc3c914bfb


L2D2Grafana (approved):
👍 Very nice thank you for the videos and the link to core!

https://github.com/user-attachments/assets/8c008fb7-b67e-4cd3-95bf-118fa4313e3e

PR #1411, opened 2025-07-14
fix(links): interpolate expression and datasource variable

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1409

Currently datasource and other template variables in the logQL expression are not being interpolated when building the link to Logs Drilldown. This causes errors in Logs Drilldown when linking to Logs Drilldown from a panel that utilizes a datasource variable, and breaks the link extension (the extension will not display for otherwise valid logQL expressions) for expressions containing template variables.

This PR simply interpolates those when building the links.


src/services/extensions/links.ts
```diff
@@ -165,11 +165,14 @@ function contextToLink<T extends PluginExtensionPanelContext>(context?: T) {
     return undefined;
   }
   const lokiQuery = context.targets.find((target) => target.datasource?.type === 'loki') as LokiQuery | undefined;
-  if (!lokiQuery || !lokiQuery.datasource?.uid) {
+  const templateSrv = getTemplateSrv();
+  const dataSourceUid = templateSrv.replace(lokiQuery?.datasource?.uid, context.scopedVars);
```
matyax, line 169: Is this a thing?

PR #1415, opened 2025-07-15
fix(JSON): tooltip copy not updating

gtk-grafana:
We memoized this component in #1407 but this broke the tooltip state. Refactoring to pass in filters array instead of variable to bust cache on change.

### DIFF
--- a/src/Components/ServiceScene/JSONPanel/JsonLineItemType.tsx
+++ b/src/Components/ServiceScene/JSONPanel/JsonLineItemType.tsx
@@ -1,10 +1,10 @@
-import React, { memo } from 'react';
+import React, { memo, useState } from 'react';
 
 import { css } from '@emotion/css';
 
 import { GrafanaTheme2 } from '@grafana/data';
 import { t } from '@grafana/i18n';
-import { AdHocFiltersVariable, SceneObject } from '@grafana/scenes';
+import { AdHocFiltersVariable, AdHocFilterWithLabels, SceneObject } from '@grafana/scenes';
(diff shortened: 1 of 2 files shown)


src/Components/ServiceScene/JSONPanel/JsonLineItemType.tsx
```diff
@@ -15,25 +15,25 @@ import { addToFilters } from '../Breakdowns/AddToFiltersButton';
...
 
   return (
     <Tooltip
       content={t(
         'logs.json.line.detectedLevel.toggleButton',
-        existingLevel ? `Remove ${detectedLevel} filter` : `Include logs with ${detectedLevel} level`
+        existingFilter ? `Remove ${detectedLevel} filter` : `Include logs with ${detectedLevel} level`
       )}
     >
       <button
```
matyax, line 39: Unrelated, and this is as far below as I can add a comment, but I think default exports are generally discouraged.

gtk-grafana: oh lame, it solves the annoying named exports warning and is a nice way to ensure that files don't get too bloated.
I can keep this in mind going forward but do we have any docs defining why?

matyax: https://github.com/grafana/grafana/blob/main/contribute/style-guides/frontend.md#exports

There are some reasons about autocompletion and stuff if you google. Not a huge problem anyway.

PR #1418, opened 2025-07-15
feat(JSON): hover & selected styles

gtk-grafana:
image
image
image
image


matyax:
Edit: I don't know why I thought I was always seeing the same color.

Issue #1419, opened 2025-07-15
[FEAT]: JSON - Links


gtk-grafana:
Add derived field links to logs JSON panel

PR #1420, opened 2025-07-15
feat(JSON): links

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1419

image

### DIFF
--- a/provisioning/datasources/default.yaml
+++ b/provisioning/datasources/default.yaml
@@ -9,6 +9,25 @@ datasources:
     uid: gdev-loki
     access: proxy
     url: http://host.docker.internal:3100
+    jsonData:
+      derivedFields:
+        - datasourceUid: gdev-jaeger
+          matcherRegex: traceID
+          matcherType: label
+          name: traceID-jaeger
+          url: '$${__value.raw}'
+          urlDisplayLabel: jaeger
+        - datasourceUid: gdev-zipkin
+          matcherRegex: traceID
+          matcherType: label
+          name: traceID-zipkin
+          url: '$${__value.raw}'
+        - datasourceUid: gdev-tempo
+          matcherRegex: traceID
+          matcherType: label
+          name: traceID-tempo
+          url: '$${__value.raw}'
+          urlDisplayLabel: tempo
   - name: gdev-tempo
     type: tempo
     uid: gdev-tempo
--- a/src/Components/ServiceScene/LogsJsonScene.tsx
+++ b/src/Components/ServiceScene/LogsJsonScene.tsx
@@ -119,6 +119,9 @@ export type AddMetadataFilter = (
   variableType: InterpolatedFilterType
 ) => void;
 
+type ParsedJsonLogLineValue = string | string[] | Record<string, string> | Array<Record<string, string>>;
+type ParsedJsonLogLine = Record<string, ParsedJsonLogLineValue> | Array<Record<string, string>>;
+
 export const JsonDataFrameTimeName = 'Time';
 export const JsonDataFrameLineName = 'Line';
 export const StructuredMetadataDisplayName = 'Metadata';
@@ -647,9 +650,7 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
                           }
                         });
                       }
-                      const line:
-                        | Record<string, string | string[] | Record<string, string> | Array<Record<string, string>>>
-                        | Array<Record<string, string>> = {
+                      const line: ParsedJsonLogLine = {
                         [JsonDataFrameLineName]: parsed,
                         [JsonDataFrameTimeName]: renderJSONVizTimeStamp(time?.values?.[i], timeZone),
                       };
@@ -670,7 +671,7 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
                     .filter((f) => f),
                 };
               }
-              return { ...field };
+              return field;
             }),
           };
         }),
--- a/src/Components/ServiceScene/LogsJsonScene.tsx
+++ b/src/Components/ServiceScene/LogsJsonScene.tsx
@@ -92,6 +92,7 @@ import {
   getLogOption,
   setLogOption,
 } from 'services/store';
+import { getJsonDerivedFieldsLinks } from '../../services/derivedFields';
 
 interface LogsJsonSceneState extends SceneObjectState {
   data?: PanelData;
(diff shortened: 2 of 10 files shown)


gtk-grafana:
Text links might fit the JSON UI better, instead of trying to match the buttons in the Logs Panel

gtk-grafana:
Never bothered to remove filters for links 🤦 . Back to draft for now

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -567,6 +463,113 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
     );
   };
 
+  private setStateFromUrl() {
```
gtk-grafana, line 471: ugh ****ing whitespace

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -638,7 +647,9 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
                           }
                         });
                       }
-                      const line: Record<string, Record<string, string> | string> = {
+                      const line:
+                        | Record<string, string | string[] | Record<string, string> | Array<Record<string, string>>>
```
gtk-grafana: clean this up

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -648,13 +659,18 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
+                          line[JsonDataFrameLinksName] = jsonLinks;
+                        }
+                      }
                       return line;
                     })
                     .filter((f) => f),
                 };
               }
-              return f;
+              return { ...field };
```
gtk-grafana: ??

tests/exploreServicesJsonBreakDown.spec.ts
```diff
@@ -385,11 +385,35 @@ test.describe('explore nginx-json breakdown pages ', () => {
...
+        .first();
+      const traceId = (await traceIdLoc.textContent())?.split(':')?.[1];
+      if (!traceId) {
+        throw new Error('No trace ID found.');
+      }
+
+      const href = await page.getByRole('link', { name: 'traceID-tempo' }).first().getAttribute('href');
+      expect(href).toContain(traceId);
+
+      // Tempo occasionally throws an error breaking the entire query editor for specific UUIDs, but works most of the time.
```
gtk-grafana: we don't really need to assert that the links that are generated by core actually work, just that Logs Drilldown is associating the traceID on the log with the link, so let's remove this commented out crap

src/Components/ServiceScene/LogsJsonScene.tsx
gtk-grafana: I think `yarn lint:fix` causes damage, I think I ran it when I was trying to debug an error with eslint complaining about the yaml file updates. 

I'd vote for removing it and replacing with whatever we run in the precommit

provisioning/datasources/default.yaml
```diff
@@ -9,6 +9,25 @@ datasources:
...
+          matcherRegex: traceID
+          matcherType: label
+          name: traceID-zipkin
+          url: '$${__value.raw}'
+        - datasourceUid: gdev-tempo
+          matcherRegex: traceID
+          matcherType: label
+          name: traceID-tempo
+          url: '$${__value.raw}'
+          urlDisplayLabel: tempo
```
L2D2Grafana, line 30: Didn't know about this, cool! These all might be related to the nature of the fake data

I aways think of traceIds opening with tempo or in Traces Drilldown. This one opened as a random walk in explore
Image

This one opened in Traces explore but had an error, maybe expected?
Image

src/Components/ServiceScene/LogsJsonScene.tsx
L2D2Grafana: precommit is running lint-staged.config.js which runs prettier and eslint --fix which is very similar to what's in yarn lint:fix. I also install the prettier plugin in vsCode and it fixes everything on save. Does your IDE have that?

src/Components/ServiceScene/LogsJsonScene.tsx
```diff
@@ -599,21 +605,27 @@ export class LogsJsonScene extends SceneObjectBase<LogsJsonSceneState> {
...
+    const templateSrv = getTemplateSrv();
+    const replace = templateSrv.replace.bind(templateSrv);
+
     const timeZone = getTimeZone();
     if (dataFrame && newState.data) {
       const isRerooted = getLineFormatVariable(this).state.filters.length > 0;
+      const derivedFields: Field[] =
+        dataFrame?.fields
+          .filter((f) => f.config.links)
+          .map((field) => ({ ...field, getLinks: getLinksSupplier(dataFrame, field, {}, replace) })) ?? [];
```
L2D2Grafana, line 617: Are the derivedFields template variables? haven't seen this before getTemplateSrv

provisioning/datasources/default.yaml
gtk-grafana: For the tempo one: https://raintank-corp.slack.com/archives/C06M1RVAA0Z/p1752685814474789.
And yeah, the links won't generate correctly if the associated datasources don't exist, and we're not spinning up a jaeger or zipkin, but core generates the href so it's not something we should need to write test coverage for in the plugin, but if we set up a zipkin/jaeger datasource with those UIDs it should work.

Creating multiple mostly so we can assert that the link names and other stuff gets pulled in correctly.
Also when running server:ci the tempo isn't spun up so e2e tests need to be agnostic to the datasource getting created.

src/Components/ServiceScene/LogsJsonScene.tsx
gtk-grafana: The table does this as well in `/src/Components/Table/LogLinePill.tsx`, I think it's just in case the links contain template variables

src/Components/ServiceScene/LogsJsonScene.tsx
gtk-grafana: yeah I run prettier on save in the IDE, but I think `yarn lint:fix` rearranged my methods which prettier doesn't do afaik

L2D2Grafana (approved):
Sounds like the errors I was wondering about were just the nature of the fake data.

PR #1421, opened 2025-07-16
fix(LogsPanelScene): pass setDisplayedFields

matyax:
Follow up to https://github.com/grafana/grafana/pull/107635, where along with `onClickShowField` and `onClickHideField`, we also pass the callback to reorder fields. Otherwise, the internal default one is used and the application becomes out of sync with the visualization.


gtk-grafana (approved):
LGTM

Issue #1399, opened 2025-07-10
[FEAT]: Table updates for logs control panel


L2D2Grafana:
- Show labels/show raw log line button might need a disabled state if the Log Line column is not visualized.
<img width="90" height="190" alt="Image" src="https://github.com/user-attachments/assets/c1f2ed2b-5ddf-467b-ae67-7eee76bafa52" />

- Column width estimator isn't taking the new icons into effect, so we're seeing some columns with titles that are longer then the values push the icons out of the viewport:
<img width="556" height="596" alt="Image" src="https://github.com/user-attachments/assets/1a1c2a76-4d9b-4245-9049-d14519be41bc" />

PR #1422, opened 2025-07-16
feat(table): column width estimator and toggle logsPanelControls

L2D2Grafana:
- #1399  

- Disable the labels/show raw log line button if the Log Line column is not visualized.

https://github.com/user-attachments/assets/208be83d-5849-447e-8728-a1978aca7d7e

- Updated the Column width estimator isn't taking the new icons into effect. Also removed columns from the columnWidthMap on hide. Not sure we want that change, but it took me a while to realize that was skipping the `getInitialFieldWidth` function.
Screenshot 2025-07-16 at 12 01 35 PM

### DIFF
--- a/src/Components/ServiceScene/LogsTableScene.tsx
+++ b/src/Components/ServiceScene/LogsTableScene.tsx
@@ -30,6 +30,7 @@ import { LogListControls } from './LogListControls';
 import { LogsListScene } from './LogsListScene';
 import { getLogsPanelFrame } from './ServiceScene';
 import { logger } from 'services/logger';
+import { DATAPLANE_BODY_NAME_LEGACY, DATAPLANE_LINE_NAME } from 'services/logsFrame';
 import { narrowLogsSortOrder, unknownToStrings } from 'services/narrowing';
 import { logsControlsSupported } from 'services/panel';
 
@@ -37,8 +38,8 @@ let defaultUrlColumns = DEFAULT_URL_COLUMNS;
 
 interface LogsTableSceneState extends SceneObjectState {
   emptyScene?: NoMatchingLabelsScene;
+  isDisabledLineState: boolean;
   menu?: PanelMenu;
-  showLineState: boolean;
   sortOrder: LogsSortOrder;
 }
 export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
@@ -50,7 +51,7 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
     super({
       ...state,
       sortOrder: getLogOption<LogsSortOrder>('sortOrder', LogsSortOrder.Descending),
-      showLineState: false,
+      isDisabledLineState: false,
     });
 
     this.addActivationHandler(this.onActivate.bind(this));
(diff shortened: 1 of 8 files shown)


matyax:
With these changes I no longer see the log line control in the controls. 

imagen
 

Also I think we should update the labels icon from the one used by syntax highlitgting to maybe this

 imagen

What do you think @gtk-grafana ?

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -65,6 +67,7 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
   getUrlState() {
     return {
       sortOrder: JSON.stringify(this.state.sortOrder),
+      body: this.state.showLineState ? 'true' : 'false',
```
gtk-grafana: body is not a url sync param?

gtk-grafana:
@matyax 

> With these changes I no longer see the log line control in the controls.

I think that is on purpose because the log line might not be a visualized column? It can still be toggled in the line column header

> Also I think we should update the labels icon from the one used by syntax highlitgting to maybe this

Yeah, switching out the Icon is a good call now that we're using that for highlighting in the other panels. 

@L2D2Grafana 
LGTM, only comment is that the Line field should have a higher min-width as more columns are added, but I think it only appears worse on this branch vs main is because the extra space dedicated to other columns, so no need to block this PR with that.

matyax:
> I think that is on purpose because the log line might not be a visualized column?

But it was, unless I'm still under the effects of the anesthesia. The "Line" was a checked column.

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -89,18 +92,46 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
...
+  };
+
   // on activate sync displayed fields with url columns
   onActivateSyncDisplayedFieldsWithUrlColumns = () => {
     const searchParams = new URLSearchParams(locationService.getLocation().search);
     let urlColumnsUrl: string[] | null = [];
     try {
       urlColumnsUrl = unknownToStrings(JSON.parse(decodeURIComponent(searchParams.get('urlColumns') ?? '')));
+      // If body or line is in the url columns, show the line state controls
+      if (urlColumnsUrl.includes('body') || urlColumnsUrl.includes('line')) {
```
gtk-grafana: use `DATAPLANE_BODY_NAME_LEGACY` and `DATAPLANE_LINE_NAME` plz

gtk-grafana:
@matyax 

> But it was, unless I'm still under the effects of the anesthesia. The "Line" was a checked column.

Yeah I'm saying that since folks can already do this in the column header it makes more sense to do that action there then have it in the options on the side which implies that the toggle does something to the entire table.

matyax:
Ohh. I don't know. We had a discoverability issue with that, so I really expected it to be a global control.

gtk-grafana:
That is true, lots of folks didn't know how to see the actual log line. Yeah, maybe we should conditionally add the back in for now

L2D2Grafana:
Okay sounds like I was over zealous with removing the button, I've changed the log line control to disabled and updated the description with a video.  Also I updated the label icons to use `alt-tag`, an excellent idea👍

matyax:
> Okay sounds like I was over zealous with removing the button,

No worries! The background is that it was a recurring comment "the table is amazing but I wish there was a way to see the actual log line", so people were not finding the header option. Similar to the manage columns situation, which is now permanently visible ✨

PR #1426, opened 2025-07-17
feat(assistant): add `Explain in Assistant` panel option

svennergr:
https://github.com/user-attachments/assets/4442b423-104d-4bae-b6c3-98e9ebc787c6


src/Components/Panels/PanelMenu.tsx
```diff
@@ -139,6 +140,19 @@ export class PanelMenu extends SceneObjectBase<PanelMenuState> implements VizPan
         }),
       });
```
matyax, line 142: Until we add it as a core part of the panel, you can also add it as a custom option https://github.com/grafana/logs-drilldown/blob/main/src/Components/ServiceScene/LogsPanelScene.tsx#L366

L2D2Grafana:
I ran this locally in Logs Drilldown and Logs Drilldown inside Grafana locally but couldn't see it. Is there something else I need to configure?
Screenshot 2025-07-18 at 9 29 04 AM

svennergr:
> I ran this locally in Logs Drilldown and Logs Drilldown inside Grafana locally but couldn't see it. Is there something else I need to configure? Screenshot 2025-07-18 at 9 29 04 AM

I think you tested the case when Grafana Assistant is not installed - so that's good the button is not rendering 😆 
You would need to install Grafana Assistant following https://github.com/grafana/grafana-assistant-app?tab=readme-ov-file#development and https://github.com/grafana/grafana-assistant-app?tab=readme-ov-file#local-grafana-setup

gtk-grafana (approved):
code looks good, I don't have assistant installed but I'll assume that you've checked that this works with logs volume, logs/table/json panels, labels/fields tabs (rollup and value).

L2D2Grafana (approved):
I could not click the button but I saw it, ty! 
Image

Issue #1400, opened 2025-07-10
[BUG]: Table adding filters multiple times creates duplicates


L2D2Grafana:
**Describe the bug**
Table adding filters that already exist from the table does not actually toggle, but keeps adding duplicate filters.

**To Reproduce**
1. Go to table
2. Add a label to a column
3. -/+ to add label to filter

**Expected behavior**
Toggle the filter

**Screenshots**
<img width="1409" height="367" alt="Image" src="https://github.com/user-attachments/assets/fa5e5407-5c31-4bf3-89f1-9492223fe484" />

PR #1430, opened 2025-07-17
fix(ad-hoc filters): fix duplicate filters

L2D2Grafana:
- #1400

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
+++ b/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
@@ -218,6 +218,17 @@ export function addToFilters(
 
   const variable = getUIAdHocVariable(variableType, key, scene);
 
+  // Check if the filter already exists
+  const operatorSymbol = operator === 'include' ? FilterOp.Equal : FilterOp.NotEqual;
+  const isDuplicate = variable.state.filters.some(
+    (filter) => filter.key === key && filter.value === value && filter.operator === operatorSymbol
+  );
+
+  // Return early do not add duplicate filters
+  if (isDuplicate) {
+    return;
+  }
+
   let valueObject: string | undefined = undefined;
   let valueLabel = value;
   if (variableType === VAR_FIELDS) {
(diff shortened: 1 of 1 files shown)


src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx
```diff
@@ -218,6 +218,17 @@ export function addToFilters(
...
 
+  // Check if the filter already exists
+  const operatorSymbol = operator === 'include' ? FilterOp.Equal : FilterOp.NotEqual;
+  const isDuplicate = variable.state.filters.some(
+    (filter) => filter.key === key && filter.value === value && filter.operator === operatorSymbol
+  );
+
+  // Return early do not add duplicate filters
+  if (isDuplicate) {
+    return;
```
matyax: Following the pattern of the Logs Panel and the table in Explore, this should remove the filter. But unfortunately, we don't have in any of these places an active state for this filters, which I think we should add.

matyax: If we decide this is good enough for now, we could potentially follow up with supporting active filter state and removing positive filters.

https://github.com/user-attachments/assets/359fe580-5563-4c90-bc3e-297456a39d0c

gtk-grafana: I'm not sure we need any changes to the addToFilters method, this should already be the behavior if we call addToFilters with `toggle` instead of `include`?

gtk-grafana: Yeah this change can be reverted, I think we just need to make this change in 
/src/Components/ServiceScene/Breakdowns/AddToFiltersButton.tsx#L76

```
- const type: FilterType = filter.operator === '=' ? 'include' : 'exclude';
+ const type: FilterType = filter.operator === '=' ? 'toggle' : 'exclude';
```

gtk-grafana: The change in this PR also introduces a bug with browser history since we still add the current url to history but then don't make a change, so an unnecessary entry is made in browser history requiring users to go back an extra time to get to the desired state.

L2D2Grafana: ty, updated!

gtk-grafana (approved):

PR #1432, opened 2025-07-18
feat(LogsPanel): customize interaction reporting and add storage key for patterns

matyax:
Requires https://github.com/grafana/grafana/pull/108272

- Excluded landing page and patterns from reporting interactions
- Added a custom key for the pattern logs, so that users can have different settings between these


matyax:
@gtk-grafana @L2D2Grafana Curious to know what you think of this. I think it makes sense.

PR #1439, opened 2025-07-22
chore(analytics): Visualization type on init

gtk-grafana:
Instead of only tracking when the viz is changed, let's track what viz is active when the logs tab loads.

### DIFF
--- a/src/Components/EmbeddedLogsExploration/EmbeddedLinkScene.tsx
+++ b/src/Components/EmbeddedLogsExploration/EmbeddedLinkScene.tsx
@@ -4,6 +4,7 @@ import { urlUtil } from '@grafana/data';
 import { SceneComponentProps, sceneGraph, SceneObjectBase, sceneUtils } from '@grafana/scenes';
 import { LinkButton } from '@grafana/ui';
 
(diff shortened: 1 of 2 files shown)


src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -181,6 +181,10 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
         }
       })
     );
+
+    reportAppInteraction(USER_EVENTS_PAGES.service_details, USER_EVENTS_ACTIONS.service_details.visualization_init, {
```
matyax, line 185: What do you think about sending this once per session, so it doesn't get noise from going back and forth to breakdowns?

gtk-grafana: yeah that might be a good way to get a smaller sample size.
Be cool if `reportAppInteraction` took an option that would say only fire once per session

matyax: I almost, almost, add a "reportInteractionOnce".

PR #1441, opened 2025-07-22
chore(JSON): memoize to prevent uncessary re-renders

gtk-grafana:
WIP

Checking https://github.com/aidenybai/react-scan/blob/main/BROWSER_EXTENSION_GUIDE.md against dev assets to identify components to be memoized.


gtk-grafana:
This is pretty insignificant performance wise at this time, but if components are made more complex down the road any additional renders we can remove should help.

The icons are still >90% of the performance impact, showing icons on hover or swapping out the svgs for sprites would make it much faster...

matyax:
Yeah, this rarely makes a huge difference in most cases.

matyax (approved):

L2D2Grafana (approved):
👍

PR #1442, opened 2025-07-23
omit: hide certain commit msgs

L2D2Grafana:
Allow certain commits to be hidden. `generate-changelog` doesn't have an automatic hook for github labels `no-changelog`, so not entirely sure how to add that without further research. This is the recommended way to omit commits for conventional commits, generate-changelog, pr title checking.


matyax (approved):
👏 👏

matyax:
Unintended consequence 

imagen

PR #1445, opened 2025-07-23
chore(ci): turn on the argo workflows for releasing to prod without auto merge

L2D2Grafana:
Okay I think this should be good. This will kick off argo workflows for release and I'll try to find the auto update deployment tools button in the workflow. I took out the auto merge to prod. I'm following what Traces has [done](https://github.com/grafana/traces-drilldown/blob/main/.github/workflows/release.yml) ty @adrapereira.


gtk-grafana (approved):
image

PR #1448, opened 2025-07-24
test(e2e): fix e2e

gtk-grafana:
Fixing testids that changed in core.
Skipping tests that touch the logs panel options as the new `logsPanelControls` is default as of the new Grafana release.

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -33,7 +33,7 @@ jobs:
 
       # Same settings as publish
       run-playwright-with-skip-grafana-dev-image: true
-      run-playwright-with-grafana-dependency: '>=12.1.3'
+      run-playwright-with-grafana-dependency: '>=12.1.0'
       upload-playwright-artifacts: true
       playwright-docker-compose-file: docker-compose.dev.yaml
       playwright-grafana-url: http://localhost:3001/grafana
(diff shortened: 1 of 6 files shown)


matyax (approved):

tests/exploreServicesBreakDown.spec.ts
```diff
@@ -201,14 +203,18 @@ test.describe('explore services breakdown page', () => {
...
   });
 
   test(`sync log panel displayed fields with table url columns`, async ({ page }) => {
     await explorePage.goToLogsTab();
 
     // Open log details
+    //ng-logs-panel
+    // const logPanelMenuLoc = page.getByLabel('Log menu').nth(1);
+    // await logPanelMenuLoc.click();
+    // await page.getByRole('menuitem', { name: 'Show log details' }).click();
```
L2D2Grafana, line 217: Is this for later?

gtk-grafana: yeah, we can remove if we want

.github/workflows/pr-ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       # Same settings as publish
       run-playwright-with-skip-grafana-dev-image: true
-      run-playwright-with-grafana-dependency: '>=12.0.0'
+      run-playwright-with-grafana-dependency: '>=12.1.0'
```
L2D2Grafana, line 36: Unfortunately 'latest' didn't work, but in the future we can increment to the latest version it's pulling like this.  By using `>=` we'll still auto-update and break on the next highest version, then need to update our workflow again.

PR #1451, opened 2025-07-24
chore(dep): update grafana dependencies for compatibility

L2D2Grafana:
Updated grafana dependencies to 12.1.0 for compatibility. rxjs needed to be updated, because two versions were getting installed.


gtk-grafana (approved):
jeez, who knew upgrading would be such a PITA. Great work on this

Issue #1453, opened 2025-07-24
[BUG]: Go to drilldown dashboard panel extension - multi variables not interpolating properly


gtk-grafana:
**Describe the bug**
If a dashboard query has a template **multi** variable, and the user navigates to drilldown via the go to logs drilldown link extension, the line filter value is wrapped by curly braces which breaks the query:

for example:
```
{service_name="service", kind="exception"} !~ `$errors` 
```

Creates a broken line filter variable in logs drilldown like:
```
{variableValue1,variableValue2}
```

But if you go to explore, and then go to logs drilldown (or just manually replace the variable in the logQL query), you get the expected (not broken) value:
```
variableValue1|variableValue2
```

[ops example](https://ops.grafana-ops.net/d/gaw2hh8/logs-drilldown-frontend-errors)

gtk-grafana:
Looks as though this is a bug in [sceneInterpolator](https://github.com/grafana/scenes/blob/main/packages/scenes/src/variables/interpolation/sceneInterpolator.ts)

gtk-grafana:
Wrong we need to provide the `interpolateQueryExpr` that is used in the loki datasource.

PR #1454, opened 2025-07-24
fix(Links): multi dashboard variable interpolation

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1453

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -15,3 +15,10 @@ jobs:
       - uses: amannn/action-semantic-pull-request@0723387faaf9b38adef4775cd42cfd5155ed6017 # v5.5.3
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          types: |
+            # Commit types omitted from changelogs
+            omit
+            internal
+            chore
+            skip
--- a/.versionrc.js
+++ b/.versionrc.js
@@ -1,0 +1,9 @@
+// Commit types omitted from changelogs
+module.exports = {
+  types: [
+    { type: 'omit', section: 'Omitted', hidden: true },
+    { type: 'internal', section: 'Internal', hidden: true },
+    { type: 'chore', section: 'Chores', hidden: true },
+    { type: 'skip', section: 'Skipped', hidden: true },
+  ],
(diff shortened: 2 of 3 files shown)


src/services/extensions/links.ts
```diff
@@ -276,3 +285,18 @@ export function escapeUrlPipeDelimiters(value: string | undefined): string {
...
+  // if no multi or include all do not regexEscape
+  if (!variable.multi && !variable.includeAll) {
+    return value;
+  }
+
+  if (typeof value === 'string') {
+    return escapeLabelValueInExactSelector(value);
+  }
+
+  const escapedValues = lodashMap(value, lokiSpecialRegexEscape);
```
matyax, line 300: Is map from lodash really needed for this?

gtk-grafana: We get it for free from grafana afaik, it's not in the bundle when I checked, and otherwise I just copied the method from core

matyax: Yeah, I mean a regular array map should do the same? Not really important anyway, but I was wondering about the reason to use it.

gtk-grafana: yeah i think it's just a style thing

gtk-grafana:
@L2D2Grafana I'm ignoring the conventional commit error: https://github.com/grafana/logs-drilldown/actions/runs/16521040424/job/46722716285?pr=1454 and gonna merge anyway. I think we need to fix that

matyax:
I reverted it in my pr

src/services/extensions/scenesMethods.ts
```diff
@@ -11,3 +11,11 @@
...
   return labelValue.replace(/\\/g, '\\\\').replace(/\n/g, '\\n').replace(/"/g, '\\"');
 }
+
+// Pulled from loki datasource
+export function lokiSpecialRegexEscape<T>(value: T) {
+  if (typeof value === 'string') {
+    return value.replace(/\\/g, '\\\\\\\\').replace(/[$^*{}\[\]+?.()|]/g, '\\\\$&');
+  }
+  return value;
+}
```
L2D2Grafana, line 21: These would be cool things to export in a shared package, maybe Q3!

PR #1455, opened 2025-07-25
fix(ServiceSelection): reset service selection after changing data source

matyax:
The Service Selection scene has logic to reuse existing panels, which was causing panels to be reused when switching data sources, which had the unintended consequence of not updating them when the data source changes. https://github.com/grafana/logs-drilldown/blob/main/src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx#L948

Example:

https://github.com/user-attachments/assets/743c4d92-d938-4e30-9485-b8b922ba113d

After these changes, we'll reset the layout when the data source is changed.

Additionally:
- The "splitDuration" experimental attribute has been removed.
- Fixed non-nullish conditional.
- Reverted [conventional commits change](https://github.com/grafana/logs-drilldown/pull/1442) preventing commits and failing in CI.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,7 +1,3 @@
 module.exports = {
   extends: ['@commitlint/config-conventional'],
-  rules: {
-    // Commit types omitted from changelogs
-    'type-enum': [2, 'always', ['omit', 'internal', 'chore', 'skip']],
-  },
 };
--- a/src/Components/IndexScene/ToolbarScene.tsx
+++ b/src/Components/IndexScene/ToolbarScene.tsx
@@ -34,7 +34,7 @@ export class ToolbarScene extends SceneObjectBase<ToolbarSceneState> {
         aggregatedMetrics: {
           active: active ?? false,
           disabled: false,
-          userOverride: userOverride === 'true' ?? false,
+          userOverride: userOverride === 'true',
         },
       },
       ...state,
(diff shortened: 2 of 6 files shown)


commitlint.config.js
```diff
@@ -1,7 +1,3 @@
 module.exports = {
   extends: ['@commitlint/config-conventional'],
-  rules: {
-    // Commit types omitted from changelogs
-    'type-enum': [2, 'always', ['omit', 'internal', 'chore', 'skip']],
```
matyax, line 5: imagen

src/Components/IndexScene/ToolbarScene.tsx
```diff
@@ -34,7 +34,7 @@ export class ToolbarScene extends SceneObjectBase<ToolbarSceneState> {
         aggregatedMetrics: {
           active: active ?? false,
           disabled: false,
-          userOverride: userOverride === 'true' ?? false,
```
matyax, line 37: ack

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -776,6 +771,17 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
     );
   }
 
+  private subscribeToDataSourceChanges() {
```
gtk-grafana: there'a already a `subscribeToDatasource` called method on line 645, let's merge those?

matyax: Oh for sure. Completely missed it

matyax: Done.

PR #1456, opened 2025-07-25
fix(serviceSelection): despect displayed fields if previously set

matyax:
This PR updates the service selection panel to use the displayed fields settings if they were previously set in the service details page.

How to test:
- Browse a service
- Set displayed fields
- Return to the service selection
- Displayed fields for the label/value pair should be respected*

Additionally, fixed the storage key when using a label other than service.

* It currently seems to require a refresh in the service selection page due to scene caching.

### DIFF
--- a/src/services/store.ts
+++ b/src/services/store.ts
@@ -6,11 +6,13 @@ import { Options } from '@grafana/schema/dist/esm/raw/composable/logs/panelcfg/x
 import { AvgFieldPanelType, CollapsablePanelText } from '../Components/Panels/PanelMenu';
 import { SortBy, SortDirection } from '../Components/ServiceScene/Breakdowns/SortByScene';
 import pluginJson from '../plugin.json';
+import { replaceSlash } from './extensions/links';
 import { isDedupStrategy } from './guards';
 import { logger } from './logger';
 import { unknownToStrings } from './narrowing';
-import { getDataSourceName, getServiceName } from './variableGetters';
-import { SERVICE_NAME } from './variables';
+import { getRouteParams } from './routing';
+import { getDataSourceName } from './variableGetters';
+import { SERVICE_NAME, SERVICE_UI_LABEL } from './variables';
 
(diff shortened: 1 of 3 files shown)


src/services/store.ts
```diff
@@ -196,9 +198,25 @@ export function setSortByPreference(target: string, sortBy: string, direction: s
...
 function getExplorationPrefix(sceneRef: SceneObject) {
+  const { labelName, labelValue } = getRouteParams(sceneRef);
+  return getExplorationPrefixForlabelValue(sceneRef, labelName, labelValue);
+}
+
+function getExplorationPrefixForlabelValue(sceneRef: SceneObject, label: string, value: string) {
   const ds = getDataSourceName(sceneRef);
-  const serviceName = getServiceName(sceneRef);
-  return `${ds}.${serviceName}`;
+  if (label === SERVICE_NAME || label === SERVICE_UI_LABEL) {
```
matyax, line 207: This is mainly to keep backwards compatibility with previously stored displayed fields, which did not contain the label in the path.

gtk-grafana:
Looks good but I'm not sure if the columns persisting when changing the primary label is expected?

matyax:
> columns persisting when changing the primary label is expected?

Do you mean displayed fields being kept after changing the label value? That seems to preexist these changes, but it should probably be reset, although is one of those bugs that feels like a feature.

gtk-grafana:
Yeah, it gets a bit weird if you hop around between different services or labels, primarily if you had a column set already set for cluster="us-east-1", and you navigate to us-east-2, and set a different column and then switch to us-east-1, the column set for us-east-1 is overwritten with the one from us-east-2, I would expect it to only set the default if the user changed the columnset while visualizing that label

matyax:
Yeah, that's probably getting stuck in the scene state. Let's fix it.

gtk-grafana:
Eh, actually that behavior appears to be on main as well, so you can tackle in another PR if you want

matyax:
Will do it in another PR mainly because I ran out of time today. We'll discuss there what do we want to do with displayed fields when the user changes the primary label.

Issue #1457, opened 2025-07-28
[FEAT]: Patterns: filter by level


matyax:
Patterns now include level information. We should support filtering by level.

Proposal: don't add a new control and use the level filter from the header.

PR #1459, opened 2025-07-28
feat(Patterns): filter by level

gtk-grafana:
Merges patterns with the same pattern name.
image

Fixes: https://github.com/grafana/logs-drilldown/issues/1457
Also fixes a bug where the pattern count can be incorrect if a pattern contains multiple levels.

Notes for reviewers:
You must re-run `yarn server` to see levels on the pattern responses.

### DIFF
--- a/docker-compose-syslog.dev.yaml
+++ b/docker-compose-syslog.dev.yaml
@@ -20,7 +20,7 @@ services:
       - 'host.docker.internal:host-gateway'
 
   loki:
-    image: grafana/loki:main-c40053a
+    image: grafana/loki:k263-d446d47
     environment:
       LOG_CLUSTER_DEPTH: '8'
       LOG_SIM_TH: '0.3'
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
@@ -90,7 +90,7 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
           )}
 
           {!error && !loading && patternFrames?.length === 0 && timeRangeTooOld && <PatternsTooOld />}
-          {!error && !loading && patternFrames?.length === 0 && !timeRangeTooOld && <PatternsNotDetected />}
+          {!error && !loading && !patternFrames && !timeRangeTooOld && <PatternsNotDetected />}
           {!error && !loading && patternFrames && patternFrames.length > 0 && (
             <div className={styles.content}>{body && <body.Component model={body} />}</div>
           )}
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
@@ -41,6 +41,7 @@ export interface PatternsBreakdownSceneState extends SceneObjectState {
 
 export type PatternFrame = {
   dataFrame: DataFrame;
+  level?: string;
   pattern: string;
   status?: 'exclude' | 'include';
   sum: number;
@@ -174,11 +175,13 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
       const existingPattern = appliedPatterns?.find((appliedPattern) => appliedPattern.pattern === dataFrame.name);
 
       const sum: number = dataFrame.meta?.custom?.sum;
+      const level: string | undefined = dataFrame.meta?.custom?.level;
       const patternFrame: PatternFrame = {
         dataFrame,
         pattern: dataFrame.name ?? '',
         status: existingPattern?.type,
         sum,
+        level,
       };
 
       return patternFrame;
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
@@ -41,7 +41,7 @@ export interface PatternsBreakdownSceneState extends SceneObjectState {
 
 export type PatternFrame = {
   dataFrame: DataFrame;
-  level?: string;
+  level: string[];
   pattern: string;
   status?: 'exclude' | 'include';
   sum: number;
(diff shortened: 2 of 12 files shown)


src/Components/ServiceScene/Breakdowns/Patterns/PatternsBreakdownScene.tsx
```diff
@@ -174,11 +175,13 @@ export class PatternsBreakdownScene extends SceneObjectBase<PatternsBreakdownSce
       const existingPattern = appliedPatterns?.find((appliedPattern) => appliedPattern.pattern === dataFrame.name);
 
       const sum: number = dataFrame.meta?.custom?.sum;
+      const level: string[] = dataFrame.meta?.custom?.level;
```
gtk-grafana: rename all these to `levels`

src/services/datasource.ts
```diff
@@ -513,6 +540,24 @@ export class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+  s2.forEach((value) => {
+    if (lokiSamplesMap.has(value[0])) {
+      const existingValue = lokiSamplesMap.get(value[0]);
+      lokiSamplesMap.set(value[0], value[1] + (existingValue ?? 0));
+    } else {
+      lokiSamplesMap.set(value[0], value[1]);
+    }
+  });
+  const samples = Array.from(lokiSamplesMap);
+  samples.sort();
```
gtk-grafana: we don't need to sort every time we call this if we sort once before creating the df

src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
```diff
@@ -187,6 +193,31 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
         id: 'include',
       },
     ];
+
+    if (patternFrames.some((pattern) => pattern.level !== undefined)) {
```
gtk-grafana: wrong

src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
```diff
@@ -187,6 +193,31 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
...
+      columns.splice(1, 0, {
+        header: 'Level',
+        id: 'level',
+        sortType: 'string',
+        cell: (props: CellProps<PatternsTableCellData>) => {
+          return props.cell.row.original.level.map((level) => (
+            <Button
```
gtk-grafana, line 202: needs active state

gtk-grafana:
TODO: 
* Empty view when level filters are active
* Custom sort method for levels column
* e2e coverage
* e2e coverage for patterns API call on time range range

src/services/datasource.ts
```diff
@@ -201,54 +206,75 @@ export class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
-              custom: {
-                sum,
-              },
-              preferredVisualisationType: 'graph',
-            },
-            name: pattern.pattern,
-            refId: interpolatedTarget.refId,
-          });
-        }) ?? [];
+      let lokiPatternsProcessed: Record<string, LokiPatternProcessed> = {};
```
matyax, line 209: Would you mind adding at least one simple unit test to back all these calculations up?

src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
```diff
@@ -187,6 +190,35 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
...
       },
     ];
+
+    if (patternFrames.some((pattern) => pattern.levels.length > 0)) {
+      columns.splice(1, 0, {
+        header: 'Levels',
+        id: 'levels',
+        // @todo custom sort method?
+        cell: (props: CellProps<PatternsTableCellData>) => {
+          return props.cell.row.original.levels.map((level) => (
```
matyax, line 201: I like a lot that we can filter from the list. The one thing that thew me off for a good 10 seconds was the color:

imagen

My first thought was "are we color coding levels?", not realizing that these were link buttons. Maybe we should use outlined buttons, like most other filters to prevent this.

docker-compose-syslog.dev.yaml
```diff
@@ -20,7 +20,7 @@ services:
       - 'host.docker.internal:host-gateway'
 
   loki:
-    image: grafana/loki:main-c40053a
+    image: grafana/loki:k263-d446d47
```
L2D2Grafana, line 23: How do you know to use this loki image?

src/services/datasource.ts
```diff
@@ -513,6 +539,24 @@ export class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
   }
 }
 
+function mergeLokiSamples(s1: PatternSample[], s2: PatternSample[]) {
```
L2D2Grafana: Is this the part doing doing deduplication? Perhaps a comment would be nice on why we need to merge numbers with the same timestamp.

src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
L2D2Grafana: A tooltip with +/- or include/exclude might also be nice

gtk-grafana: we don't have exclude anywhere in the UI for levels (besides the log viz panels)

gtk-grafana: image
image
image

Fixed in https://github.com/grafana/logs-drilldown/pull/1459/commits/5aa57dde8659652258b99ad7da416558877c6439

docker-compose-syslog.dev.yaml
gtk-grafana: I asked Trevor 😆

src/services/datasource.ts
gtk-grafana: what you don't trust a sleep deprived Galen merging dataframes?
Good call, the sort was messed up (although it looks like it doesn't matter) :P 

Fixed in https://github.com/grafana/logs-drilldown/pull/1459/commits/4e80242874a48cc0532c9e5726425ab76508a525

src/services/datasource.ts
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1459/commits/a9cb22a04d56c5ac20c251222b2ecb146a192d01

src/services/datasource.ts
matyax: Relatively complex code is always a good candidate for a unit test. Most of the time I discover it doesn't exactly does what I think it does.

Issue #1465, opened 2025-07-28
[BUG]: Support new max series behavior in k263


gtk-grafana:
**Description of the bug**
Loki now returns partial results that contain more series then the max limit instead of throwing an error.

For users of latest Loki, we no longer show the toggle to show error fields, but users are not told that the results are partial.
We need to show a warning in the UI and update e2e tests.

PR #1467, opened 2025-07-30
fix(fields): show max series notice in panel

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1465

Note for reviewer: requires loki k263 added in https://github.com/grafana/logs-drilldown/pull/1459.
Leaving in draft until that PR is merged.

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/FieldsAggregatedBreakdownScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/FieldsAggregatedBreakdownScene.tsx
@@ -25,7 +25,7 @@ import {
   isAvgField,
 } from '../../../services/fields';
 import { logger } from '../../../services/logger';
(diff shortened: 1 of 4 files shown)


L2D2Grafana (approved):
👏 lgtm
Image

How did you find out only partial data was being returned for fields? Would something like adaptive logs not also affect labels or patterns?

gtk-grafana:
> How did you find out only partial data was being returned for fields? Would something like adaptive logs not also affect labels or patterns?

Ooooh **** I didn't notice that adaptive logs throws a notice in the meta.notices. We probably don't want to show the warning in that case, I'll add an inclusion regex on the max series notice. Good catch!
Fixed in https://github.com/grafana/logs-drilldown/pull/1467/commits/e4d8bfeb27b16800cf1da708426e8ad825785aec

matyax:
👌

PR #1469, opened 2025-07-31
fix(shardQuerySplitting): stop when hitting max series

matyax:
Before some recent Loki changes, max series reached was triggering an error response. Now, it changed to return partial data and a notice. That caused sharding to continue querying, which not only was unnecessary, but caused lag by trying to merge many many series in the frontend.


gtk-grafana (approved):
LGTM much better, but we need to find a way to avoid merging all those unvized series

PR #1471, opened 2025-08-01
feat(combineResponses): workerize

matyax:
An option.

### DIFF
--- a/src/services/shardQuerySplitting.ts
+++ b/src/services/shardQuerySplitting.ts
@@ -6,6 +6,7 @@ import { DataFrame, DataQueryRequest, DataQueryResponse, LoadingState, QueryResu
 import { MaxSeriesRegex } from '../Components/ServiceScene/Breakdowns/QueryErrorAlert';
 import pluginJson from '../plugin.json';
 import { combineResponses } from './combineResponses';
+import { combineResponsesWorker } from './combineResponsesWrapper';
 import { logger } from './logger';
 import { addShardingPlaceholderSelector, getSelectorForShardValues, interpolateShardingSelector } from './logql';
 import { isValidQuery } from './logqlMatchers';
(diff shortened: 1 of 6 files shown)


L2D2Grafana (commented):
This is a great idea 💡 !

I'm seeing some errors around workerizer-loader, I wonder if we could try library like comlink? https://npmtrends.com/comlink-vs-workerize-loader 
Image

Also to fix the test you'd want to update the jest.config.js moduleNameMapper with the workerizer-loader dep
```
module.exports = {
  // Jest configuration provided by Grafana scaffolding
  ...config,
  moduleNameMapper: {
    ...config.moduleNameMapper,
    '/@bsull/augurs/changepoint/': '@bsull/augurs/changepoint.js',
    '/@bsull/augurs/outlier/': '@bsull/augurs/outlier.js',
    '^marked$': '<rootDir>/src/__mocks__/marked.js',
    '^react-calendar$': '<rootDir>/src/__mocks__/react-calendar.js',
  },
  transform: {
    ...config.transform,
    '^.+\\.svg$': '<rootDir>/.config/svgTransform.js',
  },
  transformIgnorePatterns: [nodeModulesToTransform([...grafanaESModules, '@bsull/augurs'])],
};
```

matyax:
I quickly glanced Comlink, but also seemed as old as workerize loader. I'll give it a look, because I'm miserably failing with any webpack-native approach.

src/services/shardQuerySplitting.ts
```diff
@@ -167,7 +168,9 @@ function splitQueriesByStreamShard(
           done();
           return;
         }
-        nextRequest();
+        if (!combineResponsesWorker) {
```
gtk-grafana, line 183: lets make sure we test this after disabling web workers, in an iframe, and in a mobile webview

matyax: Yeah, definitely. It needs to run if workers are not available.

matyax: Can't find a way to disable web workers, but added a bit of extra safety + overrode the worker globals, and it used original code as a fallback.

From Comlink:

imagen

gtk-grafana:
Sounds like we need to put this on the back burner as web workers won't work when the plugin is served from the CDN :( 

Blocked by https://github.com/grafana/plugin-tools/issues/2007

PR #1472, opened 2025-08-01
chore(performance): Sharding performance - always join on labels

gtk-grafana:


src/services/combineResponses.ts
```diff
@@ -11,18 +11,56 @@ import {
...
+      const key = getFrameKey(newFrame);
+
+      if (currentResultLabelsMap.has(key)) {
+        currentFrame = currentResultLabelsMap.get(key);
+        mergeFrames(currentFrame!, newFrame);
+      } else {
+        currentResult.data.push(cloneDataFrame(newFrame));
+      }
+    } else {
+      console.log('fallback merge');
```
gtk-grafana: I think this can get removed, it's never been called in my testing, new dataplane spec or old

src/services/combineResponses.ts
```diff
@@ -11,18 +11,56 @@ import {
 
 import { logger } from './logger';
 
+function getFrameKey(frame: DataFrame): string {
+  const field = frame.fields.find((f) => f.type === FieldType.number);
+  if (!field) {
+    throw new Error(`Unable to find number field on sharded dataframe!`);
```
matyax, line 17:
```suggestion
    throw new Error(`Unable to find number field on sharded dataframe!`);
```

I'm a bit hesitant to throw these errors. If this would ever happen (e.g. bug in the data source), we can skip and continue.

src/services/combineResponses.ts
matyax: Or just leave 

```
currentResult.data.push(cloneDataFrame(newFrame));
```

src/services/combineResponses.ts
gtk-grafana: Can I just wrap `combineResponses` in a catch block, log the error, and bail out there? It makes the code so nasty if everything needs to optionally return undefined

matyax:
Feel free to remove all logs-related code.

src/services/combineResponses.ts
matyax: Definitely!

src/services/combineResponses.ts
gtk-grafana: I threw an error instead. If there are bad frames getting sharded somehow I want to know about it

matyax: You want, but the user probably doesn't.

gtk-grafana: we're catching the error and logging through faro, user shouldn't be aware

gtk-grafana:
@leeoniya reminded me to use `amendTable` https://github.com/grafana/grafana/blob/9f4773c9a5a35f7452e3df92006ac050fd397052/packages/grafana-data/src/table/amendTimeSeries.ts#L8, for the actual merge which might allow us to clean up some of this in a later PR.

Although I think amendTable replaces values for overlapping time stamps, so maybe not without some tweaks

Issue #1481, opened 2025-08-05
[FEAT]: Report query usage


gtk-grafana:
**Description of the new feature**
Explore shows the estimated processing size, e.g. `This query will process approximately 116.3 MiB.`.

Although it would be difficult for Logs Drilldown to display the query usage stats before queries are made, we should at least display the total bytes processed for the current page or session.

**Acceptance Criteria**
Logs drilldown displays total bytes processed by all queries.

PR #1482, opened 2025-08-05
feat: show processed bytes

gtk-grafana:
image

Fixes: https://github.com/grafana/logs-drilldown/issues/1481

Sums all totalBytesProcessed received from query runners.
Note, not all queries currently return this, like detected_fields, detected_labels, patterns, volume, so the count will not be entirely accurate.
Value is persisted in session storage for the current tab.


matyax:
This is great, but I don't think that's the right place to put it. It's floating text that is not alinged with anything, other than arbitrarily to the Data Source picker, and is permamently visible on top.

474692256-53b41826-08a2-49ee-a9d0-8c599eec38d7

What about putting it behind an (i) icon that would show this data on mouse over? 

imagen

Similar to Traces, but not necessarily the same.

imagen

gtk-grafana:
@matyax

> I don't think that's the right place to put it. It's floating text that is not alinged with anything, other than arbitrarily to the Data Source picker, and is permamently visible on top.

Yeah I'm also not in love with the placement, but I do think this should be permanently visible to help reinforce better querying practices. 7 day query over all fields eats up hundreds of GB, but adding a single label filter reduces it down to <10, etc.

How about aligning the feedback form/ bytes processed wrapper to flex-end instead of space-between?
image

> Similar to Traces, but not necessarily the same.

If we add something similar to traces (but different functionality) that's going to cause more confusion for users jumping across traces/logs?

Also we already have an options wheel when aggregated metrics is enabled
image

matyax:
> How about aligning the feedback form/ bytes processed wrapper to flex-end instead of space-between?

Looks better there. The only extra change that I would suggest is to use secondary text font, not so brilliant.

> but I do think this should be permanently visible to help reinforce better querying practices

I'd be very selective with what is placed in such a visible and high spot.

> but adding a single label filter reduces it down to <10, etc

For the average user, there's currently no way to arrive to that conclusion, other than by deduction. What do you think of also adding a tooltip with a few suggestions to accompany it?

gtk-grafana:
Changed in https://github.com/grafana/logs-drilldown/pull/1482/commits/fb4e3a5dcdce967b307f2523cb1e80930df70198
image

matyax:
Yeah, that's much better. 

I still think it's not the place for it and I'm not 100% convinced it should be permanently visible (~~I appreciate a lot the remove button 👌~~ I confused the x with a remove button), but if we can agree on finding a better place for it if we need to continue iterating on it (i.e. showing more data, etc), let's give it a try.

matyax:
> Note, not all queries currently return this, like detected_fields, detected_labels, patterns, volume, so the count will not be entirely accurate.

This is also another reason to be cautious with prominently showing this as permanently visible information. It makes me think that it should be prefixed with "Approximately total bytes processed" or have a *.

gtk-grafana:
image
image

src/Components/IndexScene/VariableLayoutScene.tsx
```diff
@@ -71,7 +72,27 @@ export class VariableLayoutScene extends SceneObjectBase<VariableLayoutSceneStat
...
               <div className={styles.controlsWrapper}>
-                {!indexScene.state.embedded && <GiveFeedbackButton />}
+                <div className={styles.feedbackWrapper}>
+                  <span className={styles.bytesProcessed}>
+                    <Tooltip
+                      content={`Processed bytes are an approximate lower bound, the actual value may be slightly higher`}
+                    >
+                      <span>
+                        Bytes processed: {bytesProcessed.text} {bytesProcessed.suffix}
+                        <sup>*</sup>
```
matyax, line 82: Cool! You could also use this opportunity to add these as i18n literals.

matyax:
Looks great. Let's have a quick chat today before merging to make sure we're all aligned around this.

src/services/panel.ts
```diff
@@ -324,19 +327,45 @@ export function getQueryRunner(queries: LokiQuery[], queryRunnerOptions?: Partia
...
+  queryRunner.addActivationHandler(() => {
+    const indexScene = sceneGraph.getAncestor(queryRunner, IndexScene);
+    queryRunner.subscribeToState((newState, prevState) => {
+      if (
+        newState.data?.state === LoadingState.Done ||
+        newState.data?.state === LoadingState.Streaming ||
+        newState.data?.state === LoadingState.Error
+      ) {
+        const df = newState.data?.series.find((df) => df.meta?.stats);
+        const stats = df?.meta?.stats;
```
L2D2Grafana, line 355: Ah cool it was there the whole time 🪄

PR #1486, opened 2025-08-06
feat(FARO): add faro instrumentation

L2D2Grafana:
### Details
Implement FARO for observability and potentially SLOs. Shoutout to @grafakus and Metrics for leading the way here 👏 Let me know if I've missed anything. [Docs](https://grafana.com/docs/grafana-cloud/monitor-applications/frontend-observability/)

- Created FARO environments - [local,dev,ops,prod](https://ops.grafana-ops.net/a/grafana-kowalski-app)
- Configured the web sdk
- Added Source Maps for dev, ops, prod - Unsure this will work as the token needs to be passed to webpack by a github action
- Alerts for errors in prod
- Updated logger to integrate with our instance of Faro, unsure of the potential ramifications I believe these would show up as they previously were in our logs.
- Log info message`Plugin loaded successfully` This is to use in a potential future SLO against `Could not load plugin`

### Test
- In faroEnv uncomment, yarn dev, load the app
```
  // [
  //   'local',
  //   {
  //     environment: 'local',
  //     appName: 'grafana-logsdrilldown-app-local',
  //     faroUrl: 'https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/54fb57556eb45f8857edb9f0345c0d7a',
  //   },
  // ],
```
- View local faro in ops https://ops.grafana-ops.net/a/grafana-kowalski-app/apps/58/overview 
Screenshot 2025-08-06 at 12 14 10 PM
Screenshot 2025-08-06 at 12 19 17 PM
Screenshot 2025-08-06 at 12 15 15 PM

### Todo
-  Token mgmt


src/Components/App.tsx
```diff
@@ -1,10 +1,14 @@
...
 
 import { AppRootProps } from '@grafana/data';
+
+import { initFaro } from 'faro/faroInit';
 const LogExplorationView = lazy(() => import('./LogExplorationPage'));
 
 const PluginPropsContext = React.createContext<AppRootProps | null>(null);
 
+// Initialize Faro for internal observability
+initFaro();
```
gtk-grafana, line 11: can we lazy load this to keep the bundle size down?

L2D2Grafana:
@gtk-grafana took me a minute to figure out where the logs were going for the logger.ts, but they are getting sent to the Logs Drilldown scoped ops envs.
Screenshot 2025-08-27 at 11 35 38 AM
I also cannot seem to find the info logs in the scoped envs and need to ask the team about that. 

I believe before the logError in grafana/runtime was getting piped to some global faro instance and we would be losing that capability with this change. Do you think it's okay to track errors in those ops environments? If not I could revert the changes to logger.ts and continue using logError. We'd have to monitor in both places.

gtk-grafana:
I'll take a closer look tomorrow but the scoped envs are probably fine, I don't think we want to monitor in multiple locations? Are we just losing access to any errors thrown outside of Logs Drilldown?

src/faro/faroInit.ts
```diff
@@ -0,0 +1,66 @@
...
+        version: pluginVersion,
+        environment,
+      },
+      user: {
+        email: userEmail,
+      },
+      instrumentations: [
+        // Mandatory, omits default instrumentations otherwise.
+        ...getWebInstrumentations({
+          captureConsole: false,
```
gtk-grafana: should capture console be true?

src/faro/faroInit.ts
```diff
@@ -0,0 +1,66 @@
...
+        }),
+        // Tracing package to get end-to-end visibility for HTTP requests.
+        new TracingInstrumentation(),
```
gtk-grafana: noice

src/faro/getEnv.ts
```diff
@@ -0,0 +1,30 @@
...
+  {
+    regExp: /grafana-dev\.net/,
+    environment: 'dev',
+  },
+  {
+    regExp: /grafana-ops\.net/,
+    environment: 'ops',
+  },
+  {
+    regExp: /grafana\.net/,
```
gtk-grafana: I don't think all prod instances contain `grafana.net`?

src/faro/faroEnv.ts
```diff
@@ -0,0 +1,38 @@
...
+import { type FaroEnvironment } from './getFaroEnv';
+
+export const FARO_ENVIRONMENTS = new Map<Environment, FaroEnvironment>([
+  // Uncomment this map entry to test from your local machine
+  // [
+  //   'local',
+  //   {
+  //     environment: 'local',
+  //     appName: 'grafana-logsdrilldown-app-local',
+  //     faroUrl: 'https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/54fb57556eb45f8857edb9f0345c0d7a',
```
gtk-grafana: i think we listen on `http://localhost:12347/collect` right now for local faro

gtk-grafana: also I'd like to instrument this so it only sends when the docker-compose.local.yaml is kicking things off instead of requiring code changes

L2D2Grafana: Updated this to be on all the time.

src/faro/faroInit.ts
L2D2Grafana: This was to cut down on logs we might not need, but turning it on for now. We can turn it off later if it's too noisy.

src/faro/faroInit.ts
L2D2Grafana: In the captured sessions the traces should show up here.
Image

src/faro/getEnv.ts
L2D2Grafana: Updated to default in ops changed to `*` which a few other apps are doing like Assistant.

gtk-grafana:
@L2D2Grafana appears to be working locally when running the local docker config `yarn server`, but I'm seeing a bunch of CORS errors trying to post to `https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/54fb57556eb45f8857edb9f0345c0d7a`

```
Access to fetch at 'https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/54fb57556eb45f8857edb9f0345c0d7a' from origin 'http://localhost:3000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

L2D2Grafana:
> @L2D2Grafana appears to be working locally when running the local docker config `yarn server`, but I'm seeing a bunch of CORS errors trying to post to `https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/54fb57556eb45f8857edb9f0345c0d7a`
> 
> ```
> Access to fetch at 'https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/54fb57556eb45f8857edb9f0345c0d7a' from origin 'http://localhost:3000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
> ```

I just fired it all up again locally, but I'm not seeing the CORS errors. Anything else you did to reproduce?
Screenshot 2025-09-03 at 7 55 28 AM

gtk-grafana:
Oh my bad, I'm not seeing the cors error on the yarn server grafana (3001), just when running Grafana main locally from source.

gtk-grafana:
But I don't think we want to pollute ops with logs from local anyway?

src/Components/App.tsx
```diff
@@ -1,10 +1,16 @@
...
+
 const LogExplorationView = lazy(() => import('./LogExplorationPage'));
 
 const PluginPropsContext = React.createContext<AppRootProps | null>(null);
 
+// Initialize Faro for internal observability
+import('faro/faroInit').then((module) => {
+  module.initFaro();
+});
+
```
matyax: Out of curiosity, is there any reason to use a dynamic import?

```suggestion
// Initialize Faro for internal observability
import { initFaro } from 'faro/faroInit';

initFaro();
```

L2D2Grafana: I originally had the vanilla import, but Galen wanted the dynamic import. I can go either way. Metrics used the vanilla import.

L2D2Grafana:
> But I don't think we want to pollute ops with logs from local anyway?

Ah okay so running it with grafana/grafana is the issue. I can go back to commenting it out by default and people just turn it on when they want to debug something locally with Faro? It will not pollute ops, I've separated out the envs.

gtk-grafana:
@L2D2Grafana I guess I'm curious as to why 3001 works but we get a cors error on port 3000, but I guess commenting it out for local should work since we're still sending local payloads to `http://localhost:12347/collect` when running `yarn server`. 

But I do wonder if there is a solution here using the `GF_LOG_FRONTEND_CUSTOM_ENDPOINT`, each Grafana instance is already configured where to send frontend logs? 

I'm not sure there's much value in sending frontend errors from local env to ops, while it might be interesting to creep on y'alls errors, lots of stuff will get thrown when developing that isn't actionable, so yeah commenting out is probably best for now. 

Although this makes me wonder about OSS users, will they see CORS errors in prod since this instrumentation can only get sent in cloud?

src/Components/App.tsx
gtk-grafana: I wanted faro to get loaded async so it's not in the module.js bundle, but it looks like it is again
image
image

gtk-grafana (changes requested):
We cannot have any faro urls defined in the source code, they need to be pulled from env variables or OSS users will get CORS errors and/or send us telemetry they haven't consented to

src/faro/faroEnv.ts
```diff
@@ -0,0 +1,37 @@
+import { type Environment } from './getEnv';
+import { type FaroEnvironment } from './getFaroEnv';
+
+export const FARO_ENVIRONMENTS = new Map<Environment, FaroEnvironment>([
+  [
+    'local',
+    {
+      environment: 'local',
+      appName: 'grafana-logsdrilldown-app-local',
+      faroUrl: 'https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/54fb57556eb45f8857edb9f0345c0d7a',
```
gtk-grafana, line 10: These need to get pulled from env vars which are defined in the grafana instance instead of being defined in the code

src/Components/App.tsx
L2D2Grafana: Updated!
Screenshot 2025-09-04 at 2 21 34 PM

src/faro/faroEnv.ts
L2D2Grafana: AFAIK these are defined in ops in the web sdk config. @eskirk do you know if there is some way to get this info?

gtk-grafana: I think we need to add a new config field to core Grafana and populate these in deployment_tools for dev/ops/prod

L2D2Grafana:
Closing for now.

Issue #1060, opened 2025-02-11
Table: Column header menu is not keyboard accessible


gtk-grafana:
There's no way to select an option from the keyboard.

<img width="535" alt="Image" src="https://github.com/user-attachments/assets/7b3fd1fa-cfb3-4a52-a496-749bdddb4ff3" />

PR #1490, opened 2025-08-07
fix(table): make column menu keyboard accessible

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1060

image

There is still quite a bit sub-optimal behavior around focusing elements in the table, primarily any action that moves or removes a column in the dialog will reset the tabindex to the start.

It might even be best to leave the column headers inaccessible to keyboard navigation as there are other UI elements that provide all functionality within the table header component with less unexpected behavior (column management (left) sidebar, or table options (right)).

Argument 1:
If we add more functionality in the column headers that isn't available elsewhere in the UI, then this PR should be merged, but now I'm thinking that we should not merge this PR, and potentially have a discussion about removing the column header component all together.

Argument 2:
Move left/right is nice, we should merge this PR to see if folks are getting use out of this, and then reevaluate argument 1.

### DIFF
--- a/src/Components/Table/LogsTableHeader.tsx
+++ b/src/Components/Table/LogsTableHeader.tsx
@@ -3,11 +3,12 @@ import React, { PropsWithChildren, useRef } from 'react';
 import { css } from '@emotion/css';
 
 import { Field, GrafanaTheme2 } from '@grafana/data';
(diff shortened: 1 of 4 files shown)


L2D2Grafana (approved):
This is really nice, the ❌ to close 👍 

Image

PR #1500, opened 2025-08-12
feat(EmbeddedLogs): fix missing keyLabel in parsed line filters

matyax:
Parsed line filters were being passed as "defaultLineFilters", but because they came from the parser, they were lacking a generated attribute called `keyLabel`.

Additionally, switched from using mutation to pure functions and added a few unit tests.

Error

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -711,7 +711,11 @@ function getVariableSet(
 
   const lineFiltersVariable = new AdHocFiltersVariable({
     expressionBuilder: renderLogQLLineFilter,
-    filters: defaultLineFilters ?? [],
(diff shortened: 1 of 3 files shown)


matyax:
Oh, I forgot to add how to repro. Just go to the /embed route and you'll see it. http://localhost:3000/a/grafana-lokiexplore-app/embed?ld-patterns=%5B%5D&ld-var-all-fields=&pageSlug=logs&ld-displayedFields=%5B%5D&ld-urlColumns=%5B%5D&ld-visualizationType=%22logs%22&ld-var-filters=service_name%7C%3D%7Ctempo-distributor&ld-prettifyLogMessage=false&ld-sortOrder=%22Ascending%22&ld-wrapLogMessage=false

imagen

Issue #1503, opened 2025-08-13
[BUG]: Levels: allow custom options in filter


matyax:
The values endpoint often omits certain options that are not present in the sample taken. We should at least allow users to use any arbitratry value instead of limiting to the available options.

PR #1509, opened 2025-08-19
feat(levels): allow custom options in level variable

gtk-grafana:
Allows users to add a custom level value.
Since the detected_fields endpoint that returns level values is sampled, some valid values might not be presented to the user in the variable dropdown. Lets let users add a custom value since the list of options in the dropdown is not exhaustive. 

Fixes: https://github.com/grafana/logs-drilldown/issues/1503


matyax (approved):
👏 👏

Issue #1512, opened 2025-08-20
[BUG]: JSON - Icons hard to see in light mode


gtk-grafana:
**Describe the bug**

<img width="813" height="508" alt="Image" src="https://github.com/user-attachments/assets/12819484-7555-4457-a7dc-8255366d6ef7" />

https://play.grafana.org/goto/2MFVYVuHg?orgId=1

PR #1513, opened 2025-08-20
fix(json): light icon buttons

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1512

Also fixes the copy/share link buttons which were still using the inline svg icons (`IconButton` component), so while this PR does have a small increase in JSON bundle size (which is lazy loaded and will only be loaded when the JSON viz is active), this PR will enhance render performance by removing 2 * logLineCount svg renders.


matyax (approved):
Icons look good with light themes!

PR #1518, opened 2025-08-26
chore: downgrade field errors

gtk-grafana:
Problem:
The base panels don't have any concept of notices, there are panel errors or there aren't.
With recent Loki changes we've added "expected" notices for high cardinality fields so users are able to see panels for all of the fields, but when max series limits are exceeded we return partial data. 

But for users with many high cardinality fields, they are greeted with an onslaught of red errors in every panel, which implies the user did something wrong or has an action to take to correct the problem, but fields with unique IDs will always return these errors if there are more logs then the max series limit set in Loki.

image

<hr />

This PR simply hacks the CSS of the panel to "tone it down" a bit.

image

image
image
image

### DIFF
--- a/src/Components/Panels/PanelMenu.tsx
+++ b/src/Components/Panels/PanelMenu.tsx
@@ -404,7 +404,6 @@ async function subscribeToAddToInvestigation(exploreLogsVizPanelMenu: PanelMenu)
 
 export const getPanelWrapperStyles = (theme: GrafanaTheme2) => {
   return {
-    errorWrapper: css({}),
     panelWrapper: css({
       display: 'flex',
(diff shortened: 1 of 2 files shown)


matyax:
I'm thinking out loud. Can we consider moving these to the bottom? Or, at least, that these super cardinality fields are not always at the top? Because not only these will always be showing all the warning icons that are throwing people off, but also don't seem to be super valuable to be the first things that you see in this page.

imagen

matyax (approved):
Looks good!

gtk-grafana:
@matyax I'll open a new issue from your comment to discuss the sort order, it gets complex

PR #1519, opened 2025-08-26
chore: add tenant id to generator, and enable auth in loki

trevorwhitney:
This PR changes how logs are generated and queried in the local dev stacks to use a tenant id of `1`.

### DIFF
--- a/config/loki-config.yaml
+++ b/config/loki-config.yaml
@@ -1,4 +1,4 @@
-auth_enabled: false
+auth_enabled: true
 
 server:
   http_listen_port: 3100
--- a/docker-compose-syslog.dev.yaml
+++ b/docker-compose-syslog.dev.yaml
@@ -49,6 +49,6 @@ services:
     build:
       context: ./generator
       dockerfile: Dockerfile.syslog
-    command: -otel=false -syslog=true -syslog-network=udp -syslog-addr=alloy:1514
+    command: -otel=false -syslog=true -syslog-network=udp -syslog-addr=alloy:1514 -tenant-id=1
     depends_on:
       - alloy
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -35,6 +35,6 @@ services:
   generator:
     build:
       context: ./generator
-    command: -url http://loki:3100/loki/api/v1/push
+    command: -url http://loki:3100/loki/api/v1/push -tenant-id=1
     environment:
       - OTLP_ENDPOINT=http://loki:3100/otlp
--- a/docker-compose.local.yaml
+++ b/docker-compose.local.yaml
@@ -41,7 +41,7 @@ services:
   generator:
     build:
       context: ./generator
-    command: -url http://loki:3100/loki/api/v1/push
+    command: -url http://loki:3100/loki/api/v1/push -tenant-id=1
     environment:
       - OTLP_ENDPOINT=http://loki:3100/otlp
   alloy:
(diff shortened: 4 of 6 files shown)


gtk-grafana:
Seems to work! As a note, anyone running Grafana from source will need to add `X-Scope-OrgID` with a value of `1` to their gdev-loki.

image

If you are piggybacking this datasource in core grafana devenv `make devenv sources=loki`, you'll need to add additional changes to the provisioning files or just change the `read_only` column in the `data_source` table so you can add the header options manually.
```sql
UPDATE data_source SET read_only = 0 WHERE name = 'gdev-loki'
```
But the next time you run devenv it'll probably break again.
@matyax @L2D2Grafana ^^

@trevorwhitney Can we enable the tenant config endpoint? I'm seeing `Tenant configs not enabled` when I hit `http://localhost:3100/config/tenant/v1/limits`

trevorwhitney:
@gtk-grafana try again. I added an overrides file, the presence of which should make the limits API work but I haven't tested it locally yet as the was a port conflict I didn't want to deal with.

gtk-grafana:
@trevorwhitney nice! The tenant config endpoint seems to work when I add the `X-Scope-OrgID` header.

Although it looks like this PR breaks the otel services, seeing an error from alloy and the generator:
```
alloy-1          | {"ts":"2025-08-27T15:57:11.688889252Z","level":"error","msg":"final error sending batch","component_path":"/","component_id":"loki.write.default","component":"client","host":"host.docker.internal:3100","status":401,"tenant":"","error":"server returned HTTP status 401 Unauthorized (401): no org id"}
generator-1      | 2025-08-27T15:57:11.937Z	error	internal/queue_sender.go:52	Exporting failed. Dropping data.	{"resource": {"service.instance.id": "e09d2163-51bd-44fd-8731-c3e48a06473d", "service.name": "otelcol-contrib", "service.version": "0.133.0"}, "otelcol.component.id": "otlphttp/logs", "otelcol.component.kind": "exporter", "otelcol.signal": "logs", "error": "not retryable error: Permanent error: rpc error: code = Unauthenticated desc = error exporting items, request to http://loki:3100/otlp/v1/logs responded with HTTP Status Code 401", "dropped_items": 100}
```

trevorwhitney:
ahh, good call, I didn't check otel, that's probably a change to the otel configuration rather than the flag to the generator, I'll look into it.

gtk-grafana (dismissed):
LGTM! Thanks for taking a look at this!

matyax:
With zero context, why adding a tenant id? Is this a required setting, or is this adding any other capability?

gtk-grafana:
@trevorwhitney oh no, nvm, I missed an error, the alloy instance is failing to start:
```
 Error: /etc/alloy/config.alloy:9:25: missing ',' in field list

 8 |         headers = {
 9 |             "X-Scope-OrgID" = "1"
   |                                  
10 |         }

interrupt received
Error: could not perform the initial load successfully
Error: /etc/alloy/config.alloy:9:25: missing ',' in field list

 8 |         headers = {
 9 |             "X-Scope-OrgID" = "1"
   |                                  
10 |         }
interrupt received

Error: could not perform the initial load successfully
```

gtk-grafana:
@matyax this is pre-req work for the new [config ](https://github.com/grafana/loki/pull/19028) api endpoint in Loki.

Basically each Loki tenant maps to an org id in Grafana and can have separate configuration options, the per-tenant config endpoint returns those specific changes from the global loki configuration, and the new API endpoint will merge the global and tenant config so we get a consistent API response if tenant specific config exists or not.

matyax:
Ahhhh, fantastic! Thank you.

Issue #1199, opened 2025-04-17
Do a better job handling unsupported loki config and versions


gtk-grafana:
The app still works somewhat when patterns isn't configured, or with Loki versions that are not technically supported.

This is a constant source of confusion for users as the UI doesn't always imply what to change.

Ideally we could detect the required Loki configuration and version and provide simple upgrade steps, and remove portions of the UI that are not supported for their configuration.

This would probably require something like: https://github.com/grafana/loki/issues/16903, which would allow us to  get more information about how the current Loki is configured.

Issue #1166, opened 2025-04-03
Limit available quick time ranges related to available retention


jage:
Got multiple Loki data sources (Grafana Cloud), some with 30 day retention and one with 60 days retention.
Unfortunately there's no 60 days quick range option, and selecting a longer time range triggers a warning/error message:

<img width="644" alt="Image" src="https://github.com/user-attachments/assets/048a298e-da57-4d28-9c8f-603d5d5cf0c4" />

It would be great if the available time either was configurable, or automatically adapted according to Loki retention, e.g. if there's only 60 days of data then 60 days is the longest range I can select from the quick time ranges.

<img width="667" alt="Image" src="https://github.com/user-attachments/assets/c73b2356-5790-44d0-972a-7a1b37bf5422" />

jage:
@gtk-grafana thanks for working on this, I just learned there's probably more to this than just the `retention_period` and `max_query_lookback` in Loki. There seems to be some other limit to how long a query time range can be, that's still set to 30 days on our Grafana Cloud stack and I'm not sure it can be increased. (Ticket ##178203 for reference)

Just wanted to give you heads-up in case it could be an issue for others as well, but I guess for self-hosted Loki systems this is still very useful (which would be an option for us to the the longer retention).

gtk-grafana:
Yeah so the fix proposed in https://github.com/grafana/logs-drilldown/pull/1174 doesn't use the Loki settings unfortunately, that would require Loki implement an API endpoint that can tell the frontend the tenant configs (see https://github.com/grafana/loki/issues/16903 for more information), instead there is a plugin configuration for all queries in the Logs Drilldown app at `/plugins/grafana-lokiexplore-app` which is set by administrators that want to enforce a maximum time range.

<img width="1061" alt="Image" src="https://github.com/user-attachments/assets/37367839-23dc-4c86-a16c-4e410107748a" />

Issue #1138, opened 2025-03-25
Support level label in volume histogram when detected_level is disabled


gtk-grafana:
Support level as a label when `discover_log_levels: false`

gtk-grafana:
While changing the query is simple enough, supporting level as a label or as structured metadata is not as simple.

We'd need to know before sending a query to Loki if the level is structured metadata or a level or parsed field as to get the values for the levels variable, as structured metadata and labels use different API endpoints to return labels and label values.

If we knew the status of Loki's discover_log_levels config, we could potentially hide the "Log levels" variable input, and have users to use the labels combobox to query levels - assuming that we define that the level must be structured metadata or a label and not a parsed field, or even change the endpoints called by the Log Levels variable depending on loki config state, but this is not as simple as aggregating by multiple labels in the query.

Since Loki does not (currently) have an endpoint to expose loki config to the frontend, the short-term solution is to continue to assume that Loki instances have `discover_log_levels` enabled.

gtk-grafana:
With https://github.com/grafana/loki/issues/16903, we should have a path forward for the frontend to know when `discover_log_levels` is disabled, so this issue should be revisited if/when that is merged.

gtk-grafana:
So I started on a [PoC](https://github.com/grafana/logs-drilldown/compare/gtk-grafana/loki-tenant-config-api...gtk-grafana/issues/1138/config__discover-log-level-disabled) of this, and while it almost work it's a lot more complicated and hacky then we'd hoped:
* The multi-select detected_level variable is tied to query interpolation as metadata, scenes doesn't play nice with variables being conditionally interpolated based on application state
* I looked into simply hiding that variable for a quicker MVP, but since variables are rendered on the initial activation of the top-level scene, and the config is loaded async from an API (even if it's cached) we get flickering and variables moving around.
* The only solution I can think of is to prevent rendering of the entire app until we get a response from the config endpoint, which would make the app slower and more brittle for everyone.

For now I'm going to suggest that this is backlogged and we focus on easier wins from the limits_config, and come back to this later.

If we are able to get some instrumentation on /config timings in prod we will be able to make a better decision about waiting to render the entire app until we get a config response, but for now we should continue on features where the config is optional and loaded async

PR #1526, opened 2025-09-03
feat(config): Support Loki config API endpoint

gtk-grafana:
While this is ready for review, we don't want to release this until Loki 3.6 is available as earlier versions of Loki do not have the [new API endpoint](https://github.com/grafana/loki/pull/19028).

This PR doesn't add any functionality on top of the new `/drilldown-limits` Loki API endpoint exposed in https://github.com/grafana/loki/pull/19028, it just queries the endpoint and makes the config available throughout the application via the `IndexScene.lokiConfig` state. It is intended to be the parent branch of features that require Loki config, such as:
* https://github.com/grafana/logs-drilldown/issues/1199
* https://github.com/grafana/logs-drilldown/issues/1138
* https://github.com/grafana/logs-drilldown/issues/1166

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -310,6 +310,8 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     );
 
     this._subs.add(this.subscribeToLokiConfigAPI());
+    this._subs.add(this.subscribeToDataSourceChange());
+
     return () => {
       clearKeyBindings();
       assistantUnregister.forEach((callback) => callback.unregister());
@@ -347,6 +349,14 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     } else {
       return provideServiceBreakdownQuestions();
     }
+  }
+
+  private subscribeToDataSourceChange() {
+    getDataSourceVariable(this).subscribeToState((newState, prevState) => {
+      if (newState.value !== prevState.value) {
+        this.state.$lokiConfig.runQueries();
+      }
+    });
   }
 
   /**
--- a/src/services/datasource.ts
+++ b/src/services/datasource.ts
@@ -521,7 +521,9 @@ export class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
           // Cache for 1 day
           headers: {
             'X-Grafana-Cache': `private, max-age=86400`,
+            'X-Query-Tags': `Source=${PLUGIN_ID}`,
           },
+          requestId: ds.uid,
         }
       );
       const df = createDataFrame({


src/services/datasource.ts
```diff
@@ -497,6 +506,37 @@ export class WrappedLokiDatasource extends RuntimeDataSource<DataQuery> {
...
+          headers: {
+            'X-Grafana-Cache': `private, max-age=86400`,
+          },
+        }
+      );
+      const df = createDataFrame({
+        fields: [{ name: 'drilldown-limits', values: [config] }],
+      });
+
+      subscriber.next({ data: [df], state: LoadingState.Done });
```
L2D2Grafana, line 533: Do we want this in a dataFrame? The response didn't look like time series data to me https://github.com/grafana/loki/pull/19028/files#diff-d148bdd57e0460b8c36485f7fa3d2e04c4831f4df8191a5f4f720372fecc67e4R5

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -329,6 +333,40 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
...
+    return this.state.$lokiConfig.subscribeToState((newState, prevState) => {
+      // Loki versions before 3.6 will not have the new API endpoint, so we expect a 404 response
+      if (newState.data?.state === LoadingState.Error) {
+        this.setState({ lokiConfig: LOKI_CONFIG_API_NOT_SUPPORTED });
+        getMetadataService().setLokiConfig(LOKI_CONFIG_API_NOT_SUPPORTED);
+      } else if (newState.data?.state === LoadingState.Done && newState.data?.series.length > 0) {
+        const lokiConfig = newState.data?.series[0].fields[0].values[0];
+        if (lokiConfig) {
+          // we can't subscribe to metadata singleton like we can scene state, so we shouldn't pull config from singleton except to set the initial indexScene state
+          this.setState({ lokiConfig: newState.data?.series[0].fields[0].values[0] });
```
L2D2Grafana: Can't this just be this.setState({lokiConfig})

src/services/datasource.ts
L2D2Grafana: Oh maybe I don't understand because lokiConfig does look like the api response.

Image

L2D2Grafana (approved):
lgtm 🫡

src/services/datasource.ts
gtk-grafana: Scenes will only return dataframes from a query runner or the response gets stripped out, so with every call we make (patterns, detected_labels, detected_values, etc) we convert it to a dataframe and then pull it out again.

One of those annoying things we live with in order to use the runtime datasource

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: Yes it can and should!

src/services/metadata.ts
```diff
@@ -1,7 +1,9 @@
 import { ServiceSceneCustomState } from '../Components/ServiceScene/ServiceScene';
+import { LokiConfig, LokiConfigNotSupported } from './datasourceTypes';
 
 let metadataService: MetadataService;
 
+type LokiConfigState = LokiConfig | undefined | LokiConfigNotSupported;
```
matyax, line 6: This multi-type state seems a bit awkward. Can you consider make it LokiConfig | undefined and add a boolean in MetaDataService holding if it's supported or not?

matyax:
Changing the data source doesn't seem to be triggering a new request to the endpoint.

matyax (commented):
Other than the other two comments, LGTM!

src/services/metadata.ts
gtk-grafana: Look at the usages: https://github.com/grafana/logs-drilldown/pull/1621/files#diff-e3f3d606958f0611261996a826f7b8a8d3e3dd002ab13ebe7afc92ccb788690fR269

Application contexts need to differentiate if we haven't received a response, or it's not supported. We don't want to have to call the metadata every time we check the state

matyax: That explains why you went that route 👍

gtk-grafana:
> Changing the data source doesn't seem to be triggering a new request to the endpoint.

@matyax fixed in https://github.com/grafana/logs-drilldown/pull/1526/commits/f9af638eb038d6e361c350bfa22f4710002f1d0e

PR #1527, opened 2025-09-04
feat(logs): expose line limit in all visualizations

### LINKED ISSUE #1134: Logs Visualizations: Expose line limit to request more/less logs
Allow users to set line limit used in Logs Drilldown

matyax:
This PR introduces a new control to select the amount of log lines to display.

imagen

The criteria to select the component was to pick one that allowed to define a size that matches the height of the visualization type controls, and that supported to select one of multiple options. The Combobox was also a good option, but it's taller and looked out of place.

It's been implemented in a way that it reads the value set in the data source configuration (which can be the default 1000), and if it's an arbitrary number not present in the selector, it creates one for it. It's also stored and unique between services.

Demo:

https://github.com/user-attachments/assets/01c68c71-40c6-4809-93b9-f9d0eff66bb1

Final version using the combobox component:

imagen

Closes #1134

### DIFF
--- a/src/Components/ServiceScene/JSONLogsScene.tsx
+++ b/src/Components/ServiceScene/JSONLogsScene.tsx
@@ -8,7 +8,6 @@ import {
   SceneObjectState,
   SceneObjectUrlSyncConfig,
   SceneObjectUrlValues,
-  SceneQueryRunner,
 } from '@grafana/scenes';
 
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from '../../services/analytics';
@@ -26,6 +25,7 @@ import { NoMatchingLabelsScene } from './Breakdowns/NoMatchingLabelsScene';
 import { getDetectedFieldsFrameFromQueryRunnerState, ServiceScene } from './ServiceScene';
 import { KeyPath } from '@gtk-grafana/react-json-tree';
 import { logger } from 'services/logger';
+import { runSceneQueries } from 'services/query';
 import {
   getBooleanLogOption,
   getJSONHighlightState,
@@ -208,12 +208,7 @@ export class JSONLogsScene extends SceneObjectBase<JSONLogsSceneState> {
       return;
     }
     setLogOption('sortOrder', newOrder);
-    const $data = sceneGraph.getData(this);
-    const queryRunner =
-      $data instanceof SceneQueryRunner ? $data : sceneGraph.findDescendents($data, SceneQueryRunner)[0];
-    if (queryRunner) {
-      queryRunner.runQueries();
-    }
+    runSceneQueries(this);
     this.setState({ sortOrder: newOrder });
   };
 
--- a/src/Components/ServiceScene/ServiceScene.tsx
+++ b/src/Components/ServiceScene/ServiceScene.tsx
@@ -20,7 +20,7 @@ import {
   SceneVariableValueChangedEvent,
   VariableDependencyConfig,
 } from '@grafana/scenes';
-import { VariableHide } from '@grafana/schema';
+import { LogsSortOrder, VariableHide } from '@grafana/schema';
 import { LoadingPlaceholder } from '@grafana/ui';
 
 import { plugin } from '../../module';
@@ -55,10 +55,13 @@ import { LEVELS_VARIABLE_SCENE_KEY, LevelsVariableScene } from '../IndexScene/Le
 import { ShowLogsButtonScene } from '../IndexScene/ShowLogsButtonScene';
 import { ActionBarScene } from './ActionBarScene';
 import { breakdownViewsDefinitions, valueBreakdownViews } from './BreakdownViews';
+import { getLogsPanelSortOrderFromURL } from './LogOptionsScene';
 import { LogsListScene } from './LogsListScene';
 import { drilldownLabelUrlKey, pageSlugUrlKey } from './ServiceSceneConstants';
+import { LokiQueryDirection } from 'services/lokiQuery';
 import { getQueryRunner, getResourceQueryRunner } from 'services/panel';
 import { buildDataQuery, buildResourceQuery } from 'services/query';
+import { getLogOption, getMaxLines } from 'services/store';
 import {
   DETECTED_FIELD_VALUES_EXPR,
   EMPTY_VARIABLE_VALUE,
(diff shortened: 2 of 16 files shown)


gtk-grafana (commented):
This allows me to exceed the max lines limit set in the datasource, also the datasource limit should probably be set as the default if it exists?

Also I don't love adding more options to the panel header, but not sure if the log options sidebar is the right place for this either, although we did move the log direction in that menu. Might be worth discussing with UX?

src/Components/ServiceScene/JSONLogsScene.tsx
```diff
@@ -208,12 +208,7 @@ export class JSONLogsScene extends SceneObjectBase<JSONLogsSceneState> {
       return;
     }
     setLogOption('sortOrder', newOrder);
-    const $data = sceneGraph.getData(this);
-    const queryRunner =
-      $data instanceof SceneQueryRunner ? $data : sceneGraph.findDescendents($data, SceneQueryRunner)[0];
-    if (queryRunner) {
-      queryRunner.runQueries();
-    }
+    runSceneQueries(this);
```
gtk-grafana, line 211: Question: If this is the same query runner across all three viz can we move this up to the LogsListScene?

No need to block this PR though, this is already cleaner then what we've got.

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -159,7 +159,7 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     >
   ) {
     super({
-      $data: getServiceSceneQueryRunner(),
+      $data: undefined,
```
gtk-grafana, line 165: A tad confused as to why this change is made?

src/Components/ServiceScene/JSONLogsScene.tsx
gtk-grafana: oh wait, nvm, I thought we were setting the query runner here but we're just running the query, it's already on the serviceScene 🤦

src/services/panel.ts
```diff
@@ -282,7 +282,11 @@ export function getResourceQueryRunner(queries: LokiQuery[], queryRunnerOptions?
   });
 }
 
-export function getQueryRunner(queries: LokiQuery[], queryRunnerOptions?: Partial<QueryRunnerState>) {
+export function getQueryRunner(
```
gtk-grafana: Nit: maybe worth splitting `getQueryRunner` into getLogsPanelQueryRunner and getPanelQueryRunner so we can make sceneRef required?

Or if we still need a generic handler maybe it's time to take a single object as an argument.

src/Components/ServiceScene/ServiceScene.tsx
matyax: Definitely. Based on https://github.com/grafana/logs-drilldown/blob/main/src/Components/ServiceScene/ActionBarScene.tsx#L35 and the fact that the data source itself can have a custom or default line limit, combined with `getMaxLines` having to generate an "exploration prefix" for the stored value, `getMaxLines` receives a sceneRef which doesn't exist on the constructor.

https://github.com/grafana/logs-drilldown/blob/feb61e3f363d6883c0cb9cc08d8535e4eb3a4af5/src/services/store.ts#L216

So, to create a query runner for logs, it needs to read `maxLines`, and have a sceneRef to pass around.

src/services/panel.ts
gtk-grafana: Actually while this might introduce some refactor bloat to this PR, this method returning different max lines functionality depending on the existence of an optional sceneRef is a bit of a foot gun for future devs.

Although reverting these changes to getQueryRunner (and moving the code in the else clause) and creating a new method which is (only?) used in the serviceScene isn't that big of a change and should be much clearer for future devs. Also would always return a `SceneQueryRunner` and not a `SceneDataTransformer` which would mean less narrowing with logs panel runners

matyax:
> This allows me to exceed the max lines limit set in the datasource, also the datasource limit should probably be set as the default if it exists?

It's not a limit, but a default value. The Loki query editor allows for any arbitrary value:

imagen

> Also I don't love adding more options to the panel header, but not sure if the log options sidebar is the right place for this either, although we did move the log direction in that menu (since those options are clustered in explore classic it makes sense to keep them in the same UI?). Might be worth discussing with UX?

It's one more UI element indeed, but I don't see this going to the log controls. This is a query setting, which the controls don't have reach, and we need to account for the other visualizations.

I think it's worse to not have this option at all, but we can explore other alternatives.

gtk-grafana:
> It's not a limit, but a default value. The Loki query editor allows for any arbitrary value:


Oh snap, my bad then. That option is not named very well in the ds config:
image


> It's one more UI element indeed, but I don't see this going to the log controls. This is a query setting, which the controls don't have reach, and we need to account for the other visualizations.
> 
> I think it's worse to not have this option at all, but we can explore other alternatives.

Hmmmm, maybe moving it above the panel (inline with log text search) would work better, and maybe even simplify the implementation as it won't have to get added to each panel? 

image

matyax:
I like that! Let's give it a try.

src/Components/Table/LogsHeaderActions.tsx
```diff
@@ -14,10 +18,46 @@ import { LogsVisualizationType } from 'services/store';
  */
 export function LogsPanelHeaderActions(props: {
   onChange: (type: LogsVisualizationType) => void;
+  sceneRef: SceneObject;
   vizType: LogsVisualizationType;
 }) {
+  const [maxLines, setMaxLinesState] = useState(getMaxLines(props.sceneRef));
```
L2D2Grafana: I'll have to remember this way of mixing scenes with state.

src/services/query.ts
```diff
@@ -243,3 +243,12 @@ export function getLineFormatExpressionBuilder() {
...
 export const LINE_LIMIT = 1000;
+
+export function runSceneQueries(sceneRef: SceneObject) {
+  const $data = sceneGraph.getData(sceneRef);
+  const queryRunner =
+    $data instanceof SceneQueryRunner ? $data : sceneGraph.findDescendents($data, SceneQueryRunner)[0];
+  if (queryRunner) {
+    queryRunner.runQueries();
+  }
+}
```
L2D2Grafana, line 254: Nice!

gtk-grafana (approved):
LGTM, works well locally, if you want to merge this and create a follow up issue for the debt in the query runner handler that should be fine.

src/services/panel.ts
matyax: I now see why you pointed this out. I extracted everything that was exclusive for the logs panel query into  `getLogsQueryQueryRunner()` (formerly `getServiceSceneQueryRunner()`), so it doesn't need to know about any scene references or logs-query-related stuff.

matyax:
@gtk-grafana Made a small refactor to address the tech debt added to `getQueryRunner`. Wanna take a second look?

src/services/routing.ts
```diff
@@ -134,7 +134,7 @@ export function getPrimaryLabelFromUrl(): RouteProps {
   const routeParams = endOfUrl.split('/');
 
   let labelName = routeParams[0];
-  const labelValue = routeParams[1];
+  const labelValue = decodeURIComponent(routeParams[1]);
```
matyax, line 137: There was a running condition where, for a moment, when you're building the "exploration prefix" based on route params, you could get a raw an encoded string:

imagen

Very niche running condition, surfaced by local Loki and `{service_name="\"grafana/data\""}`

gtk-grafana:
> @gtk-grafana Made a small refactor to address the tech debt added to `getQueryRunner`. Wanna take a second look?

@matyax  Nailed it!

src/Components/Table/LogsHeaderActions.tsx
matyax: This was a pure react component, which can be rendered in the component part of a Scene. Then, you can think about scenes like a class instance holding state and methods, with a component property that renders a functional component.

Issue #1462, opened 2025-07-28
[FEAT]: Table preferences


L2D2Grafana:
**Description of the new feature**
Add any missing table configs into localstorage, url in Explore and Logs Drilldown.

- Sync displayedFields with urlColumns
- Sort order: once selected, should be kept between changes in the current query or updates
- Permalinks: [Unify url params between table + panel: use `panelState` for linking](https://github.com/grafana/logs-drilldown/issues/857)
  https://github.com/grafana/grafana/blob/main/public/app/plugins/panel/logs/LogsPanel.tsx#L759-L776
- Better defaults:
  - [ ] Include detected_level (or level if not present) as a displayed field
  For now: if both are present, show detected_level, otherwise level.
  Figure out why level is not colored.
  - [ ] Show the log line body instead of the fields by default

**Acceptance Criteria**
- [x] All functionality preserved
- [x] Sort is respected after updates
- [x] Table display settings are stored and reused when returning to a service
- [x] Features exist in Logs Drilldown
- [x] Features exist in core

**New Requirements**
- Core table: update panelstate columns to be an array to urlColumns etc
- Core and Drilldown selected fields: sync: time and line between core and LD
- Core and Drilldown share the same displayedFields

L2D2Grafana:
Now there's a further improvement that we could do here and core regarding columns and displayed fields.

Table to Logs:
If the displayed columns are the default columns and we switch to the logs visualization, no displayed fields should be selected
If the displayed columns are not the default columns and we switch to the logs visualization:
The same columns should be set as displayed fields, except timestamp
If body is shown in the table, the log line should be shown in the Logs panel


Stretch goal:

Logs to table:
If there are no displayed fields, the default columns should be used ✅
Except, if show timestamp is disabled, it could not be included in the default columns
If there are displayed fields, they should be displayed as columns ✅
If the log line is a not a displayed field, body should not be shown as a column

* We could have gotten this for free if we used displayedFields in both visualizations, and if we shared the internal name of the log line between the two viz, i think.

L2D2Grafana:
We also discussed moving the field selector sidebar up a level and to be of the left of the logspanel and table for a more wholistic experience ⭐

L2D2Grafana:
This was completed for logs drilldown and core, but the refactor to using on Displayed fields did not work out. The table will become a panel in core instead.

PR #1534, opened 2025-09-10
feat(table): preferences

L2D2Grafana:
- #1462 
- Sort order: once selected, should be kept between changes in the current query or updates
- Include detected_level (or level if not present) as a displayed field. Figure out why level is not colored.
- Show the log line body instead of the fields by default

https://github.com/user-attachments/assets/c617b7d3-96b1-4c90-9b6a-8d6dd93b91ae

# Todo 
- e2e tests not running locally, I think we used to use the data from gdev-loki and tempo-ingester and my gdev-loki no longer has that data.
- permalinks: This might be ticketed out


src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -102,6 +102,18 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
       })
     );
 
+    // Subscribe to data changes to update URL columns
+    const dataProvider = sceneGraph.getData(this);
+    this._subs.add(
+      dataProvider.subscribeToState((newState, prevState) => {
+        if (newState.data && newState.data !== prevState.data) {
```
gtk-grafana: We can combine these two conditionals, also `newState.data !== prevState.data` can be truthy even if the data is the same, try `!areArraysEqual(prevState?.data?.series, newState?.data?.series)` instead if we want this to only run when the dataframes have changed

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -146,19 +163,89 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
...
-    defaultUrlColumns = urlColumnsUrl
-      ? this.urlHasDefaultUrlColumns(urlColumnsUrl)
-        ? this.updateDefaultUrlColumns(urlColumnsUrl)
-        : defaultUrlColumns
-      : defaultUrlColumns;
-    defaultUrlColumns = defaultUrlColumns.length > 0 ? defaultUrlColumns : defaultUrlColumns;
+    let newUrlColumns = urlColumnsUrl;
+
+    // if urlColumns is populated in the url
+    if (urlColumnsUrl && urlColumnsUrl.length > 0) {
```
gtk-grafana: let's combine these conditionals as well

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -146,19 +163,89 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
...
+    const dataProvider = sceneGraph.getData(this);
+    const data = dataProvider.state.data;
+    if (!data || !data.series || data.series.length === 0) {
+      return null;
+    }
+
+    // Check if any of the data frames have a detected_level or level field with actual values
+    for (const frame of data.series) {
+      // Find the labels field which contains the log metadata
+      const labelsField = frame.fields.find((field) => field.name === 'labels' && field.values);
```
gtk-grafana: there's a `getLabelsFromSeries` method in labels.ts that should help clean this up a bit.

src/Components/Table/DefaultPill.tsx
```diff
@@ -67,7 +67,7 @@ export const DefaultPill = (props: DefaultPillProps) => {
   const { cellIndex } = useTableCellContext();
   let levelColor;
 
-  if (label === LEVEL_NAME) {
+  if (label === DETECTED_LEVEL || label === LEVEL) {
```
gtk-grafana: we make this some check in a few places, let's create a helper method for this

gtk-grafana:
Let's store the user's line text/labels visibility selection in local storage in addition to the url state. Ideally url state should override the local storage setting, but local storage is only updated by user events (i.e. clicking on the toggle).

gtk-grafana (commented):
Body column order changes on refresh:
image
Refresh
image

Same with timestamp, but other fields seem to stay in their same order

gtk-grafana:
A few nits & bugs but great work on this, this is a really nice enhancement to the UX

src/Components/ServiceScene/LogsTableScene.tsx
L2D2Grafana: Unfortnately that did not work because the labels were not in the `frame.fields[1]?.labels;` position. I did make another helper function that looks for labels regardless of index.

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -149,19 +167,79 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
...
   };
 
+  // check if url has default columns initially there are none so we need to keep default values
+  urlHasDefaultUrlColumns = (urlColumns: string[]) => {
+    return defaultUrlColumns.some((col) => urlColumns.includes(col));
+  };
+
+  // update defaultUrlColumns and match order
+  updateDefaultUrlColumns = (urlColumns: string[]) => {
+    defaultUrlColumns = defaultUrlColumns.reduce<string[]>((acc, col) => {
```
matyax, line 196: Minor and unrelated, but because DEFAULT_URL_COLUMNS is being imported and reassigned to a local const called `defaultUrlColumns`, I had to locally open the file to understand where it was coming from and what was it.

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -149,19 +167,79 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
...
+      // return the column in the same index position as urlColumns
+      if (urlColumns.includes(col)) {
+        const urlIndex = urlColumns.indexOf(col);
+        acc[urlIndex] = col;
+      }
+      return acc;
+    }, []);
```
matyax, line 203: I wonder if you should use a filter here or other combination of methods, because this will result in an array with empty entries:

imagen

src/Components/Table/constants.ts
```diff
@@ -1,2 +1,3 @@
-export const LEVEL_NAME = 'detected_level';
+export const DETECTED_LEVEL = 'detected_level';
```
matyax, line 1: 👌

src/Components/Table/DefaultPill.tsx
```diff
@@ -67,7 +67,7 @@ export const DefaultPill = (props: DefaultPillProps) => {
   const { cellIndex } = useTableCellContext();
   let levelColor;
 
-  if (label === LEVEL_NAME) {
+  if (hasLabelLevel(label)) {
```
matyax:
```suggestion
  if (isLabelLevel(label)) {
```
Maybe.

src/services/labels.ts
```diff
@@ -41,6 +42,25 @@ export function buildLabelsQuery(sceneRef: SceneObject, optionValue: string, opt
   );
 }
 
+export function getAllLabelsFromDF(series: DataFrame[]): string[] {
```
matyax:
```suggestion
export function getAllLabelsFromDataFrame(series: DataFrame[]): string[] {
```

```suggestion
export function getAllLabelsFromSeries(series: DataFrame[]): string[] {
```

src/Components/ServiceScene/LogsTableScene.tsx
L2D2Grafana: Updated to use the constant.

src/Components/Table/DefaultPill.tsx
L2D2Grafana: updated

src/services/labels.ts
L2D2Grafana: updated

L2D2Grafana:
> A few nits & bugs but great work on this, this is a really nice enhancement to the UX

Fixed the changing order and refresh 🐛

L2D2Grafana:
> Looking much better!
> 
> Question: is if I click on "show original log line" in the logs panel, do we expect the table to revert back to the original columns as well? Currently it seems to un-wed the logs and table panels until a change is made in the table.
> 
> Related: if I add a displayed field in the logs panel, do we expect it to update the columns in the table? ~Right now the behavior is that changes in the table are synced to the logs panel, but not the other way around?~
> 
> Actually the above is not correct, logs panel does not sync with table, but adding new fields to logs panel adds new fields to table:
> 
> For example, if I reset columns in the logs panel, and then visualize the duration field:
> 
> image
> Duration is added to whatever columns I have defined in the table: image
> 
> I think the bug here is the "show original log line" not updating table columns, and un-visualizing/removing a field in the logs panel does not remove a column from the table

Nice find! I had not integrated anything with 'show original log line', updated now 👍

matyax:
It all seems to be working flawlessly 👌 . Now there's a further improvement that we could do here and core regarding columns and displayed fields.

Table to Logs:
- If the displayed columns are the default columns and we switch to the logs visualization, no displayed fields should be selected
- If the displayed columns are not the default columns and we switch to the logs visualization:
  - The same columns should be set as displayed fields, except timestamp
  - If `body` is shown in the table, the log line should be shown in the Logs panel

https://github.com/user-attachments/assets/457735bf-ae08-4984-b223-f8885deb74bf

Stretch goal:

Logs to table:
- If there are no displayed fields, the default columns should be used ✅
  - Except, if show timestamp is disabled, it could not be included in the default columns
- If there are displayed fields, they should be displayed as columns ✅ 
- If the log line is a not a displayed field, body should not be shown as a column

We could have gotten this for free if we used `displayedFields` in both visualizations, and if we shared the internal name of the log line between the two viz, i think.

Feel free to address any of that in a follow up PR, this one LGTM.

Issue #1535, opened 2025-09-15
Line limit: add status component when requesting more than max_entries_limit


matyax:
We recently added support for requesting an arbitrary amount of logs (line limit/max lines). We need to follow this up with a clear error message when the requested amount is above the configured limit.

<img width="1151" height="517" alt="Image" src="https://github.com/user-attachments/assets/3bac13be-6e2b-4a61-a280-b90b5680abec" />

L2D2Grafana:
- validation on the input
- update error message

PR #1537, opened 2025-09-15
feat(LineLimitScene): error message, invalidation state

L2D2Grafana:
- #1535 
- Updated error copy
- Added invalidation border
Screenshot 2025-09-15 at 8 35 51 AM

### DIFF
--- a/src/Components/ServiceScene/LogsListScene.tsx
+++ b/src/Components/ServiceScene/LogsListScene.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { css } from '@emotion/css';
 
+import { LoadingState, PanelData } from '@grafana/data';
 import { locationService } from '@grafana/runtime';
 import {
   SceneComponentProps,
@@ -19,6 +20,8 @@ import { Options } from '@grafana/schema/dist/esm/raw/composable/logs/panelcfg/x
 import { reportAppInteraction, USER_EVENTS_ACTIONS, USER_EVENTS_PAGES } from '../../services/analytics';
 import { logger } from '../../services/logger';
 import { narrowLogsVisualizationType, narrowSelectedTableRow, unknownToStrings } from '../../services/narrowing';
+import { getVariablesThatCanBeCleared } from '../../services/variableHelpers';
+import { IndexScene } from '../IndexScene/IndexScene';
 import { LogLineState } from '../Table/Context/TableColumnsContext';
 import { SelectedTableRow } from '../Table/LogLineCellComponent';
 import { ActionBarScene } from './ActionBarScene';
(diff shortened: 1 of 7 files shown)


src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -210,6 +210,8 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
     } else if (errorResponse?.includes('response larger than the max message size')) {
       errorMessage =
         'The response is too large to process. Try narrowing your search or using filters to reduce the data size.';
+    } else if (errorResponse?.includes('max entries limit')) {
+      errorMessage = 'Max entries limit per query exceeded. Please review your Line limit setting and try again.';
```
gtk-grafana:
```suggestion
      errorMessage = 'Max entries limit per query exceeded. Please review your "Line limit" setting and try again.';
```

gtk-grafana:
Overall looks good, only weirdness is with Table/JSON panel which don't seem to have the nice error state like the logs panel.

image

This can also be reproduced in the logs panel if you run the query in the table or json and then switch to logs.

L2D2Grafana:
> ('max entries limit')



> Overall looks good, only weirdness is with Table/JSON panel which don't seem to have the nice error state like the logs panel.
> 
> image
> This can also be reproduced in the logs panel if you run the query in the table or json and then switch to logs.

This required a bit of a refactor, let me know if I went too far 😈

L2D2Grafana:
Not sure why the docs docker registry are throwing 500s 
Screenshot 2025-09-15 at 12 16 00 PM

gtk-grafana:
> Not sure why the docs docker registry are throwing 500s

There's a way to spin Grafana up with docs support, but I don't think we do that by default, probably not anything we need to concern ourselves with since that error is coming from core.

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -170,6 +180,85 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
         }
       })
     );
+
+    // Subscribe to ServiceScene data changes for error handling (all visualization types)
+    const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
+    this._subs.add(
+      serviceScene.subscribeToState((newState, prevState) => {
```
gtk-grafana: I don't think this is working as intended, still seeing the empty panels unless I refresh the page

image

gtk-grafana: ahhh I think it's because we're subscribing to the serviceScene state instead of the state of the query provider

L2D2Grafana: Did you pull the updates it's updating without refresh for me

https://github.com/user-attachments/assets/cf801535-7ae6-4709-87c0-2680139b7cad

matyax:
Not sure if this was mentioned before, but it seems that the error state is sticking. I don't have a consistent way to repro, just go into an error state and change it to a limit that works.

imagen

Also, in general, when valid custom values are used:

imagen

src/Components/ServiceScene/LogsListScene.tsx
gtk-grafana: Having a harder time reproducing now, but we should still be subscribing to the query provider instead of the serviceScene state, if the state of the query provider changes state without updating the state of the service scene this subscription won't run. This could lead to unintended consequences down the line which are no fun to debug. 

Take a look at `subscribeToLogsQuery` in the ServiceScene for an example, but I think it's as simple as changing this code from `serviceScene.subscribeToState` to `serviceScene.state.$data.subscribeToState` and updating the internals `newState.$data.state.data` => `newState.data`

L2D2Grafana: updated!

L2D2Grafana:
> Not sure if this was mentioned before, but it seems that the error state is sticking. I don't have a consistent way to repro, just go into an error state and change it to a limit that works.
> 
> imagen
> Also, in general, when valid custom values are used:
> 
> imagen

Apparently this was because the custom input was a string.

PR #1540, opened 2025-09-17
Fix(volume): Field config getting overwritten

gtk-grafana:
Fixes bug in logs volume panel where changes to the panel builders were getting overwritten by the `setLogsVolumeFieldConfigOverrides`.

This PR: 
Only sets fieldConfig.overrides instead of completely nuking the config

To test:
I think the only fieldConfig we didn't have duplicated was the short unit.

Notes:
Added some comments but there is still some debt here to pay down. Ideally we could deep compare the overrides and add new ones that don't exist while leaving the existing overrides in case someone adds more overrides in the future.

### DIFF
--- a/src/services/panel.ts
+++ b/src/services/panel.ts
@@ -124,23 +124,25 @@ export function setLabelSeriesOverrides(labels: string[], overrideConfig: FieldC
 
 /**
  * Sets labels series visibility in the panel
+ * WARNING: Overwrites any fieldConfig overrides set in the panel builder!
  */
 export function syncLevelsVisibleSeries(panel: VizPanel, series: DataFrame[], sceneRef: SceneObject) {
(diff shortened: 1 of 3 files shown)


matyax (approved):
Amazing to finally see this fixed!

imagen

src/services/panel.ts
```diff
@@ -298,7 +296,7 @@ export function getQueryRunner(queries: LokiQuery[], queryRunnerOptions?: Partia
         queries: queries,
         ...queryRunnerOptions,
       }),
-      transformations: [sortLevelTransformation],
+      transformations: [],
```
gtk-grafana, line 299: removing `sortLevelTransformation` introduced https://github.com/grafana/logs-drilldown/issues/1622

Issue #1495, opened 2025-08-08
[FEAT]: Support to expand/collapse the header filters


matyax:
Frequently requested option to collapse the header filters when more space is needed.

matyax:
This is still a frequently requested change.

PR #1541, opened 2025-09-18
feat(VariableLayoutScene): add control to expand and collapse header filters

matyax:
Frequent feedback.

Closes [#1495](https://github.com/grafana/logs-drilldown/issues/1495)

https://github.com/user-attachments/assets/51a4807b-4dcc-4b17-a652-c5039f285051

### DIFF
--- a/src/Components/IndexScene/VariableLayoutScene.tsx
+++ b/src/Components/IndexScene/VariableLayoutScene.tsx
@@ -3,11 +3,16 @@ import React from 'react';
 import { css, cx } from '@emotion/css';
 
 import { GrafanaTheme2 } from '@grafana/data';
-import { useChromeHeaderHeight } from '@grafana/runtime';
+import { t } from '@grafana/i18n';
+import { reportInteraction, useChromeHeaderHeight } from '@grafana/runtime';
 import { SceneComponentProps, SceneFlexLayout, sceneGraph, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
-import { useStyles2 } from '@grafana/ui';
+import { ToolbarButton, useStyles2 } from '@grafana/ui';
 
-import { getJsonParserVariableVisibility } from '../../services/store';
+import {
+  getCollapsibleFiltersState,
+  getJsonParserVariableVisibility,
+  setCollapsibleFiltersState,
+} from '../../services/store';
 import { AppliedPattern } from '../../services/variables';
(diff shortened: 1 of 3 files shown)


matyax:
What do you think? Probably worth adding a more prominent active state so it can be easily disabled if needed.

src/Components/IndexScene/VariableLayoutScene.tsx
```diff
@@ -89,13 +107,24 @@ export class VariableLayoutScene extends SceneObjectBase<VariableLayoutSceneStat
...
+                    }
+                  />
                 </div>
               </div>
             </div>
           )}
 
           {/* 2nd row - Combined fields (fields + metadata) + Levels - custom renderer */}
-          <div className={styles.controlsRowContainer}>
+          <div className={styles.controlsRowContainer} style={collapsed ? { display: 'none ' } : undefined}>
```
matyax: Only for demo purposes.

gtk-grafana:
Yeah I did a PoC of this once earlier, biggest concern was that folks might click that close button when they don't have any filters added and then not know how to remove filters/patterns etc

matyax:
With toolbar active state:

imagen

I like both the approach and how it looks. One thing we could try is a summarized view of the applied filters.

gtk-grafana:
Maybe on the left closer to the filters instead of the right?

matyax:
Moved to the left!

L2D2Grafana (approved):
LGTM 👏 

> [!NOTE] 
There are two other different expand/collapse icons in the codebase

Image

Image

In small the top filters go full-width then align right if you go even smaller. Full-width seems nicer.

Image

Image

matyax:
Yeah, good points. I specifically wanted to avoid using, which is for dropdowns.

imagen

We could maybe use the other expand/collapse and rotate them?

matyax:
It's going to be a challenge because we don't control the icons for those ToolbarButtons, and those are the only buttons with that active style.

L2D2Grafana:
> Yeah, good points. I specifically wanted to avoid using, which is for dropdowns.
> 
> imagen
> We could maybe use the other expand/collapse and rotate them?

I feel like I should at least update the table column sidebar to use the expand/collapse icon in this PR.

matyax:
Great call @L2D2Grafana ! We were definitely using 3 different icons for the same action. They are now unified.

imagen

imagen

imagen

In the process I created https://github.com/grafana/logs-drilldown/issues/1545 for the sidebar, because it looks very out of place in the collapsed state.

PR #1543, opened 2025-09-18
feat: Asserts insight timeline widget integration

gtk-grafana:
Adds `grafana-asserts-app/insights-timeline-widget/v1` to the logs volume panel container.

image

Currently passes the first value from the filters with a `service_name` label and inclusive operator.

This feature/externalized component is built for demonstration purposes, and is intended to be replaced by multi-lane annotations when that feature is available in the core viz.


src/Components/ServiceScene/LogsVolume/LogsVolumeContainerScene.tsx
```diff
@@ -0,0 +1,76 @@
...
+    this.addActivationHandler(this.onActivate.bind(this));
+  }
+
+  private onActivate() {
+    this.setState({
+      body: new SceneFlexLayout({
+        direction: 'column',
+        children: [
+          new SceneFlexItem({
+            body: new LogsVolumePanel({}),
```
matyax, line 34: Super minor, but if we could build these without passing an empty object, :cheffkiss:

PR #1547, opened 2025-09-24
feat(embedded): remove read-only label filters

gtk-grafana:
Embedded label filters in logql expression will no longer be read-only in `EmbeddedLogsExploration` component.

PR #1549, opened 2025-09-24
feat(embedded): allow resetting filters

ifrost:
Allows removing/changing/resetting (previously manged) filters in Embedded mode. This will also be used to switch over to Logs tab when a data link is clicked in Asserts:

https://github.com/user-attachments/assets/d4e2df49-4d48-470f-8508-a054070921a1

All PRs related to this change:

- [ ] https://github.com/grafana/logs-drilldown/pull/1549
- [ ] https://github.com/grafana/grafana/pull/111304
- [ ] https://github.com/grafana/asserts-app-plugin/pull/2150


gtk-grafana:
@ifrost I made the change to remove the readonly filters in https://github.com/grafana/logs-drilldown/pull/1547 as well, do you have a preference if we address that change in this PR or not?

ifrost:
> @ifrost I made the change to remove the readonly filters in #1547 as well, do you have a preference if we address that change in this PR or not?

Oh, nice! Feel free to merge the other one first if it can speed things up for you. I've additionally:
- removed some (hopefully) dead code, 
- added a reset button (used by https://github.com/grafana/asserts-app-plugin/pull/2150),
- added handling for the case where user removes all filters (there's no service selection in embedded view and it would throw an error). 
- added ability to pass custom namespace for https://github.com/grafana/asserts-app-plugin/pull/2150 (though can hardcode `ld` as well)

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -300,6 +305,13 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     };
   }
 
+  public currentFiltersMatchReference() {
+    return (
+      !this.state.referenceLabels ||
+      areLabelFiltersEqual(this.state.referenceLabels || [], getLabelsVariable(this).state.filters)
```
gtk-grafana, line 312: would `areArraysEqual` (order doesn't matter) or `areArraysStrictlyEqual` (order matters) work instead of adding a new method?

ifrost: The problem is that there are some extra props that are not present when filters are loaded initially but are added when users changes filters manually (e.g. `condition`, `keyLabel`, ...). For example:

Initial filters:

Screenshot 2025-09-25 at 09 35 40

On the left reference labels passed from Asserts. On the right labels after removing and adding back namespace=default (so for the user they are the same, but `areArraysEqual` would return false:

Screenshot 2025-09-25 at 09 34 18

gtk-grafana: Hmm good catch, what if we mapped the filters to only check key, operator, and value with `areArraysEqual` instead of supporting a new compare method?

Something like:
```
export function areLabelFiltersEqual(a: AdHocVariableFilter[], b: AdHocVariableFilter[]) {
  const mapAdHocFilters = (a: AdHocVariableFilter) => ({ key: a.key, operator: a.operator, value: a.value });
  return areArraysEqual(a.map(mapAdHocFilters), b.map(mapAdHocFilters));
}
```

ifrost: Nice! That should do the trick and be simpler 👍

ifrost: Updated in ee989448770718e98302aeeca16191b7e207aa2b

gtk-grafana:
@ifrost I'm seeing that the reset filters button always removes all filters and pushes to the `No labels selected` state when testing locally. Does this require changes to the props passed in from embedding application? If so let's update the [testing route](https://github.com/grafana/logs-drilldown/blob/main/src/Components/Pages.tsx#L76-L112) so this PR can be tested without external deps

src/services/variableHelpers.ts
```diff
@@ -1,38 +1,24 @@
...
-                readonlyFilter.key === filter.key &&
-                readonlyFilter.value === filter.value &&
-                readonlyFilter.operator === filter.operator
-            )
-        )
-      ) {
-        variablesToClear.push(variable);
-      }
-    } else if (variable instanceof AdHocFiltersVariable && variable.state.filters.length) {
+    if (variable instanceof AdHocFiltersVariable && variable.state.filters.length) {
```
gtk-grafana, line 21: @ifrost I think this change is what's causing the initial labels to get removed (and most likely also a bug in my other branch removing the read only filters), we were relying on the readonly variable to let us know which filters should get left alone when clearing.

ifrost:
> @ifrost I'm seeing that the reset filters button always removes all filters and pushes to the `No labels selected` state when testing locally. Does this require changes to the props passed in from embedding application? If so let's update the [testing route](https://github.com/grafana/logs-drilldown/blob/main/src/Components/Pages.tsx#L76-L112) so this PR can be tested without external deps

Oh good spot. I added a check that if reference labels are not passed at all the reset button is not shown (as there's nothing to reset to), but it was checking only if value was defined but it can be an empty array. I've changed the check in 6e1d54cb3d0f039e5e445d51b7b0d3f59227caa2

ifrost:
Now it should not require external deps - with Asserts from main it will just not show the reset button at all. If used with https://github.com/grafana/asserts-app-plugin/pull/2150 the reference labels will be passed from Asserts and the reset button will show up.

gtk-grafana:
@ifrost I added the referenceQuery to the testing route props in https://github.com/grafana/logs-drilldown/pull/1549/commits/33cd4b0ac00ab92c088d9d289acc4465bbbb6128. The testing route can be hit at `/a/grafana-lokiexplore-app/embed`

By external deps I mean PRs in external repos, the testing route is supposed to save the rest of the squad from needing to spin up Logs Drilldown in an external app, and to provide us something to run e2e tests against that should provide the same behavior as within an embedded app (e.g. Asserts).

src/Components/EmbeddedLogsExploration/EmbeddedLogs.tsx
```diff
@@ -34,13 +35,20 @@ export function buildLogsExplorationFromState({
   initRuntimeDs();
 
   const { labelFilters, lineFilters } = getMatcherFromQuery(query);
+  const referenceFilters = getMatcherFromQuery(referenceQuery || '');
```
gtk-grafana:
```suggestion
  const referenceFilters = getMatcherFromQuery(referenceQuery ?? '');
```

src/Components/EmbeddedLogsExploration/EmbeddedLogs.tsx
```diff
@@ -74,7 +83,7 @@ export default function EmbeddedLogsExploration(props: EmbeddedLogsExplorationPr
       scene={exploration}
       updateUrlOnInit={false}
       createBrowserHistorySteps={true}
-      namespace={VARIABLE_NAMESPACE}
+      namespace={props.namespace || VARIABLE_NAMESPACE}
```
gtk-grafana:
```suggestion
      namespace={props.namespace ?? VARIABLE_NAMESPACE}
```

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -647,7 +663,7 @@ function getVariableSet(
     layout: 'combobox',
     name: VAR_LABELS,
     onAddCustomValue: onAddCustomAdHocValue,
-    readonlyFilters: (readOnlyLabelFilters ?? []).map((f) => ({ ...f, origin: embedderName, readOnly: true })),
+    filters: initialLabelFilters || [],
```
gtk-grafana:
```suggestion
    filters: initialLabelFilters ?? [],
```

gtk-grafana (commented):
Everything is looking good for me locally, a few super minor nits (we use null coalesce here) in the code, but this looks good to me. 

@matyax will you have a few minutes today to review? It would be good to get this merged ASAP and get a deploy out early next week

matyax:
Surely! I've been keeping an eye on this while you were iterating on it.

ifrost:
> @ifrost I added the referenceQuery to the testing route props in [33cd4b0](https://github.com/grafana/logs-drilldown/commit/33cd4b0ac00ab92c088d9d289acc4465bbbb6128). The testing route can be hit at `/a/grafana-lokiexplore-app/embed`
> 
> By external deps I mean PRs in external repos, the testing route is supposed to save the rest of the squad from needing to spin up Logs Drilldown in an external app, and to provide us something to run e2e tests against that should provide the same behavior as within an embedded app (e.g. Asserts).

Oh that's neat! Didn't know about it. It'd be useful be useful in other apps 👍

src/Components/EmbeddedLogsExploration/EmbeddedLinkScene.tsx
```diff
@@ -41,6 +41,13 @@ export class EmbeddedLinkScene extends SceneObjectBase {
       return null;
     }
     const params = sceneUtils.getUrlState(indexScene);
+
+    // The link to breakdown cannot be created without labels
+    // Alternatively we could build a link to ServiceSelectionScene
+    if (labelsVarFilters.length === 0) {
+      return null;
```
matyax, line 52: Worth adding a warning logger to not silently fail.

src/Components/IndexScene/VariableLayoutScene.tsx
```diff
@@ -99,6 +99,16 @@ export class VariableLayoutScene extends SceneObjectBase<VariableLayoutSceneStat
                     ) : null;
                   })}
                 </div>
+                {!currentFiltersMatchReference && (
```
matyax: Just so you're aware, this can be potentially displayed if the reference filters are not defined.


```typescript
public currentFiltersMatchReference() {
    const referenceLabelsDefined = this.state.referenceLabels && this.state.referenceLabels.length > 0;
    return (
      !referenceLabelsDefined || ...
```

src/Components/IndexScene/VariableLayoutScene.tsx
```diff
@@ -99,6 +99,16 @@ export class VariableLayoutScene extends SceneObjectBase<VariableLayoutSceneStat
...
+                  <Button
+                    icon="repeat"
+                    variant="secondary"
+                    onClick={() => indexScene.resetToReferenceQuery()}
+                    tooltip="Reset filters to initial values."
+                  >
+                    Reset
+                  </Button>
```
gtk-grafana:
```suggestion
      <Button
        icon="repeat"
        variant="secondary"
        aria-label={t('logs.drilldown.buttons.reset-filters','Reset filters to initial values')}
        onClick={() => indexScene.resetToReferenceQuery()}
        data-testid="data-testid reset-filters-button"  // we should add this to `testIds.variables.resetFiltersButton` object
        tooltip={t('logs.drilldown.buttons.reset-filters','Reset filters to initial values')}
      >
        Reset
      </Button>
```

matyax (commented):
Just a couple of minor comments, but LGTM!

src/Components/EmbeddedLogsExploration/EmbeddedLinkScene.tsx
gtk-grafana: I'm not sure this is unexpected behavior that needs to be logged as this can be triggered when users remove all their labels, but I am seeing another runtime error when the last label filter is removed:

```
PanelQueryRunner Error TypeError: Cannot read properties of undefined (reading 'key')
    at getPrimaryLabelFromEmbeddedScene (variableHelpers.ts:105:42)
    at getRouteParams (routing.ts:181:82)
    at getExplorationPrefix (store.ts:203:51)
    at getMaxLines (store.ts:217:18)
    at get maxLines (ServiceScene.tsx:819:25)
    at lodash.js:2730:1
    at arrayEach (lodash.js:530:1)
    at baseClone (lodash.js:2727:1)
    at lodash.js:2733:1
    at arrayEach (lodash.js:530:1)
```

src/Components/IndexScene/VariableLayoutScene.tsx
gtk-grafana: Yeah, although currently if not embedded it will redirect to the service selection scene, but embedded without reference would display this button (which would do nothing)

gtk-grafana (commented):
Can we add a copy of the reset button to this warning?
Image

something like:

Image

gtk-grafana:
@ifrost I pushed a commit fixing the e2e test failure in CI: https://github.com/grafana/logs-drilldown/pull/1549/commits/3ff805f767a1a0572d462c73f846c7d1f69e12c2

ifrost:
Regarding the error mentioned in the comment

```
PanelQueryRunner Error TypeError: Cannot read properties of undefined (reading 'key')
```

This is a great spot @gtk-grafana 🙌  I didn't notice there was a path in ServiceScene redirection when filters are invalid (empty or without inclusive operator). 

This means that embedded mode can enter invalid state for the same reasons and instead of redirection we need to handle it internally. I've moved showing an alert to ServiceScene and put validation in one place in b9a06b47dcd2083915f583e988b0a0c2398acd86 + reused the reset button in the warning message


https://github.com/user-attachments/assets/2f986e63-80f2-4551-8a68-5a720c5c9afa

gtk-grafana:
@ifrost pretty big regression in the [last commit](https://github.com/grafana/logs-drilldown/pull/1549/commits/b9a06b47dcd2083915f583e988b0a0c2398acd86)

image
Seeing message and no logs panel when I have valid filters and not embedded, I think this is what is breaking e2e as well.

To repro: 
1. Visit service selection: http://localhost:3000/a/grafana-lokiexplore-app/explore
2. select any label (show logs)
3. Remove label which redirects to service selection
4. Select the same label
5. BAM - broken

Looks like any label you've selected before breaks, but refreshing the page seems to fix so I'm guessing something is rubbing up against the scene cache

ifrost:
> @ifrost pretty big regression in the [last commit](https://github.com/grafana/logs-drilldown/pull/1549/commits/b9a06b47dcd2083915f583e988b0a0c2398acd86)

Yep, keeping "filters validity" in state was a mistake. The state gets stale when moving back and forth between views (removing labels, or using browser history) and would require cleaning up which may feel clunky. In 7a2c847dd9d33600d7638c9fe0acd2981fd93952 I have switched from storing the state to deriving it in runtime, and split the the handling logic from the subscription handler to separate methods.

It's been a few back and forth with this PR. If it's gonna be easier I can split it into smaller PRs and start with the refactoring only 🤔  (e.g. https://github.com/grafana/logs-drilldown/pull/1551)

gtk-grafana:
Sounds good, I ran into this error while testing: https://github.com/grafana/logs-drilldown/pull/1552, but that exists on main and is not related to your changes.

Looks good to me, everything is working locally, although I agree it's a bit messy, but refactoring in https://github.com/grafana/logs-drilldown/pull/1551 should work fine.

@matyax Soft approval from me, feel free to approve and merge in the morning if you have some time, otherwise I can give final approval tomorrow afternoon.

gtk-grafana:
oh @ifrost please sign the CLA again, I don't know why it expires

src/Components/IndexScene/ResetFiltersButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+export function ResetFiltersButton({ indexScene }: Props) {
+  const { currentFiltersMatchReference } = indexScene.useState();
+
+  return (
+    !currentFiltersMatchReference && (
+      <Button
+        icon="repeat"
+        variant="secondary"
+        onClick={() => indexScene.resetToReferenceQuery()}
+        tooltip="Reset filters to initial values."
```
matyax: There's just one thing that I noticed while testing this, is that we should be more explicit with the outcome. It will reset the label filters, but it's not being mentioned.

```suggestion
        tooltip="Reset label filters to initial values."
```

Image

Labels reset, but other filters are kept.

Image

matyax: If you don't mind, I'll apply the suggestion so we can move forward with this.

src/Components/EmbeddedLogsExploration/EmbeddedLinkScene.tsx
```diff
@@ -26,7 +26,7 @@ export class EmbeddedLinkScene extends SceneObjectBase {
 
     // Rerender this scene whenever any dependent variables are updated
```
matyax, line 27: Outdated comment.

src/Components/IndexScene/VariableLayoutScene.tsx
```diff
@@ -99,6 +100,7 @@ export class VariableLayoutScene extends SceneObjectBase<VariableLayoutSceneStat
                     ) : null;
                   })}
                 </div>
+                <ResetFiltersButton indexScene={indexScene} />
```
matyax: We could omit rendering this if not embedded.

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -761,6 +810,27 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
+
+    if (!status.isValid) {
+      return (
+        <Alert title="Invalid labels selected" severity="info">
+          <div className={css({ display: 'flex', justifyContent: 'space-between', alignItems: 'center' })}>
+            {status.reason === LabelFiltersInvalidReason.PrimaryLabelRemoved && (
+              <p>You need at least one label with inclusive matching.</p>
+            )}
+            {status.reason === LabelFiltersInvalidReason.Empty && (
+              <p>Please select at least one label to see logs breakdown.</p>
```
matyax:
```suggestion
              <p>Please select at least one label to see the logs breakdown.</p>
```

src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -761,6 +810,27 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
...
   static Component = ({ model }: SceneComponentProps<ServiceScene>) => {
     const { body } = model.useState();
+    const indexScene = sceneGraph.getAncestor(model, IndexScene);
+
+    const { filters } = getLabelsVariable(model).useState();
+    const status = model.getLabelFiltersStatus(filters);
+
+    if (!status.isValid) {
+      return (
+        <Alert title="Invalid labels selected" severity="info">
```
matyax:
```suggestion
        <Alert title={status.reason === LabelFiltersInvalidReason.Empty ? "No labels selected" : "Invalid labels selected"} severity="info">
```

tests/embed.spec.ts
```diff
@@ -239,17 +239,32 @@ test.describe('embed', () => {
...
 
     await explorePage.addCustomValueToCombobox(labelName, FilterOp.RegexEqual, ComboBoxIndex.labels, `us-.+`);
     await assertFiltersAreReadOnly();
     await page.reload();
     await assertFiltersAreReadOnly();
+
+    // Remove all filters
+    await page.getByLabel('Remove filter with key service_name').click();
+    await page.getByLabel('Remove filter with key cluster').click();
+    await expect(page.getByText('Please select at least one label to see logs breakdown.')).toHaveCount(1);
```
matyax:
```suggestion
    await expect(page.getByText('Please select at least one label to see the logs breakdown.')).toHaveCount(1);
```

matyax:
Just some minor copy improvements and this should be ready to merge 👌

src/Components/IndexScene/VariableLayoutScene.tsx
ifrost: Changed in 331f42f16a61703aa541aeea2f75a6f04d5e46b5

src/Components/EmbeddedLogsExploration/EmbeddedLinkScene.tsx
ifrost: Changed in 07300e948fbe1d8275700fa58ad7c8650cbeee0a

ifrost:
Addressed [comment](https://github.com/grafana/logs-drilldown/pull/1551#discussion_r2381749732) in 333a24809e16bb689c1b318faa868b590033fe05

matyax:
Re-testing.

matyax (approved):
👌

ifrost:
Thanks a lot @gtk-grafana and @matyax for helping with it! 🙌

PR #1551, opened 2025-09-25
refactor: split label filters validation

ifrost:
Refactor that happened to be required by https://github.com/grafana/logs-drilldown/pull/1549. This is extracting the refactor to a separate PR to make it easier to comment on changes. (please see https://github.com/grafana/logs-drilldown/pull/1549#issuecomment-3335901566)

This change extracts label filters validation logic from setSubscribeToLabelsVariable into 2 new methods:

- `getLabelFiltersStatus` (responsible for figuring if labels will create valid query)
- `handleInvalidLabels` (responsible for handling invalid state, e.g. redirect to service selection)

Exposing `getLabelFiltersStatu`s will allow us to handle removing filters in embedded mode. Based on the status instead of redirect (like in standalone mode) we will be able to show a warning message.

Having `handleInvalidLabels` in a separate function allows to have exactly the same logic during `onActivate` (it checks only if filters are empty) and when filters change (checks if they are empty and if primary label is missing). At the moment it's not gonna improve anything for users but will be more consistent for the future.

Splitting the logic should also make the code easier to follow.

### DIFF
--- a/src/Components/ServiceScene/ServiceScene.tsx
+++ b/src/Components/ServiceScene/ServiceScene.tsx
@@ -101,6 +101,21 @@ export interface ServiceSceneCustomState {
   totalLogsCount?: number;
 }
 
+enum LabelFiltersInvalidReason {
+  Empty = 'empty',
+  PrimaryLabelRemoved = 'primary_label_removed',
+}
+
+type LabelFiltersStatus =
+  | {
+      isValid: true;
+    }
+  | {
+      isValid: false;
+      newPrimaryLabel?: AdHocFilterWithLabels<{}>;
+      reason: LabelFiltersInvalidReason;
+    };
+
 export interface ServiceSceneState extends SceneObjectState, ServiceSceneCustomState {
   $data: SceneDataProvider | undefined;
   $detectedFieldsData: SceneQueryRunner | undefined;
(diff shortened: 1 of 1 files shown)


src/Components/ServiceScene/ServiceScene.tsx
```diff
@@ -181,52 +196,84 @@ export class ServiceScene extends SceneObjectBase<ServiceSceneState> {
     this.addActivationHandler(this.onActivate.bind(this));
   }
 
-  private setSubscribeToLabelsVariable() {
-    const labelsVariable = getLabelsVariable(this);
-    if (labelsVariable.state.filters.length === 0) {
-      this.redirectToStart();
-      return;
+  private handleInvalidLabels(status: LabelFiltersStatus) {
```
matyax, line 199: I'm a bit conflicted with this function, after seeing it used below, like:

```ts
const { isValid } = this.handleInvalidLabels(this.getLabelFiltersStatus(newState.filters));
```

I think that, if you remove the return, and just do:

```ts
handleInvalidLabels(status: LabelFiltersStatus) {
  if (status.valid) {
    return;
  }
  // Rest of the function
}
```

And below, split it:

```ts
const { isValid } = this.getLabelFiltersStatus(newState.filters);
if (!isValid) {
  this.handleInvalidLabels();
}
```

It would be resolved.

ifrost:
Closing. I'll address [comment](https://github.com/grafana/logs-drilldown/pull/1551#discussion_r2381749732) directly in https://github.com/grafana/logs-drilldown/pull/1549

src/Components/ServiceScene/ServiceScene.tsx
ifrost: Adjusted directly in #1549 ([comment](https://github.com/grafana/logs-drilldown/pull/1549#issuecomment-3338011479))

PR #1554, opened 2025-09-29
feat(LogsListScene): add defaultDisplayedFields support

matyax:
Logs Drilldown part of https://github.com/grafana/grafana/pull/110901

Part of https://github.com/grafana/grafana/issues/101083

### DIFF
--- a/src/Components/ServiceScene/LogOptionsScene.tsx
+++ b/src/Components/ServiceScene/LogOptionsScene.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 
 import { css } from '@emotion/css';
 
-import { GrafanaTheme2, LogsSortOrder } from '@grafana/data';
+import { GrafanaTheme2, LogsSortOrder, shallowCompare } from '@grafana/data';
+import { t } from '@grafana/i18n';
 import { locationService } from '@grafana/runtime';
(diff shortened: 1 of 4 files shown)


L2D2Grafana (commented):
https://github.com/user-attachments/assets/6559cf15-2a90-4a89-ae91-9f40d07e8d81

Oddities
- Onload it shows all the OTEL fields, but doesn't show Log Attributes like core. It shows scope_name
- Switch to the table tab and scope_name is in the column, switch back to logs panel and only scope_name is there and the other OTEL fields are gone
- Clicking show original log line once clicked stays there

This might be easier to implement once we merge https://github.com/grafana/logs-drilldown/pull/1534

matyax:
Did you test this with the other branch in core Grafana? Without that it's not going to work. I assumed it was implicit but I should have mentioned in the description.

gtk-grafana:
@matyax does this work with our generated otel services? (if not can we add an otel service with the expected fields?)

matyax:
> @matyax does this work with our generated otel services? (if not can we add an otel service with the expected fields?)

Great question. Maybe, but I haven't tried it with it, and it's probably very out of date. At some point, @shantanualsi was going to update it with the version that uses the dot notation for fields that currently use underscore. That should help us to debug Drilldown Logs and Explore with it.

>  until after we deploy 1.0.29.

Yeah, let's wait until after. This is for internal usage for now.

Thanks!

PR #1555, opened 2025-09-29
feat: add "where are my logs function"

egovova11:
## Why

Addresses issue https://github.com/grafana/asserts-app-plugin/issues/2164

It often happens that there are no logs matching the filter for a given logs configuration in Asserts. For this, I would like to introduce a "Where are my logs button?" that appears whenever logs drilldown is embedded into Asserts.

### Disclaimer: code is mostly generated by AI

- Provides context for searching for logs via Grafana Assistant whenever there are no logs matching default search criteria.
- [time context is coming from Asserts app plugin](https://github.com/grafana/asserts-app-plugin/pull/2178)

## Testing notes

Tested locally against assertsdemo environment:

image

### DIFF
--- a/src/Components/ServiceScene/JSONPanel/LogsJSONComponent.tsx
+++ b/src/Components/ServiceScene/JSONPanel/LogsJSONComponent.tsx
@@ -57,7 +57,7 @@ export default function LogsJSONComponent({ model }: SceneComponentProps<JSONLog
   $data.useState();
 
   const logsListScene = sceneGraph.getAncestor(model, LogsListScene);
-  const { visualizationType, selectedLine } = logsListScene.useState();
+  const { visualizationType, selectedLine, onWhereAreMyLogs } = logsListScene.useState();
   const styles = useStyles2(getStyles, wrapLogMessage);
 
   const fieldsVar = getFieldsVariable(model);
@@ -235,7 +235,11 @@ export default function LogsJSONComponent({ model }: SceneComponentProps<JSONLog
         </>
       )}
       {error && (
-        <LogsPanelError error={error} clearFilters={canClearFilters ? () => clearVariables(model) : undefined} />
+        <LogsPanelError
+          error={error}
+          clearFilters={canClearFilters ? () => clearVariables(model) : undefined}
+          onWhereAreMyLogs={onWhereAreMyLogs}
+        />
       )}
     </div>
   );
--- a/src/Components/ServiceScene/LogsListScene.tsx
+++ b/src/Components/ServiceScene/LogsListScene.tsx
@@ -2,8 +2,8 @@ import React from 'react';
 
 import { css } from '@emotion/css';
 
-import { LoadingState, PanelData } from '@grafana/data';
 import { openAssistant } from '@grafana/assistant';
+import { LoadingState, PanelData } from '@grafana/data';
 import { locationService } from '@grafana/runtime';
 import {
   SceneComponentProps,
@@ -33,8 +33,8 @@ import { LogsPanelScene } from './LogsPanelScene';
 import { LogsTableScene } from './LogsTableScene';
 import { LogsVolumePanel, logsVolumePanelKey } from './LogsVolume/LogsVolumePanel';
 import { ServiceScene } from './ServiceScene';
-import { isEmptyLogsResult } from 'services/logsFrame';
 import { buildAssistantContext } from 'services/assistant';
+import { isEmptyLogsResult } from 'services/logsFrame';
 import {
   getBooleanLogOption,
   getDisplayedFields,
@@ -265,7 +265,7 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
     }
   }
 
-  showLogsError({ message, isEmptyResult }: { message: string; isEmptyResult: boolean }) {
+  showLogsError({ message, isEmptyResult }: { isEmptyResult: boolean; message: string }) {
     const logsVolumeCollapsedByError = this.state.logsVolumeCollapsedByError ?? !getLogsVolumeOption('collapsed');
     const indexScene = sceneGraph.getAncestor(this, IndexScene);
     const clearableVariables = getVariablesThatCanBeCleared(indexScene);
@@ -297,7 +297,12 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
     }
 
     this.whereAreMyLogsHandler = undefined;
-    this.setState({ error: undefined, isEmptyResult: undefined, logsVolumeCollapsedByError: undefined, onWhereAreMyLogs: null });
+    this.setState({
+      error: undefined,
+      isEmptyResult: undefined,
+      logsVolumeCollapsedByError: undefined,
+      onWhereAreMyLogs: null,
+    });
 
     // Recreate the panel with the cleared error state
     this.updateLogsPanel();
--- a/src/Components/ServiceScene/LogsListScene.tsx
+++ b/src/Components/ServiceScene/LogsListScene.tsx
@@ -438,7 +438,7 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
 
   private getWhereAreMyLogsHandler(indexScene: IndexScene, isEmptyResult: boolean) {
     if (
-      !isEmptyResult ||
+      !this.state.isEmptyResult ||
       !indexScene.state.embedded ||
       indexScene.state.embedderName !== 'Asserts' ||
       !indexScene.state.currentFiltersMatchReference ||
--- a/src/Components/ServiceScene/LogsListScene.tsx
+++ b/src/Components/ServiceScene/LogsListScene.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { css } from '@emotion/css';
 
 import { LoadingState, PanelData } from '@grafana/data';
+import { openAssistant } from '@grafana/assistant';
 import { locationService } from '@grafana/runtime';
 import {
   SceneComponentProps,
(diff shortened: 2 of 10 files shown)


src/Components/ServiceScene/LogsPanelError.tsx
```diff
@@ -7,18 +7,26 @@ import { GrotError } from 'Components/GrotError';
 interface Props {
   clearFilters?: () => void;
   error: string;
+  onWhereAreMyLogs?: (() => void) | null;
```
matyax, line 10:
```suggestion
  callToActionCallback?: () => void;
  callToAction?: string;
```

Let's implement this in a more generic way. Also there's no need for null, as the question sign makes it possibly undefined.

src/Components/ServiceScene/JSONPanel/LogsJSONComponent.tsx
```diff
@@ -57,7 +57,7 @@ export default function LogsJSONComponent({ model }: SceneComponentProps<JSONLog
   $data.useState();
 
   const logsListScene = sceneGraph.getAncestor(model, LogsListScene);
-  const { visualizationType, selectedLine } = logsListScene.useState();
+  const { visualizationType, selectedLine, onWhereAreMyLogs } = logsListScene.useState();
```
matyax, line 60: The callback should be a function of logsListScene, and we shouldn't use state to store callbacks.

Let's do something like:

```suggestion
  const { visualizationType, selectedLine } = logsListScene.useState();
  
  const findLogsCallback = logsListScene.getWhereAreMyLogsHandler();
```

src/Components/ServiceScene/JSONPanel/LogsJSONComponent.tsx
```diff
@@ -235,7 +235,11 @@ export default function LogsJSONComponent({ model }: SceneComponentProps<JSONLog
         </>
       )}
       {error && (
-        <LogsPanelError error={error} clearFilters={canClearFilters ? () => clearVariables(model) : undefined} />
+        <LogsPanelError
+          error={error}
+          clearFilters={canClearFilters ? () => clearVariables(model) : undefined}
+          onWhereAreMyLogs={onWhereAreMyLogs}
```
matyax, line 241:
```suggestion
          callToActionCallback={findLogsCallback}
          callToAction={findLogsCallback ? 'Find logs with Grafana Assistant' : undefined}
```

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -48,9 +50,11 @@ export interface LogsListSceneState extends SceneObjectState {
   controlsExpanded: boolean;
   displayedFields: string[];
   error?: string;
+  isEmptyResult?: boolean;
   lineFilter?: string;
   loading?: boolean;
   logsVolumeCollapsedByError?: boolean;
+  onWhereAreMyLogs?: (() => void | Promise<void>) | null;
```
matyax, line 57:
```suggestion
```

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -168,12 +173,30 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
 
   public onActivate() {
     const searchParams = new URLSearchParams(locationService.getLocation().search);
+    const indexScene = sceneGraph.getAncestor(this, IndexScene);
     this.setStateFromUrl(searchParams);
 
     if (!this.state.panel) {
       this.updateLogsPanel();
     }
```
matyax, line 182: All these changes shouldn't be necessary after not using state for storing the callback.

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -392,6 +435,37 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
...
+  private getWhereAreMyLogsHandler(indexScene: IndexScene, isEmptyResult: boolean) {
+    if (
+      !isEmptyResult ||
+      !indexScene.state.embedded ||
+      indexScene.state.embedderName !== 'Asserts' ||
+      !indexScene.state.currentFiltersMatchReference ||
+      !indexScene.state.isAssistantAvailable
+    ) {
+      this.whereAreMyLogsHandler = undefined;
+      return null;
```
matyax, line 448:
```suggestion
      return undefined;
```

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -392,6 +435,37 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
       direction: 'column',
     });
   }
+
+  private getWhereAreMyLogsHandler(indexScene: IndexScene, isEmptyResult: boolean) {
```
matyax, line 439: Shouldn't be needed.

```suggestion
  private getWhereAreMyLogsHandler(indexScene: IndexScene) {
```

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -392,6 +435,37 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
...
+        }
+
+        openAssistant({
+          origin: 'logs-drilldown-empty-results',
+          prompt:
+            'Investigate why the Asserts KPI drawer shows an empty logs table. Validate the label-to-log mapping and suggest next steps to restore logs visibility.',
+          context,
+        });
+      };
+    }
```
matyax, line 465: If the assistant is available, plus all the conditions checked above, we can return the callback.

```suggestion
    this.whereAreMyLogsHandler = async () => {
      const context = await buildAssistantContext(indexScene);
      if (!context.length) {
        return;
      }

      openAssistant({
        origin: 'logs-drilldown-empty-results',
        prompt:
          'Investigate why the Asserts KPI drawer shows an empty logs table. Validate the label-to-log mapping and suggest next steps to restore logs visibility.',
        context,
      });
    };
```

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -477,7 +477,11 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
         <span className={styles.panelWrapper}>
           {!error && <body.Component model={body} />}
           {error && (
-            <LogsPanelError error={error} clearFilters={canClearFilters ? () => clearVariables(body) : undefined} />
+            <LogsPanelError
+              error={error}
+              clearFilters={canClearFilters ? () => clearVariables(body) : undefined}
+              onWhereAreMyLogs={model.getParentScene().state.onWhereAreMyLogs}
```
matyax, line 483: Same as above, let's do:

```ts
const findLogsCallback = model.getParentScene().getWhereAreMyLogsHandler();
```

```suggestion
              callToActionCallback={findLogsCallback}
              callToAction={findLogsCallback ? 'Find logs with Grafana Assistant' : undefined}
```

src/Components/ServiceScene/LogsTableScene.tsx
```diff
@@ -323,7 +323,11 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
           </>
         )}
         {error && (
-          <LogsPanelError error={error} clearFilters={canClearFilters ? () => clearVariables(model) : undefined} />
+          <LogsPanelError
+            error={error}
+            clearFilters={canClearFilters ? () => clearVariables(model) : undefined}
+            onWhereAreMyLogs={parentModel.state.onWhereAreMyLogs}
```
matyax, line 329: Same as the other visualizations.

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -225,24 +248,38 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
...
-      this.showLogsError('No logs match your search. Please try a different time range.');
+      this.showLogsError({
+        isEmptyResult: true,
+        message: 'No logs match your search. Please try a different time range.',
+      });
     }
   }
 
-  showLogsError(error: string) {
+  showLogsError({ message, isEmptyResult }: { isEmptyResult: boolean; message: string }) {
```
matyax, line 268:
```suggestion
  showLogsError(error: string) {
```

Updated. If it's an empty result or not depends on the state in LogListScene.

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -392,6 +435,37 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
       direction: 'column',
     });
   }
+
+  private getWhereAreMyLogsHandler(indexScene: IndexScene, isEmptyResult: boolean) {
+    if (
+      !isEmptyResult ||
```
matyax:
```suggestion
      !this.state.isEmptyResult ||
```

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -168,12 +173,30 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
...
+          }
+          return;
+        }
+
+        const handler = this.getWhereAreMyLogsHandler(indexScene, true);
+        if (this.state.onWhereAreMyLogs !== handler) {
+          this.setState({ onWhereAreMyLogs: handler });
+        }
+      })
+    );
```
matyax, line 198:
```suggestion
```

Plase set `isEmptyResult` inside `handleNoData()` and clear it in `clearLogsError()`

matyax (commented):
This is a great idea! Thank you for submitting the PR.

I left a suggestion for a much simpler implementation that also leaves room for other types of CTA suggestions on error. There might be other adjustments needed, but hopefully you'll get the concept. Let me know otherwise.

gtk-grafana:
@matyax @egovova11 
This is an interesting idea, however I would suggest looking into adding this functionality as a component extension in asserts, instead of putting plugin specific code in logs drilldown. Additionally, since a misconfigured asserts would break more then just logs, moving this to a component extension would make this functionality available in metrics, traces, etc without requiring duplicate effort across multiple repos.

matyax:
Agree. Also, alternative to a link extension, this could be an optional prop passed by the embedder application, which removes the need to checking for the embedder name.

gtk-grafana:
@matyax You're absolutely right!

matyax:
Thanks again! To be superseded by https://github.com/grafana/logs-drilldown/pull/1571

PR #1561, opened 2025-10-06
fix: validate primary label correctly

### LINKED ISSUE #1558: [BUG]: Wrong error message shown - You need at least one label with inclusive matching.
**Describe the bug**
Users see the wrong error message when they have the primary label present.

You need at least one label with inclusive matching.
Image

Image

**To Reproduce**
[https://play.grafana.org/a/grafana-lokiexplore-app/explore/service/dcr/logs?from=now-[…]jsonFields=&var-patterns=&var-lineFilterV2=&var-lineFilters=](https://play.grafana.org/a/grafana-lokiexplore-app/explore/service/dcr/logs?from=now-15m&to=now&var-ds=grafanacloud-logs&var-filters=service%7C%3D%7Cdcr&var-filters=install%7C%3D%7Cvodafone&var-filters=component%7C%3D%7Csrs&var-filters=container%7C%3D%7C__CV%CE%A9__dcr-srs,dcr-srs&patterns=%5B%5D&var-lineFormat=&var-fields=&var-levels=&var-metadata=&var-jsonFields=&var-patterns=&var-lineFilterV2=&var-lineFilters=)

**Expected behavior**
Users should see the error message: No logs found in service_name

ifrost:
Fixes #1558

We have special handling to keep old ULRs like `.../explore/service=foo...?filters=service_name=foo` intact. However it still may happen that `service` is actually the primary filter and there's no `service_name` in filters: `.../explore/service=foo...?filters=service=foo`. In this case the URL is valid, but the validation logic would always assume `service_name` should be defined when URL `explore/service/value`

The bug happened only when `service` was the primary filter (the first in the selection). The validation logic assumed it's a legacy URL and looked for `service_name` in filters but couldn't find it and tried to set primary signal to `service` once again which didn't have any effect.

### DIFF
--- a/docs/sources/access/_index.md
+++ b/docs/sources/access/_index.md
@@ -17,16 +17,16 @@ To use Grafana Logs Drilldown to view your logs data, you can either access it i
 
 {{< docs/play title="the Grafana Play site" url="https://play.grafana.org/a/grafana-lokiexplore-app/explore?var-ds=ddhr3fttaw8aod&var-patterns=&var-lineFilter=&var-logsFormat=" >}}
 
-## Access in Grafana Cloud
+## Access in Grafana and Grafana Cloud
 
-To access Grafana Logs Drilldown:
+Grafana Logs Drilldown is installed by default in both Grafana and Grafana Cloud. To access Grafana Logs Drilldown:
 
 1. Open your Grafana stack in a web browser.
 1. In the main menu, select **Drilldown** > **Logs**.
 
 ## Installation
 
-If you are not using Grafana Cloud, you can install Grafana Logs Drilldown in your Grafana environment.
+Logs Drilldown has been installed by default since Grafana v11.3.0.  If you are upgrading, you may need to manually install the plugin.
 
 ### Install via Plugins catalog
 
@@ -36,7 +36,7 @@ For Enterprise and OSS Grafana users, you can install Grafana Logs Drilldown via
 1. Click the **Installation** tab.
 1. Follow the instructions to install the app.
 
-### Install in Loki
+### Enable in Loki configuration
 
 The following Loki and Grafana version and configuration are required:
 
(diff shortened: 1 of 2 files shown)


matyax:
I was in the process of a similar fix, and found that we should also update this function to keep "service" if it's the primary label.

https://github.com/grafana/logs-drilldown/blob/main/src/services/variableHelpers.ts#L40

Something like:

```ts
export function clearVariables(sceneRef: SceneObject) {
  // clear patterns: needs to happen first, or it won't work as patterns is split into a variable and a state, and updating the variable triggers a state update
  const indexScene = sceneGraph.getAncestor(sceneRef, IndexScene);
  indexScene.setState({
    patterns: [],
  });

  const variablesToClear = getVariablesThatCanBeCleared(indexScene);

  variablesToClear.forEach((variable) => {
    if (variable instanceof AdHocFiltersVariable) {
      let { labelName, labelValue } = getRouteParams(sceneRef);
      
      const filters = variable.state.filters.filter((filter) => {
        if (!isOperatorInclusive(filter.operator) && filter.value === labelValue) {
          return false;
        }
        if (filter.key === labelName || (labelName === SERVICE_UI_LABEL && filter.key === SERVICE_NAME)) {
          return true;
        }
        return false;
      });
      variable.setState({ filters });
    } else if (variable instanceof CustomConstantVariable) {
      variable.setState({
        text: '',
        value: '',
      });
    }
  });
}
```

Mind adding this too? Thanks!

ifrost:
> I was in the process of a similar fix, and found that we should also update this function to keep "service" if it's the primary label.
> 
> https://github.com/grafana/logs-drilldown/blob/main/src/services/variableHelpers.ts#L40
> 
> Something like:
> 
> ```ts
> ...
> ```
> 
> Mind adding this too? Thanks!

Added in 3f731cbc669d22e98bb91b21b61562f47fb9561e

L2D2Grafana:
Seeing this error message when service is selected. Screenshot 2025-10-06 at 1 14 05 PM When service is remove I'm automatically sent back to the first index page. 
Screenshot 2025-10-06 at 1 14 20 PM It seems like people would never see the error message `No Logs found in service_name`

L2D2Grafana (approved):

Issue #1566, opened 2025-10-07
fix: labels clear variable state not showing


gtk-grafana:
This fixes the bug preventing the redirect back to the service selection scene, however the count of the clearable variables is wrong leading to showing the `We did not find any labels for the given timerange.` state instead of the `No labels match these filters. [clear filters]` state

_Originally posted by @gtk-grafana in https://github.com/grafana/logs-drilldown/pull/1561#pullrequestreview-3310222985_

Exclusion filters or other label filters that exclude all results do not trigger the clear variable state:
<img width="860" height="862" alt="Image" src="https://github.com/user-attachments/assets/c489674e-a835-4c3f-a4d1-6ba62825d211" />

Since the clear variables will retain the primary label (inclusive label filter added on the service selection), we should provide an action for users to take when the queries do not return any results.

PR #1567, opened 2025-10-07
fix: unexpected clear variable behavior

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1566

If there are any variables that can be cleared we should be showing the "clear variables" button instead of the empty state UI.


gtk-grafana:
@L2D2Grafana good catch, we should sync the fields/labels clear filters state with the logs panel state. I created an issue from your comment above for us to follow up.

PR #1571, opened 2025-10-08
feat(EmptyLogs): add button to fix with assistant

matyax:


webpack.config.ts
```diff
@@ -15,13 +15,13 @@ const config = async (env: any): Promise<Configuration> => {
...
       // new BundleAnalyzerPlugin(),
-      new LiveReloadPlugin({
+      /*new LiveReloadPlugin({
         appendScriptTag: true,
         delay: 1000,
         hostname: 'localhost',
         port: 35828,
         protocol: 'http',
-      }),
+      }),*/
```
L2D2Grafana: Goodbye LiveReloadPlugin

src/Components/Pages.tsx
```diff
@@ -99,6 +99,10 @@ function getEmbeddedScene() {
   const props: EmbeddedLogsExplorationProps & IndexSceneState = {
     embedded: true,
     embedderName: 'EmbeddedLogs',
+    options: {
+      noLogsCustomPrompt: 'Write a haiku about Loki Logs',
+      noLogsPromptCTA: 'Where are my logs?',
```
L2D2Grafana: This PR seems great ✨ , but I keep getting stuck on Write a haiku about Loki Logs, wouldn't we always want Where are my logs?

Image

matyax: It's just a test dummy custom prompt for the embed route. If not provided it should use the default one, and regular logs Drilldown should use the default one too (investigate etc...).

gtk-grafana:
Should we add to the `EmptyLayoutScene` for the fields/labels breakdowns?

matyax:
Yeah, I don't see why not.

src/Components/EmbeddedLogsExploration/types.ts
```diff
@@ -6,10 +6,26 @@ interface SceneTimeRangeStateStub {
...
+type EmptyStateConfig = {
+  fields?: EmptyStateOptions;
+  labels?: EmptyStateOptions;
+  logs?: EmptyStateOptions;
+};
+
+export interface EmbeddedLogsOptions {
+  emptyStates: EmptyStateConfig;
+}
+
```
matyax, line 23: Refactored to expose custom prompts and call to actions for each state. The embedder can only define the one that is relevant for them.

src/Components/Pages.tsx
```diff
@@ -99,6 +99,14 @@ function getEmbeddedScene() {
   const props: EmbeddedLogsExplorationProps & IndexSceneState = {
     embedded: true,
```
matyax, line 100: Updated usage for embedders.

src/services/variableHelpers.ts
```diff
@@ -25,7 +25,16 @@ export function getVariablesThatCanBeCleared(
       continue;
     }
     if (variable instanceof AdHocFiltersVariable && variable.state.filters.length) {
-      variablesToClear.push(variable);
+      const containsPrimaryLabelFilter = variable.state.filters.some((filter) =>
+        isPrimaryLabelAdHocFilter(variable, filter, indexScene)
+      );
```
matyax, line 30: @grafana/observability-logs here I found something to improve, where if we had only one variable with the service name and value, we were incorrectly reporting it as a variable that could be cleared, and the clear button was a noop. This fixes it.

Also, if there was, for example, a field called service or service name, we will skipping clearing it.

Captura de pantalla 2025-10-09 a la(s) 17 15 40

matyax:
I refactored the exposed options to have a better shape and integrated the assistant with the other empty states. Also in the process discovered a small bug with variable clearing. Let me know what you think!

matyax:
@egovova11 Merged! Please note that the embedder props have changed a little bit.

PR #1598, opened 2025-10-24
feat(Grafana Assistant): Improve context, provide questions, and cleanup

matyax:
This PR:

- Sanitizes values so internal identifiers don't show up in Assitant context
  b
  a
  
- Adds to the context:
  - Levels filters
  - Field filters (with parser)
  - Metadata filters
 - Adds operators to the context, so the Assistant can tell when a filter is not inclusive.
  Context
  Filters
  Interpretation
  
 - Adds custom questions for service selection and service breakdown
 - Adds missing unregister() calls for the provided context.
 
 ### How to test

- Opening the assistant in the service selection page should show service selection questions
- Transitioning to breakdown details should show service details questions
- Changing filters should update the context in the assistant
- Leaving Drilldown should clean up previous context

### DIFF
--- a/.github/workflows/update-deployment-tools.yml
+++ b/.github/workflows/update-deployment-tools.yml
@@ -1,10 +1,8 @@
 name: Update Deployment Tools
 description: |
-  Updates the version of a plugin in deployment_tools libsonnet files.
+  Updates the catalog version of a plugin in deployment_tools libsonnet files.
   This action can update a single environment or all environments at once.
-  If provisioning from the plugins catalog, set catalog to true.
-  Be mindful if there is a mix of catalog and non-catalog plugins in the same environment, all will not work as intended.
-  It will modify the version string in the specified libsonnet files and create a commit with the changes.
+  It will modify the _catalog_version string in the specified libsonnet files and create a commit with the changes.
 
 on:
   workflow_dispatch:
@@ -23,11 +21,6 @@ on:
           - prod-canary
           - prod
           - all
-      catalog:
-        description: 'Provision from the plugins catalog'
-        required: true
-        type: boolean
-        default: false
 
 env:
   PLUGIN_ID: grafana-lokiexplore-app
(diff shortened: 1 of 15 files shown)


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -342,7 +358,19 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
...
     );
+    this._subs.add(
+      getLevelsVariable(this).subscribeToState(async () => {
+        await updateAssistantContext(this, setAssistantContext);
+      })
+    );
+    this._subs.add(
+      getFieldsVariable(this).subscribeToState(async () => {
+        await updateAssistantContext(this, setAssistantContext);
+      })
```
L2D2Grafana, line 369: I'm not seeing field get added as a pill.

Image

src/services/assistant.ts
```diff
@@ -34,5 +56,70 @@ export const updateAssistantContext = async (
...
+}
+
+export function provideServiceSelectionQuestions() {
+  return provideQuestions(`${PLUGIN_BASE_URL}/**`, [
+    {
+      prompt: 'How do I select the right service to see logs?',
+    },
+    {
+      prompt: 'Help me find labels with error spikes',
+    },
```
L2D2Grafana, line 109: I don't see these questions popping up, is there a different way to trigger it?

Image

src/Components/IndexScene/IndexScene.tsx
matyax: Tell the agent "show me what's in your current context and only that", something like that. It's passed as structured context, which is invisible. The current supported types are https://github.com/grafana/grafana-assistant-app/blob/main/packages/%40grafana/assistant/README.md#createassistantcontextitemttype-t-params-contexttypeparamst-chatcontextitem

src/services/assistant.ts
matyax: From your screenshot below, these are the initial qs:

imagen

L2D2Grafana (approved):
Lgtm, although it does seem like assistant already know this stuff about Logs.

matyax:
> it does seem like assistant already know this stuff about Logs

What do you mean? Before these changes it didn't know about negative filters, it didn't have custom initial questions, and it didn't know about level and field filters.

Issue #1623, opened 2025-11-07
[FEAT]: Add rolling time window options to time picker


gtk-grafana:
**Description of the new feature**
Especially for users with retention_period greater than max_query_length set in their Loki config, Log Drilldown users might want to select shorter time ranges that are further back in time.

For example: 
* 24 hour queries:  3 days ago, 7 days ago
* 1 week queries: 3 weeks ago
* 1 month queries: 2 months ago

**Acceptance Criteria**
A concise description of requirements.

PR #1621, opened 2025-11-06
feat(time-picker): filter time ranges that are too long

gtk-grafana:
Uses Loki `retention_period` or Logs Drilldown plugin config (located at `/plugins/grafana-lokiexplore-app/`) "Maximum time picker interval" to limit the time picker default options that are too long to be queried.

Loki config is used if set, if not the plugin config is used.
If neither is set all time range options are shown.

Loki config requires Loki 3.6 or greater (k276+ weekly)

Fixes: https://github.com/grafana/logs-drilldown/issues/1166
Parent branch: https://github.com/grafana/logs-drilldown/pull/1526

Related: 
If https://github.com/grafana/grafana/issues/103478 is merged we should delete the copy paste methods from grafana core in `src/services/timePicker.ts`

Note: migrated from https://github.com/grafana/logs-drilldown/pull/1174 was too old to resurrect

PR #1625, opened 2025-11-07
feat(time-picker): add rolling time window options

gtk-grafana:
Adds rolling time window options to make it easier for users to make it easier to compare different intervals with the same durations. 

Since this is built on top of #1621, any time range that is outside of the users retention period or "Maximum time picker interval" defined in the plugin config will automatically be removed. 

Parent branch: https://github.com/grafana/logs-drilldown/pull/1621
Fixes: https://github.com/grafana/logs-drilldown/issues/1623

Adds:
* Three days ago
* Four days ago
* Five days ago
* Six days ago
* Two weeks ago
* Three weeks ago
* Four weeks ago
* Two months ago
* Three months ago

Bringing the full list to:
* Last minute
* Last 5 minutes
* Last 15 minutes
* Last 30 minutes
* Last 1 hour
* Last 3 hours
* Last 6 hours
* Last 12 hours
* Last 24 hours
* Last 2 days
* Last 7 days
* Last 30 days
* Last 60 days
* Last 90 days
* Last 6 months
* Yesterday
* Day before yesterday
* Three days ago
* Four days ago
* Five days ago
* Six days ago
* This day last week
* Previous week
* Two weeks ago
* Three weeks ago
* Four weeks ago
* Previous month
* Two months ago
* Three months ago
* Today
* Today so far
* This week
* This week so far
* This month
* This month so far

### DIFF
--- a/.github/workflows/update-deployment-tools.yml
+++ b/.github/workflows/update-deployment-tools.yml
@@ -27,7 +27,7 @@ env:
 
 permissions:
   id-token: write # Required for OIDC authentication with Vault
-  contents: read # Required to read repository contents
+  contents: write # Required to create commits and update files
 
 jobs:
   build-latest-version:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,54 @@
 # Changelog
 
+## 1.0.31
+
+##### Chores
+
+* **deps-dev:**  bump glob in the npm_and_yarn group across 1 directory (#1638) (787bb202)
+* **ci:**
+  *  update permissions (#1633) (fb786b2b)
+  *  use new token generation (#1619) (075a6415)
+  *  used outputs (#1610) (57575f65)
+  *  update deployment tools wf (#1603) (d4f4af3c)
+*  update nvmrc (#1628) (48aa341e)
+*  hide toast (#1605) (c8a7663d)
+*  comment local live reload (#1607) (1890b8b3)
+*  update playwright (#1606) (acf6cb1b)
+*  update border radii (#1602) (4c353583)
+*  set override modifier (#1599) (0bc9211b)
+* **deps:**
+  *  update dependency style-loader to v3.3.4 (#1614) (d1a05e05)
+  *  update grafana/alloy:latest docker digest to 8c7256f (#1613) (529289ea)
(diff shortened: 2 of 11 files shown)


gtk-grafana:
@jage I'm curious if you have any thoughts on adding the above options to the time picker? Anything missing? Too much?

jage:
> @jage I'm curious if you have any thoughts on adding the above options to the time picker? Anything missing? Too much?

I think it looks reasonable! 👍 

So if I understand correctly it will limit the timespans to "what's available for me" depending on Loki configuration?

gtk-grafana:
Exactly! Starting with Loki 3.6 (which is being prepped for release right now) there is a new `drilldown-limits` api which allows us to pull Loki limits into the frontend:

We're planning on:
1. [Removing options](https://github.com/grafana/logs-drilldown/pull/1621) that are longer then query retention by default, but not throwing an error if retention is exceeded.
2. Adding a new [plugin config option](https://github.com/grafana/logs-drilldown/pull/1629) to allow admins to limit any query longer than max_query_length. If this option is enabled (disabled by default) the UI will remove time range options that exceed the limit, throw an error if a time range is manually entered, and revert to the default time range - the same behavior as the current "Maximum time picker interval" setting, just using the Loki config.

The point of this specific PR is to add some more time options to make it easier to compare shorter intervals that are further back in time, particularly for users that have longer retention then max query length, but hopefully valuable for all users. 

More context: [query splitting](https://github.com/grafana/grafana/pull/75876) lets you sidestep Loki limits like when max_query_length > 24h, some users see this as a bug, others see it as a feature, depending on the motivation for using query splitting and if limiting query usage is a goal of setting the max query length. 

While long-term we'd like this decision to get made in the Loki API and managed in the Loki config, for now we're exposing this new option in the plugin config to get something out to users that are concerned there's no way to limit queries coming from Logs Drilldown in tandem with max_query_length.

L2D2Grafana (commented):
👍

L2D2Grafana (approved):

matyax:
I don't know if you noticed, but again changing data sources doesn't update the time picker, so the options from the previous selection are displayed until the page is refreshed.

matyax (approved):

PR #1627, opened 2025-11-07
fix(ci): remove node version

L2D2Grafana:
debug ci


L2D2Grafana:
Yeah we still need this in ci and now locally in https://github.com/grafana/logs-drilldown/pull/1628

PR #1629, opened 2025-11-10
feat: max_query_length support

gtk-grafana:
Adds support for Loki [`limits_config.max_query_length`](https://grafana.com/docs/loki/latest/configure/#limits_config) in the Logs Drilldown application.

This PR adds a new plugin configuration option (disabled by default) which will:
* Limit time ranges available in the time picker to intervals that are less than the max_query_length if defined
* Throws a client-side error and resets the time range when durations greater than max_query_length are set

Context:
Loki query splitting (`lokiQuerySplitting`) can side-step Loki side limits if the `max_query_length` is greater than the query split interval (24 hours), but less than the user selected time range in Logs Drilldown.

Since this behavior may or may not be desirable depending if the goal of setting max_query_length was to limit user behavior from requesting too much data or to limit simultaneous pressure on queriers, we're opting to add this as a plugin-side configuration option for now while we have discussions with the Loki squad on longer-term solutions. Ideally, since `max_query_length` is a per-tenant Loki config option, the ability to control whether a particular Loki tenant can receive split queries exceeding `max_query_length` should also be a per-tenant configuration option.

Related: https://github.com/grafana/logs-drilldown/pull/1625
Parent branch: https://github.com/grafana/logs-drilldown/pull/1621

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -497,55 +497,60 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
   private limitMaxInterval(timeRange: SceneTimeRangeLike) {
     return (newState: SceneTimeRangeState, prevState: SceneTimeRangeState) => {
       const { jsonData } = plugin.meta as AppPluginMeta<JsonData>;
-      if (jsonData?.interval || jsonData?.limitMaxQueryLength) {
-        let maxInterval: number;
-        let notice: string;
-        const lokiConfig = sceneGraph.getAncestor(timeRange, IndexScene).state.lokiConfig;
 
-        try {
-          // Check if max_query_length should limit time range selections
-          if (jsonData.limitMaxQueryLength && lokiConfig && lokiConfig !== LOKI_CONFIG_API_NOT_SUPPORTED) {
-            maxInterval = getMaxQueryLengthSeconds(lokiConfig);
-            notice = 'Time range exceeds Loki max_query_length limit.';
-          } else {
-            maxInterval = rangeUtil.intervalToSeconds(jsonData.interval ?? '');
-            notice = 'Time range interval exceeds maximum interval configured by the administrator.';
-          }
+      if (!(jsonData?.interval || jsonData?.limitMaxQueryLength)) {
+        return;
+      }
 
-          if (maxInterval) {
-            const timeRangeInterval = newState.value.to.diff(newState.value.from, 'seconds');
-            if (timeRangeInterval > maxInterval) {
-              const prevInterval = prevState.value.to.diff(prevState.value.from, 'seconds');
-              if (timeRangeInterval <= prevInterval) {
-                timeRange.setState({
-                  from: prevState.from,
-                  to: prevState.to,
-                  value: prevState.value,
-                });
-              } else {
-                const defaultRange = new SceneTimeRange(DEFAULT_TIME_RANGE);
-                timeRange.setState({
-                  from: defaultRange.state.from,
-                  to: defaultRange.state.to,
-                  value: defaultRange.state.value,
-                });
-              }
+      let maxInterval: number;
+      let notice: string;
+      const lokiConfig = sceneGraph.getAncestor(timeRange, IndexScene).state.lokiConfig;
 
-              const appEvents = getAppEvents();
-              appEvents.publish({
-                payload: [notice],
-                type: AppEvents.alertWarning.name,
-              });
-
-              reportAppInteraction('all', 'interval_too_long', {
-                attempted_duration_seconds: timeRangeInterval,
-                configured_max_interval: maxInterval,
-              });
-            }
-          }
-        } catch (e) {
-          console.error(e);
+      try {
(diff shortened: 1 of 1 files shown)


gtk-grafana:
We might want to deploy 1.0.31 before merging this, let's discuss in standup

src/services/timePicker.ts
```diff
@@ -5,44 +5,61 @@ import { JsonData } from '../Components/AppConfig/AppConfig';
...
+
+          if (maxQueryLengthSeconds && !maxQueryLengthGreaterThanInterval) {
+            return false;
+          }
+
           // Pad retention by 10%, there's no downside to querying over retention besides some empty space in the query, and it might be frustrating to not get a time range if retention is close
           const retentionGreaterThanInterval = intervalSeconds <= maxRetentionSeconds * 1.1;
           const pluginConfigGreaterThanInterval = intervalSeconds <= maxPluginConfigSeconds;
+
           return intervalSeconds === 0 || retentionGreaterThanInterval || pluginConfigGreaterThanInterval;
```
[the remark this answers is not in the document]
gtk-grafana: I keep reading this over and over, I can't tell if copilot is drunk or if I am

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -497,39 +498,50 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     return (newState: SceneTimeRangeState, prevState: SceneTimeRangeState) => {
       const { jsonData } = plugin.meta as AppPluginMeta<JsonData>;
       if (jsonData?.interval) {
```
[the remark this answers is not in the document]
gtk-grafana: Good catch bot

src/Components/AppConfig/AppConfig.tsx
```diff
@@ -92,13 +103,44 @@ const AppConfig = ({ plugin }: Props) => {
...
+          <Checkbox
+            id="disable-patterns"
+            data-testid={testIds.appConfig.pattern}
+            label={`Disable patterns`}
+            value={state?.patternsDisabled}
+            onChange={onChangePatternsDisabled}
+          />
+        </Field>
+
+        <hr />
```
L2D2Grafana, line 133: It's nice to break these into sections Image

gtk-grafana:
@L2D2Grafana 

> One possible bug is that without the max_query_length enforced I do see up to 30 days in the quick options listed. Choosing the 30days leads ends in an error query time exceeds the limit. Should we hide the 30days option?

You must have splitting disabled? Check the `lokiQuerySplitting` flag, it should be enabled by default

L2D2Grafana:
> You must have splitting disabled? Check the `lokiQuerySplitting` flag, it should be enabled by default

I believe I see 30days with max_query_length off because in my lokiConfig I have retention_period set to 30days and max_query_length set to 7days. So the error makes sense. Unless they enforce max_query_length we will prevent the call from going through, otherwise we just show them the error that tells them about the restriction.
Screenshot 2025-11-13 at 8 12 19 AM

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -496,40 +497,51 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
   private limitMaxInterval(timeRange: SceneTimeRangeLike) {
     return (newState: SceneTimeRangeState, prevState: SceneTimeRangeState) => {
       const { jsonData } = plugin.meta as AppPluginMeta<JsonData>;
-      if (jsonData?.interval) {
```
matyax, line 499: Nit, but I would appreciate if the indentation level in this function is improved. Can return early a bunch of times.

matyax:
I think we should use one or the other, right?

imagen

Because with both settings defined, the user doesn't know which has precedence over the other.

gtk-grafana:
> I think we should use one or the other, right?

@matyax I tried to call the precedence by saying "overrides other time range limits if set", I don't know if we can get rid of the old one as it applies to all datasources in the app and can't be used by Loki <3.6.

But yeah, it's confusing

matyax:
Would it be worth to introduce a radio? So you choose one or the other.

gtk-grafana:
Maybe simpler but it gets complicated since the Loki limit is per tenant and the old setting is for the whole app.

gtk-grafana:
With the max bytes limit there's an argument we don't need this anymore, after chatting with Trevor I don't think many people use this Loki setting anyway, we could close and wait for more feedback 🤷‍♂️

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: Cleaned up in https://github.com/grafana/logs-drilldown/pull/1629/commits/586df31b4251de98735f19e36fb49d3aa134e442

gtk-grafana:
I am going to close this, I think https://github.com/grafana/grafana/pull/114112 is a better pattern on multiple fronts (max bytes is a better limit to set then limiting duration on queries that might be minuscule in terms of resource consumption, avoids adding more app config that is confusing).

We can revisit at a later time if there is demand for more support with max_query_length

Issue #1622, opened 2025-11-07
[FEAT]: show logs with unknown log level in timeline


4strodev:
**Description of the new feature**
When queryless log explore was inside explore, you could see log volumes of unknown log levels in the timeline. Something like this.

<img width="1648" height="149" alt="Image" src="https://github.com/user-attachments/assets/6b82cff3-e9d0-42e9-bc3a-651e9839c7dd" />

Now with drilldown you get this. This is pretty useful when you have to store logs for services like **nginx**, **kraken** or **redis** which has limited options for log configuration.

<img width="1854" height="906" alt="Image" src="https://github.com/user-attachments/assets/13d0469b-ac30-4b1a-a1c8-25d08934585c" />

**Acceptance Criteria**
- Show volume of logs of with log level unknown as it was shown in **prior versions**
- When you click to unknown label, show logs with unknown level. Now it's broken 

<img width="1852" height="823" alt="Image" src="https://github.com/user-attachments/assets/9b8daea2-006f-4d00-ae3d-5dc466fa88c2" />

<img width="1860" height="713" alt="Image" src="https://github.com/user-attachments/assets/d413383b-d919-438a-a0d9-b7b902653036" />

gtk-grafana:
Thank you @4strodev for reporting, this does appear to be a regression when `discover_log_levels` is disabled. I'll take a look at this ASAP and will try to get a fix in the next release.

As a workaround I would suggest enabling `discover_log_levels`, even if it fails to detect the level it should add the unknown value to the `detected_level` field.

4strodev:
This happens even with `discover_log_levels` enabled. And when you use `unknown` as a value to filter log levels it shows nothing too.

4strodev:
My team and I noticed that you have this fix almost done. Do know more or less when this will be released?

gtk-grafana:
@4strodev by Monday or Tuesday next week at the latest

gtk-grafana:
@4strodev I wasn't able to reproduce this without disabling `discover_log_levels`, when `discover_log_levels` is enabled every log line gets a detected_level value, but it could be a Loki side bug that has been fixed in the Loki version I'm testing with. Are you running an older version of Loki? 

Otherwise the Logs Drilldown fix has been merged and will go out with our next 1.0.31 release. Please re-open or create another issue if you are still seeing this behavior in 1.0.31.

PR #1630, opened 2025-11-10
fix: logs volume not showing logs without detected_level

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1622

When missing the detected_level label in the data frame from the response (i.e. `discover_log_levels` is disabled), unknown logs would not show in the log volume panel:
image

And when clicking on the "logs" label in the logs volume legend, a broken filter would get added to the query, returning no results in the subsequent query. I also fixed the label getting generated so it toggles like other levels, and shows with label of `logs` instead of `""`

image

I introduced this regression in https://github.com/grafana/logs-drilldown/pull/1540, and while `discover_log_levels` is currently a required configuration option for Logs Drilldown the fix is simple enough to warrant merging this patch. 

To test: disable `discover_log_levels` in the loki config or runtime config:
e.g.
```
configs:
  '1':
    log_push_request: false
    log_push_request_streams: false
overrides:
  '1':
    allow_structured_metadata: true
    pattern_persistence_enabled: true
    volume_enabled: true
    discover_log_levels: false
```

### DIFF
--- a/config/loki-overrides.yaml
+++ b/config/loki-overrides.yaml
@@ -7,3 +7,5 @@ overrides:
     allow_structured_metadata: true
     pattern_persistence_enabled: true
     volume_enabled: true
+    retention_period: 30d
+    max_query_length: 0d
--- a/smoke-tests/play.spec.ts
+++ b/smoke-tests/play.spec.ts
@@ -17,10 +17,17 @@ test.describe('play', () => {
     explorePage.echoConsoleLogsOnRetry();
   });
 
-  test('can load', async ({ page }) => {
+  test('app can load', async ({ page }) => {
     await page.goto('/a/grafana-lokiexplore-app/explore');
-    await page.pause();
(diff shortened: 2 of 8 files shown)


L2D2Grafana (commented):
Code looks good! I'm not sure I'm testing this properly. I can find unknown logs and they filter properly. However they show up as unknown instead of logs. Is there something else I should try to see logs instead of unknown? 

Image

gtk-grafana:
@L2D2Grafana you need to disable `discover_log_levels` in the Loki config, otherwise the `detected_level` metadata field is always populated with a level value.

`unknown` means we attempted to detect the level, but failed, "logs" just means we never even tried (and is just a nicer label for an empty string).

I updated the description to include better instructions on how to test. My bad yo!

L2D2Grafana (approved):
Seeing the logs level now with discover_log_levels: false ty! Image

Issue #1634, opened 2025-11-13
[FEAT]: support `avg_over_time` for `int` fields


gtk-grafana:
**Description of the new feature**
Allow users to toggle int fields to be displayed as `avg_over_time` queries.

**Questions**
Allow toggling on individual panels or all int fields at once?
Use cardinality sample from detected_fields call to allow users to set a cardinality limit for ints?
Or should the problem be addressed in Loki, and users should use `float` for integer fields that should have a value on the y-axis, and `int` fields are used to denote integer labels?

gtk-grafana:
`float` should be used instead of `int` if you want to plot the value on the y-axis by default for all users, but I think it's reasonable to support per-panel overrides.

PR #1637, opened 2025-11-14
feat(fields): support `avg_over_time` for `int` fields

gtk-grafana:
Allows users to plot `int` fields as avg_over_time queries:
image
image

Notes for reviewers:
The iconography is not great, and kinda conflicts with the time series | histogram toggle. Not sure what else to use though.
image

This setting doesn't persist across route changes or page refreshes, something that can be followed up in a later PR.

This setting isn't very discoverable, but the use-case is a bit esoteric so I think it makes sense to put it in the panel menu instead of exposing it more prominently. We can check the usage stats and re-evaluate at a later time.

Fixes: #1634

### DIFF
--- a/src/Components/Panels/PanelMenu.tsx
+++ b/src/Components/Panels/PanelMenu.tsx
@@ -317,7 +317,7 @@ function addToggleQueryType(items: PanelMenuItem[], sceneRef: PanelMenu) {
       }
       onSwitchQueryTypeTracking(newQueryType);
     },
-    text: isAvgQuery ? 'Plot series' : 'Plot values',
+    text: isAvgQuery ? 'Plot series' : 'Plot average',
   });
 }
 
--- a/tests/exploreServicesBreakDown.spec.ts
+++ b/tests/exploreServicesBreakDown.spec.ts
@@ -1966,7 +1966,7 @@ test.describe('explore services breakdown page', () => {
     // Open menu
     await page.getByTestId('data-testid Panel menu values').click();
     // Convert panel to avg_over_time query
-    await page.getByTestId('data-testid Panel menu item Plot values').click();
+    await page.getByTestId('data-testid Panel menu item Plot average').click();
     // Assert the last request is avg_over_time
     await expect
       .poll(() => {
--- a/src/Components/Panels/PanelMenu.test.tsx
+++ b/src/Components/Panels/PanelMenu.test.tsx
@@ -419,7 +419,7 @@ describe('PanelMenu', () => {
       expect(timeseriesItem?.iconClassName).toBe('chart-line');
     });
 
-    it('should show plot values option for int fields', () => {
+    it('should show plot average option for int fields', () => {
       const menu = new PanelMenu({
         addInvestigationsLink: false,
         panelType: TimeSeriesPanelType.histogram,
@@ -435,7 +435,7 @@ describe('PanelMenu', () => {
       menu.activate();
 
       const items = menu.state.body?.state.items;
-      const timeseriesItem = items?.find((item: PanelMenuItem) => item.text === 'Plot values');
+      const timeseriesItem = items?.find((item: PanelMenuItem) => item.text === 'Plot average');
 
       expect(timeseriesItem?.iconClassName).toBe('heart-rate');
     });


src/Components/Panels/PanelMenu.tsx
```diff
@@ -274,43 +285,66 @@ function addCollapsableItem(items: PanelMenuItem[], menu: PanelMenu) {
...
+ * "int" fields are ambiguous if they should be count_over_time or avg queries, so we allow the user to toggle individual panels between avg and count queries
+ * @param items
+ * @param sceneRef
+ */
+function addToggleQueryType(items: PanelMenuItem[], sceneRef: PanelMenu) {
+  const vizPanelWrapper = sceneGraph.getAncestor(sceneRef, FieldsVizPanelWrapper);
+  const isAvgQuery = vizPanelWrapper.state.queryType === TimeSeriesQueryType.avg;
+
+  items.push({
+    iconClassName: 'heart-rate',
```
L2D2Grafana, line 299: We could try `chart-line`

src/services/queryRunner.ts
```diff
@@ -0,0 +1,41 @@
+import { sceneGraph, SceneObject, SceneQueryRunner } from '@grafana/scenes';
+
+import { FieldValuesBreakdownScene } from '../Components/ServiceScene/Breakdowns/FieldValuesBreakdownScene';
+import { LabelValuesBreakdownScene } from '../Components/ServiceScene/Breakdowns/LabelValuesBreakdownScene';
+import { logger } from './logger';
+import { interpolateExpression } from './query';
+import { getQueryRunnerFromChildren } from './scenes';
+
+export function getQueryRunnerFromSceneGraph(sceneRef: SceneObject) {
```
L2D2Grafana, line 9: Should we note somewhere that this is for PanelMenus

src/services/fields.ts
```diff
@@ -373,19 +373,25 @@ export function isAvgField(fieldType: DetectedFieldType | undefined) {
   return fieldType === 'duration' || fieldType === 'bytes' || fieldType === 'float';
 }
 
+export const buildAvgOverTimeFloatExpr = (options: LogsQueryOptions, optionValue: string) => {
+  return `avg_over_time(${getLogsStreamSelector(options)} | unwrap ` + optionValue + ` | __error__="" [$__auto]) by ()`;
+};
+
```
L2D2Grafana, line 379: How did you get these queries? why does it ned unwrap by ()

tests/exploreServicesBreakDown.spec.ts
```diff
@@ -1946,6 +1954,46 @@ test.describe('explore services breakdown page', () => {
     await expect(explorePage.getAllPanelsLocator()).toHaveCount(2);
   });
 
+  test.only('int fields should allow avg_over_time queries', async ({ page }) => {
```
L2D2Grafana: Hello .only()

L2D2Grafana (commented):
This looks cool 😎 . I get a little lost switching between plot values vs plot series then it becomes histogram and time series. Should plot values alway be Time series? It also seems to affect all the int panels. I'm not sure if that's a bug 🐛  or not. I could see people liking that.

https://github.com/user-attachments/assets/647570cf-4b0a-42b1-83da-79db1d623df2

tests/exploreServicesBreakDown.spec.ts
gtk-grafana: oopsie

src/services/fields.ts
gtk-grafana: This was preexisting code, I thought I needed to re-create the queries so I broke this out a bit more but the query runners don't like having their state changed so I decided to just re-build the panels.

[avg_over_time](https://grafana.com/docs/loki/latest/query/metric_queries/#unwrapped-range-aggregations) and other unwrapped aggregations are how Loki displays extracted labels as sample values in metric queries, so you always need to unwrap the value when aggregating by stuff that isn't count/rate.

by () just means aggregate by `Value` afaik, since unwrap aggregations should always return a single value series we don't need to specify which field we are aggregating by.
TIL: `by (Value)` also works apparently

src/services/queryRunner.ts
gtk-grafana: It's also called in the FieldsAggregatedBreakdownScene

src/Components/Panels/PanelMenu.tsx
gtk-grafana: We're already using for the "Time series" in the panel menu

src/services/fields.ts
gtk-grafana: Although actually `by (kajshdkjashjdka)` also works, so I'm guessing there's some automagic cleaning up our messes if you try to aggregate by something that isn't a field it will return the unwrap value.

Further learning 🌈 
[Here's the query without the aggregation](https://play.grafana.org/explore?schemaVersion=1&panes=%7B%221pp%22:%7B%22datasource%22:%22ddhr3fttaw8aod%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22avg_over_time%28%7Bservice_name%3D%5C%22tempo-distributor%5C%22%7D%20%7C%20logfmt%20%7C%20bytes%21%3D%5C%22%5C%22%20%7C%20unwrap%20bytes%28bytes%29%20%7C%20__error__%3D%5C%22%5C%22%20%5B$__auto%5D%29%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22ddhr3fttaw8aod%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%22now-1s%22,%22to%22:%22now%22%7D,%22compact%22:false%7D%7D&orgId=1)

[And then with it](https://play.grafana.org/explore?schemaVersion=1&panes=%7B%221pp%22:%7B%22datasource%22:%22ddhr3fttaw8aod%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22avg_over_time%28%7Bservice_name%3D%5C%22tempo-distributor%5C%22%7D%20%7C%20logfmt%20%7C%20bytes%21%3D%5C%22%5C%22%20%7C%20unwrap%20bytes%28bytes%29%20%7C%20__error__%3D%5C%22%5C%22%20%5B$__auto%5D%29%20by%20%28%29%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22ddhr3fttaw8aod%22%7D,%22editorMode%22:%22code%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%22now-1s%22,%22to%22:%22now%22%7D,%22compact%22:false%7D%7D&orgId=1)

Basically if we don't add the `by ()` each log is returned as a separate series, meaning we'll hit our max series limit in no time (why those queries ^ have a time range of 1s), but the unwrap still works and we're getting a time series result with the y-axis value getting pulled from the bytes label. Neat! 🤖 📷

gtk-grafana:
> I get a little lost switching between plot values vs plot series then it becomes histogram and time series. 
> Should plot values alway be Time series? It also seems to affect all the int panels. I'm not sure if that's a bug 🐛 or not. I could see people liking that.

Right now all avg_over_time queries use the time series or histogram panel depending on what the user set last (it's saved in local storage for the user, not specific to a panel or datasource). All avg_over_time queries should always get the same panel type, if not then we have a bug

src/Components/Panels/PanelMenu.tsx
```diff
@@ -274,43 +285,70 @@ function addCollapsableItem(items: PanelMenuItem[], menu: PanelMenu) {
...
+        sceneRef,
+        (o) => o instanceof FieldsAggregatedBreakdownScene,
+        FieldsAggregatedBreakdownScene
+      );
+      if (fieldsAggregatedBreakdownScene) {
+        fieldsAggregatedBreakdownScene.rebuildChangedPanels('queryType');
+      }
+      onSwitchQueryTypeTracking(newQueryType);
+    },
+    text: isAvgQuery ? 'Plot series' : 'Plot values',
```
matyax: We have:
- Histogram
- Time series
- Plot values / Plot series

The first two are clear, but plot values not so much, and there's no indication that it's running an average_over_time. Maybe something along the lines of "Plot average" or "Average series/values".

gtk-grafana: Changed to "plot average" in https://github.com/grafana/logs-drilldown/pull/1637/commits/f10c9895881cda3f6440cbdf4fb573787e80f319, which does look a bit better. 
image

PR #1649, opened 2025-11-24
feat(links): add sortOrder support

matyax:
Updates link extensions to include sortOrder.

Drilldown part of https://github.com/grafana/grafana/pull/114350 (requires this in core)

- Updates the extension link to receive sortOrder
- Adds sort order to the copy link to log line, which affects the query direction

### DIFF
--- a/src/services/text.test.ts
+++ b/src/services/text.test.ts
@@ -1,4 +1,4 @@
-import { dateTime, TimeRange } from '@grafana/data';
+import { dateTime, LogsSortOrder, TimeRange } from '@grafana/data';
 
 import { PLUGIN_ID } from './plugin';
 import { generateLogShortlink } from './text';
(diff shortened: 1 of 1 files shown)


gtk-grafana:
Copying links from drilldown via share button or via browser works as expected, but panel menu link does not include sort order.

matyax:
Should be fixed. Also included in links.

src/Components/ServiceScene/LogsPanelScene.tsx
```diff
@@ -382,7 +382,11 @@ export class LogsPanelScene extends SceneObjectBase<LogsPanelSceneState> {
       generateLogShortlink(
         'panelState',
         {
-          logs: { displayedFields: parent.state.displayedFields, id: row.uid },
+          logs: {
+            displayedFields: parent.state.displayedFields,
+            id: row.uid,
+            sortOrder: this.state.sortOrder,
```
gtk-grafana, line 388: I don't think this works as expected, we need to set `direction: 'forward' | 'backward'` in addition or instead of the sortOrder in the panel state.

If you open this explore link up in a private window you'll see that the local storage default is pulled into explore instead of what is in the URL

matyax: Ignore my previous comment. Query direction is set when the query is executed, based on the sort order, is not a property that we set in panel state.

I tested this generating a link in a browser and opening it in another one, and the updated direction was always used. Is not critical, but since it changes query direction I think it's an important attribute to keep in the link.

gtk-grafana: My bad, wrong spot in the code, and didn't have core PR checked out.

Issue #1293, opened 2025-05-21
Ability to line format logs


extrange:
I'm using Vector to send logs in JSON to Loki, and in the Drilldown view they appear like this:

![Image](https://github.com/user-attachments/assets/97e7481c-88fb-4be6-b6a4-f73557a7fb6f)

It's a bit hard to see the individual messages, since most are timestamps/metadata.

Wondering if there's a way to do something like `line_format {{.message}}` within this view, maybe as a default?

JStickler:
There is a new (experimental) [JSON view](https://grafana.com/docs/grafana-cloud/visualizations/simplified-exploration/logs/viewing-json-logs/) that probably solves your problem.

itd-seb:
Is there any way to configure the line format by some workaround for now?

suikast42:
I am ingesting all my logs in ecs log format. This coulüd be ver handy to give the log panels the ability to set a log format. This is my view view if I use custom log format 

<img width="2271" height="871" alt="Image" src="https://github.com/user-attachments/assets/13cc2367-c771-4ccd-a2ee-a09537bc3bc4" />

axsuul:
Yes, I'm also having this problem and need a way to format the logs

PR #1664, opened 2025-12-02
feat: App config default columns

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1293

Requires: [Grafana 12.4+](https://github.com/grafana/grafana/pull/114309) 

Feature flags:
* `kubernetesLogsDrilldown`
* `grafanaAPIServerWithExperimentalAPIs`

(`grafanaAPIServer` was made default in https://github.com/grafana/grafana/pull/81030)

Follow up:
* Add buttons to table viz in https://github.com/grafana/logs-drilldown/pull/1685
* Add tests: https://github.com/grafana/logs-drilldown/pull/1686
* Add tracking events

### Description
This PR adds a new admin configuration page: `/plugins/grafana-lokiexplore-app?page=admin-default-fields` which allows administrators to configure a set of default displayed fields for the logs and table panels for all users of logs drilldown.

### Specification

Default columns will be used in panels whenever the users query contains all labels defined in the record, even If the users query has more labels then defined in the default columns config. But if the admin config has a label defined that is not present in the users query the defaults will not be used. 

If multiple default column records match the users query we will display the columns that contain the most labels in common with the users query. 

If there are multiple default column records with the same count of labels in common with the labels in the users query then we should pull the default columns for the label in the query with the lower index (i.e. the first matching label).

Local user changes to panels (stored in local storage) will overwrite the default columns from the backend.

A button in the logs panel allows users to revert back to the admin defined default columns when the defaults are not applied: 

image

Admin UI:
image

image

image

Example panel output:

image

image

image

image

### DIFF
--- a/src/Components/ServiceScene/LogsTableScene.tsx
+++ b/src/Components/ServiceScene/LogsTableScene.tsx
@@ -304,7 +304,7 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
     const controlsExpanded = parentModel.state.controlsExpanded;
 
     return (
-      <div className={styles.panelWrapper} ref={panelWrap}>
+      <div className={styles.panelWrapper}>
         {!error && (
           <>
             {/* @ts-expect-error todo: fix this when https://github.com/grafana/grafana/issues/103486 is done*/}
@@ -336,23 +336,25 @@ export class LogsTableScene extends SceneObjectBase<LogsTableSceneState> {
                     disabledLineState={!model.state.isDisabledLineState}
                   />
                 )}
-                {dataFrame && (
-                  <TableProvider
-                    controlsExpanded={controlsExpanded}
-                    panelWrap={panelWrap}
-                    addFilter={addFilter}
-                    timeRange={timeRangeValue}
-                    selectedLine={selectedLine}
-                    urlColumns={urlColumns ?? []}
-                    displayFields={parentModel.state.displayedFields}
-                    setUrlColumns={setUrlColumns}
-                    dataFrame={dataFrame}
-                    clearSelectedLine={clearSelectedLine}
-                    setUrlTableBodyState={setUrlTableBodyState}
-                    urlTableBodyState={tableLogLineState ?? getTableLogLine() ?? LogLineState.text}
-                    logsSortOrder={sortOrder}
-                  />
-                )}
+                <div className={styles.tableContainer} ref={panelWrap}>
+                  {dataFrame && (
+                    <TableProvider
+                      controlsExpanded={controlsExpanded}
+                      panelWrap={panelWrap}
+                      addFilter={addFilter}
+                      timeRange={timeRangeValue}
+                      selectedLine={selectedLine}
+                      urlColumns={urlColumns ?? []}
+                      displayFields={parentModel.state.displayedFields}
+                      setUrlColumns={setUrlColumns}
+                      dataFrame={dataFrame}
+                      clearSelectedLine={clearSelectedLine}
+                      setUrlTableBodyState={setUrlTableBodyState}
+                      urlTableBodyState={tableLogLineState ?? getTableLogLine() ?? LogLineState.text}
+                      logsSortOrder={sortOrder}
+                    />
+                  )}
+                </div>
                 {emptyScene && dataFrame && dataFrame.length === 0 && (
                   <NoMatchingLabelsScene.Component model={emptyScene} />
                 )}
@@ -378,6 +380,13 @@ const getStyles = (theme: GrafanaTheme2) => ({
     display: 'flex',
     flexDirection: 'row-reverse',
     justifyContent: 'space-between',
+    height: '100%',
+    flex: 1,
+  }),
+  tableContainer: css({
+    overflow: 'hidden',
+    flex: '1 1 auto',
+    minWidth: 0,
   }),
   panelWrapper: css({
     height: '100%',
--- a/src/Components/Table/Table.tsx
+++ b/src/Components/Table/Table.tsx
@@ -107,10 +107,15 @@ const getStyles = (theme: GrafanaTheme2, height: number, sideBarWidth: number) =
       // Hacky but without inspect turned on the table will change the width of the row on hover, but we don't want the default icons to show
       display: 'none !important',
     },
+    flex: 1,
+    minWidth: 0, // Allows flex child to shrink below content width, preventing overflow
+    overflow: 'hidden',
   }),
   wrapper: css({
     display: 'flex',
+    overflow: 'hidden',
     position: 'relative',
+    flexWrap: 'wrap', // Stop inch worm effect in browsers outside of chrome
   }),
 });
 
--- a/src/Components/Table/TableWrap.tsx
+++ b/src/Components/Table/TableWrap.tsx
@@ -1,6 +1,5 @@
 import React, { useCallback, useState } from 'react';
 
-import { css } from '@emotion/css';
 import { useResizeObserver } from '@react-aria/utils';
 
 import {
@@ -13,12 +12,10 @@ import {
   LogsSortOrder,
 } from '@grafana/data';
 
-import { CONTROLS_WIDTH, CONTROLS_WIDTH_EXPANDED } from '../ServiceScene/LogListControls';
 import { useQueryContext } from 'Components/Table/Context/QueryContext';
 import { LogLineState, TableColumnContextProvider } from 'Components/Table/Context/TableColumnsContext';
 import { Table } from 'Components/Table/Table';
 import { FieldNameMeta, FieldNameMetaStore } from 'Components/Table/TableTypes';
-import { logsControlsSupported } from 'services/panel';
 
 export type SpecialFieldsType = {
   body: FieldWithIndex;
@@ -41,12 +38,6 @@ interface TableWrapProps {
   urlTableBodyState?: LogLineState;
 }
 
-const getStyles = () => ({
-  section: css({
-    position: 'relative',
-  }),
-});
-
 export const TableWrap = (props: TableWrapProps) => {
   const { logsFrame } = useQueryContext();
 
@@ -68,7 +59,6 @@ export const TableWrap = (props: TableWrapProps) => {
     ref: props.panelWrap,
   });
 
-  const styles = getStyles();
   const timeZone = getTimeZone();
 
   // This function is called when we want to grab the column names that are currently stored in the URL.
@@ -123,31 +113,26 @@ export const TableWrap = (props: TableWrapProps) => {
     setSpecialFieldMeta(active, specialFields, pendingLabelState);
   }
 
-  const logControlOptionsWidth = props.controlsExpanded ? CONTROLS_WIDTH_EXPANDED : CONTROLS_WIDTH;
-  const tableWidth = panelWrapSize.width - 25 + (logsControlsSupported ? logControlOptionsWidth * -1 : 0);
-
   return (
-    <section className={styles.section}>
-      <TableColumnContextProvider
-        setUrlTableBodyState={props.setUrlTableBodyState}
+    <TableColumnContextProvider
+      setUrlTableBodyState={props.setUrlTableBodyState}
+      logsFrame={logsFrame}
+      initialColumns={pendingLabelState}
+      setUrlColumns={props.setUrlColumns}
+      urlColumns={props.urlColumns}
+      displayFields={props.displayFields}
+      clearSelectedLine={props.clearSelectedLine}
+      urlTableBodyState={props.urlTableBodyState}
+    >
+      <Table
         logsFrame={logsFrame}
-        initialColumns={pendingLabelState}
-        setUrlColumns={props.setUrlColumns}
-        urlColumns={props.urlColumns}
-        displayFields={props.displayFields}
-        clearSelectedLine={props.clearSelectedLine}
-        urlTableBodyState={props.urlTableBodyState}
-      >
-        <Table
-          logsFrame={logsFrame}
-          timeZone={timeZone}
-          height={panelWrapSize.height - 50}
-          width={tableWidth}
-          labels={labels}
-          logsSortOrder={props.logsSortOrder}
-        />
-      </TableColumnContextProvider>
-    </section>
+        timeZone={timeZone}
+        height={panelWrapSize.height}
+        width={panelWrapSize.width}
+        labels={labels}
+        logsSortOrder={props.logsSortOrder}
+      />
+    </TableColumnContextProvider>
   );
 };
 


src/Components/AppConfig/DefaultColumns/Unsupported.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <section>
+      <h2>Default columns is not supported on this instance of Grafana</h2>
+      {!config.featureToggles.kubernetesLogsDrilldown ||
+        (!config.featureToggles.grafanaAPIServerWithExperimentalAPIs && (
+          <p>
+            Default columns requires <code>kubernetesLogsDrilldown</code> and{' '}
+            <code>grafanaAPIServerWithExperimentalAPIs</code> feature flags to be enabled.
+          </p>
+        ))}
+      {config.buildInfo.version < '12.4' && <p>Default columns requires Grafana 12.4 or greater.</p>}
```
matyax, line 16: I'm not seeing any of these messages 

imagen

Also I would suggest an alternative ui, like.

```

## Default Columns

Default columns requires `kubernetesLogsDrilldown` and `grafanaAPIServerWithExperimentalAPIs` feature flags to be enabled.

Default columns requires Grafana 12.4 or greater
```

Additionally, do we have control to decide whether to render the Tab or not, so users don't reach this page while not meeting requirements?

src/module.tsx
```diff
@@ -38,13 +38,26 @@ const AppConfig = lazy(async () => {
...
+    body: AppConfig,
+    icon: 'cog',
+    id: 'configuration',
+    title: 'Configuration',
+  })
+  .addConfigPage({
+    body: DefaultColumnsConfig,
+    icon: 'columns',
+    id: 'admin-default-columns',
+    title: 'Default columns',
```
matyax: `Default columns` doesn't sound great to me for this option. What about something like Visualization options, Logs preferences, etc.

src/Components/AppConfig/DefaultColumns/Unsupported.tsx
matyax: The PR description also mentions grafanaAPIServer

src/Components/AppConfig/DefaultColumnsAddRecord.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+import { Button } from '@grafana/ui';
+
+import { useDefaultColumnsContext } from './DefaultColumnsContext';
+
+interface Props {}
+
+export function DefaultColumnsAddRecord({}: Props) {
```
matyax:
```suggestion
export function DefaultColumnsAddRecord() {
```

src/Components/AppConfig/DefaultColumnsFooter.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { DefaultColumnsAddRecord } from './DefaultColumnsAddRecord';
+import { DefaultColumnsSubmit } from './DefaultColumnsSubmit';
+
+export function DefaultColumnsFooter() {
+  const styles = useStyles2(getStyles);
+
+  return (
+    <footer className={styles.wrap}>
+      <DefaultColumnsAddRecord />
```
matyax: Can we add an initial row by default?

imagen

Also, for "Add record", I don't think we need to be that _databasey_. "+ Add" should be enough.

Finally, can we hide the option instead of disabling it?

src/Components/AppConfig/DefaultColumns/Labels.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const { localDefaultColumnsState, dsUID } = useDefaultColumnsContext();
+  const styles = useStyles2(getStyles);
+  if (!localDefaultColumnsState?.[dsUID]) {
+    throw new Error('Records::missing localDefaultColumnsState');
+  }
+  const record = localDefaultColumnsState[dsUID].records[recordIndex];
+  return (
+    <>
+      <h5 className={styles.labelTitle}>
+        Labels match
```
matyax, line 27: What do you think of bumping `service_name` to the top? Or making it a default value if it exists.

imagen

src/Components/AppConfig/DefaultColumns/Labels.tsx
```diff
@@ -0,0 +1,68 @@
...
+            </div>
+          </div>
+        );
+      })}
+
+      <DefaultColumnsAddLabel recordIndex={recordIndex} />
+    </>
+  );
+}
+
```
matyax, line 48: I couldn't find an option to include the log line body

imagen

matyax:
Other than any implementation comments, I absolutely love this! Fantastic work @gtk-grafana and @L2D2Grafana

gtk-grafana:
@matyax thanks for taking a look but this branch is no where near ready quite yet, just the API in core

src/Components/AppConfig/DefaultColumns/Unsupported.tsx
gtk-grafana: Copy has been cleaned up, eslint was adding parentheses in the wrong spot which broke the messages showing up, and `grafanaAPIServer` was made default so we only have 2 flags to deal with now.

src/Components/ServiceSelectionScene/ServiceSelectionScene.tsx
```diff
@@ -1000,7 +1017,8 @@ export class ServiceSelectionScene extends SceneObjectBase<ServiceSelectionScene
   }
 
   private getLogExpression(labelName: string, labelValue: string, levelFilter: string) {
-    return `{${labelName}=\`${labelValue}\` , ${VAR_LABELS_REPLICA_EXPR} }${levelFilter}`;
+    // @todo if config for this labelName/labelValue add parsers
+    return `{${labelName}=\`${labelValue}\` , ${VAR_LABELS_REPLICA_EXPR} }${levelFilter} ${DETECTED_FIELDS_MIXED_FORMAT_EXPR_NO_JSON_FIELDS}`;
```
gtk-grafana: can we avoid adding parsers unless the feature is enabled? We probably can't add parsers only if there is an API response for this service

src/Components/AppConfig/DefaultColumns/Labels.tsx
gtk-grafana: Moved to top in https://github.com/grafana/logs-drilldown/pull/1664/commits/d148115c6c9db31ed3fec83fb5e95e92cb06952d

src/Components/AppConfig/DefaultColumns/Columns.tsx
```diff
@@ -0,0 +1,129 @@
...
+  recordIndex: number;
+}
+
+export function Columns({ recordIndex, containerDragging }: Props) {
+  const { dsUID, records, setRecords } = useDefaultColumnsContext();
+  const record = records?.[recordIndex];
+  const columns = record?.columns ?? [];
+  const styles = useStyles2(getStyles, containerDragging);
+
+  if (!record) {
```
matyax, line 25: Everywhere where you pass a recordIndex I would pass the record as a prop, so you don't have to go through this.
Very low prio anyway, as this is config, but I think it would result in a cleaner component.

src/Components/AppConfig/DefaultColumns/CollapsibleFields.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const record = records[recordIndex];
+  const expandedRecordIdx = expandedRecords?.findIndex((i) => i === recordIndex);
+  const expandedRecord = expandedRecordIdx !== -1 ? expandedRecords[expandedRecordIdx] : undefined;
+  const isOpen = expandedRecord === recordIndex;
+  return (
+    <Collapse
+      className={styles.collapse}
+      isOpen={isOpen}
+      label={<RecordsCollapsibleLabel record={record} />}
```
matyax: Also low prio/just an idea, but if we only showed the fields in the label when it's collapsed, and removed then on expansion, I think it would be cool. Just to not show the same thing twice.


https://github.com/user-attachments/assets/1086a76d-07a2-4c0e-afc0-cb19da6fe3b9

src/Components/AppConfig/DefaultColumns/Context.tsx
```diff
@@ -0,0 +1,182 @@
...
+    // Use a local reference to records to force useMemo to re-run when the localDefaultColumnsState reference changes
+    const records = localDefaultColumnsState?.[dsUID]?.records ?? null;
+    const invalidRecords = records?.filter((r) => isRecordInvalid(r));
+    const hasDuplicates = records ? recordsHaveDuplicates(records) : false;
+    const hasInvalidRecords = (invalidRecords && invalidRecords?.length > 0) ?? false;
+    const isInvalid = hasInvalidRecords || hasDuplicates;
+    const hasPendingChanges = isDefaultColumnsStateChanged(records, apiRecords) ?? false;
+
+    return { hasDuplicates, hasInvalidRecords, isInvalid, hasPendingChanges };
+    // ¡if you just pass in records to this dep array we won't run validation on changes to record labels/columns!
```
matyax:
```suggestion
    // if you just pass in records to this dep array we won't run validation on changes to record labels/columns!
```

src/Components/AppConfig/DefaultColumns/DefaultColumns.tsx
```diff
@@ -0,0 +1,92 @@
...
+      if (!defaultColumnsFromAPI.metadata.name) {
+        const error = new Error('DefaultColumns::Unexpected result for defaultColumnsFromAPI - missing metadata name');
+        logger.error(error);
+        throw error;
+      }
+      if (defaultColumnsFromAPI.metadata.name !== dsUID) {
+        const error = new Error('DefaultColumns::Unexpected result for defaultColumnsFromAPI - invalid datasource uid');
+        logger.error(error);
+        throw error;
+      }
```
matyax, line 56: If this would kill the UI, I'd suggest a different approach. E.g. temporary network issue resulting in a broken UI.

src/Components/AppConfig/DefaultColumns/DefaultColumns.tsx
```diff
@@ -0,0 +1,92 @@
...
+    defaultColumnsFromAPI,
+    defaultColumnsAPIError,
+    isLoading,
+    dsUID,
+    setApiDefaultColumnsState,
+  ]);
+
+  // @todo if grafana 12.4-pre is run and it doesn't have the API changes isLoading is always true and we get stuck in a loading state
+  // Shouldn't be a problem as long as we make sure the API changes are deployed to cloud before releasing, unless OSS users are running pre-release versions.
+  if (isLoading) {
```
matyax, line 85: Can we prevent this for anyone with an older Grafana?

src/Components/AppConfig/DefaultColumns/Fields.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+  };
+
+  return (
+    <div className={styles.fieldsContainer}>
+      <ColumnsDragContext recordIndex={recordIndex} />
+
+      <Button
+        disabled={!recordHasValues}
+        tooltip={
```
matyax, line 49: I love this, but we should find a better place for this text. Even after looking at the implementation, it took me more than a minute to find where the tooltip was going to be displayed, as it was an unexpected location.

imagen

We could probably use an Alert component for this.

src/Components/ServiceScene/LogsListScene.tsx
```diff
@@ -275,15 +368,56 @@ export class LogsListScene extends SceneObjectBase<LogsListSceneState> {
...
+      },
+      output: {
+        user: userDisplayedFieldsUrl,
+        display: displayedFieldsUrl,
+      },
+      storage: {
+        user: userDisplayFieldsFromStorage,
+        display: displayFieldsFromStorage,
+      },
+    });
```
matyax:
```suggestion
```

matyax:
One thing that keeps happening to me is that the collapsible closes after clicking something: reordering, removing a field.

matyax:
It looks and works great! There's a lot to review, so I'll do another pass later.

src/Components/AppConfig/DefaultColumns/DefaultColumns.tsx
gtk-grafana: network errors would return an error and not the `currentData` from the `useGetLogsDrilldownDefaultColumnsQuery` hook, there would have to be breaking changes in the API in order to throw this error. If the API were in such a state where we were sending a different record then what is requested we want the UI to be broken so users don't update the wrong record.

src/Components/AppConfig/DefaultColumns/DefaultColumns.tsx
gtk-grafana: I think we already have with the version check in Config.tsx, I think this comment is stale and can be removed now.

src/Components/AppConfig/DefaultColumns/Fields.tsx
gtk-grafana: Fixed in https://github.com/grafana/logs-drilldown/pull/1664/commits/ba2b2a39ba5d656e2f67e5b2297ac4ff4b25d36c

But only for the log line case, the empty column throwing an alert (causing layout to shift) is too loud for something that is expected

gtk-grafana:
> One thing that keeps happening to me is that the collapsible closes after clicking something: reordering, removing a field.

@matyax Hmm you must be clicking on the collapsible header (inline with "Display columns")?

Cleaned up in https://github.com/grafana/logs-drilldown/pull/1664/commits/a2fb10d1c9b6ea95fb0790ad1be749d2a813f2ff by adding a border so the division between the collapsible header and the body is more apparent. 
image

src/Components/AppConfig/DefaultColumns/CollapsibleFields.tsx
gtk-grafana: Cleaned up in https://github.com/grafana/logs-drilldown/commit/a2fb10d1c9b6ea95fb0790ad1be749d2a813f2ff

src/Components/AppConfig/DefaultColumns/Columns.tsx
gtk-grafana: I think the reasoning was that sometimes you need the index in order to splice it out of the full array, so we'd need props for index and the actual record, so I opted to just pass the minimal set of props and let the children components grab what they need from the context

src/Components/AppConfig/DefaultColumns/Config.tsx
```diff
@@ -0,0 +1,67 @@
...
+const Config = () => {
+  const dsUID = getLastUsedDataSourceFromStorage() ?? getDefaultDatasourceFromDatasourceSrv();
+  const styles = useStyles2(getStyles);
+  if (!dsUID) {
+    return <NoLokiSplash />;
+  }
+  if (
+    !config.featureToggles.kubernetesLogsDrilldown ||
+    !config.featureToggles.grafanaAPIServerWithExperimentalAPIs ||
+    config.buildInfo.version < '12.4'
```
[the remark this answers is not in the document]
L2D2Grafana: This seem like a good comment to fix.

src/Components/AppConfig/DefaultColumns/Context.tsx
```diff
@@ -0,0 +1,182 @@
...
+        setExpandedRecords: handleSetExpandedRecords,
+      }}
+    >
+      {children}
+    </Context.Provider>
+  );
+};
+
+export const useDefaultColumnsContext = () => {
+  return useContext(Context);
```
L2D2Grafana, line 181: Hello Context my old friend 🤩

src/Components/AppConfig/DefaultColumns/DeleteRecord.tsx
```diff
@@ -0,0 +1,44 @@
...
+export function DeleteRecord({ recordIndex }: Props) {
+  const styles = useStyles2(getStyles);
+  const { records, setRecords } = useDefaultColumnsContext();
+
+  if (!records) {
+    return null;
+  }
+
+  const deleteRecord = () => {
+    records.splice(recordIndex, 1);
```
L2D2Grafana, line 22: I do like always creating a new object, but this works.

L2D2Grafana (commented):
This PR is 🦸 🪄 

I was able to get this PR running by:

- Adding the feature flags
- Updating the grafana version to main
- Adding an alias for i18n in webpack

The 404 I saw onload of http://localhost:3000/plugins/grafana-lokiexplore-app?page=admin-default-fields&from=now-24h&to=now&timezone=browser makes sense because there is nothing at the endpoint yet.

I was able to set default fields and see the display columns in the table and panel!

Image

Image

Some UX suggestions:
- Adding a display column is required, but not outlined in pink like the other inputs and easy to miss.
Image
- Could the buttons be sticky or at the top? If you return to the page they get buried at the bottom.
Image

gtk-grafana:
> Adding a display column is required, but not outlined in pink like the other inputs and easy to miss.

Good catch @L2D2Grafana, fixed in https://github.com/grafana/logs-drilldown/pull/1664/commits/96f8a11a6a321a68d0fe5a593e2e0d1807c27fe8,

>  Could the buttons be sticky or at the top? If you return to the page they get buried at the bottom.

Yuppers, added in https://github.com/grafana/logs-drilldown/pull/1664/commits/cdf7b67ab44cca4f03095735b0e3dc252c45691d

src/Components/AppConfig/DefaultColumns/DeleteRecord.tsx
matyax: Yeah, better than mutating.

matyax:
I personally like how this looks, but I would suggest to remove the shadow, as there's no other usage of shadows in settings, so we would be drifting a lot in terms of designs from the current design system.

imagen

matyax:
imagen

:cheff-kiss:

src/services/analytics.ts
```diff
@@ -66,6 +66,7 @@ export const USER_EVENTS_ACTIONS = {
     level_in_logs_volume_clicked: 'level_in_logs_volume_clicked',
     // Clear all displayed fields
     logs_clear_displayed_fields: 'logs_clear_displayed_fields',
+    logs_show_backend_fields: 'logs_show_backend_fields',
     // Fires when logs panel query returns successfully
```
matyax, line 72: Can we add some analytics to the configuration page?

matyax (approved):
Really looking forward to releasing this!

src/services/analytics.ts
gtk-grafana: Yeah I'll create a follow up PR

PR #1665, opened 2025-12-03
feat(table): use display fields

L2D2Grafana:
Companion PR to core https://github.com/grafana/grafana/pull/113892

- Uses displayedFields in the logs panel and table. 
- Removed urlColumns
- Migrates the urlColumns in a url to displayedFields


matyax:
Would it be possible to not carry over detected_level to the logs viz, in the same way we filter OTel attributes when going to the table? I'm not entirely convinced this is worth extra code, since "Reset" will take you where you need to be.

https://github.com/user-attachments/assets/134828db-998f-45d4-9e39-da793de0c466

L2D2Grafana:
Closing this for now, as we rethink a wholistic refactor that takes into account the new default columns and syncing timestamps and detected level.

PR #1666, opened 2025-12-04
feat(embedded): allow hiding time picker

gtk-grafana:
And fix bug introduced by routing in embedding application, and fixing runtime error from i18n library not being inited


matyax:
You need to pass the new prop in the test embedded scene.

imagen

Issue #1667, opened 2025-12-05
Declarative configuration


LilWatson:
First: I love this plugin, its makes browsing logs way easier!

Is there a way to set the configuration via a configuration file of even via terraform (to e.g. disable the patters panel on a multi tenant loki)? 
Or is the only option a HTTP request to `/api/plugins/grafana-lokiexplore-app/settings `? Which would make provisioning quiet cumbersome.

gtk-grafana:
@LilWatson good point, as of Loki 3.6 there is the new drilldown-limits API endpoint which should allow the frontend app to disable patterns automatically if the `pattern_ingester_enabled` limits_config is set as false in the Loki config. I'll take a quick look at that now

gtk-grafana:
@LilWatson this should be resolved in the next release (1.0.32) if you're running Loki 3.6+ it'll automatically remove the patterns from the UI if Loki has `pattern_ingester_enabled` set to false

PR #1669, opened 2025-12-05
feat(lokiConfig): disable patterns if pattern_ingester_enabled is false

gtk-grafana:
Checks the Loki config before adding the pattern query runner.

To test: 
* Check out this branch
* in `config/loki-config.yaml` change `pattern_ingester.enabled` to `false`
```
pattern_ingester:
  enabled: false
```
* Re-run `yarn server`
* Patterns tabs should no longer show up.

Note: requires Loki 3.6+

Fixes: https://github.com/grafana/logs-drilldown/issues/1667

### DIFF
--- a/config/loki-overrides.yaml
+++ b/config/loki-overrides.yaml
@@ -8,13 +8,13 @@ configs:
 overrides:
   '1':
     allow_structured_metadata: true
-    #    pattern_persistence_enabled: true
+    pattern_persistence_enabled: true
     volume_enabled: true
     retention_period: 30d
     max_query_length: 0d
   '2':
     allow_structured_metadata: true
-    #    pattern_persistence_enabled: false
+    pattern_persistence_enabled: false
     volume_enabled: true
     retention_period: 30d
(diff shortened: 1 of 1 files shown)


docker-compose-syslog.dev.yaml
```diff
@@ -22,7 +22,7 @@ services:
       - 'host.docker.internal:host-gateway'
 
   loki:
-    image: grafana/loki:k276-c416352@sha256:deee0771b54d6ff5c1cc826b5eb4900605e7f7977a4dd709c9ef47894138b0b8
+    image: grafana/loki:3.6
```
L2D2Grafana, line 25: Is it okay that renonvate keeps updating this?

L2D2Grafana (approved):
LGTM

Image

docker-compose-syslog.dev.yaml
gtk-grafana: Rennovate is updating our loki deps? No it should not be, (but I made this update)

PR #1678, opened 2025-12-16
fix(EmbeddedLogs): use custom value prefix for regular expressions

matyax:
This PR updates the embedded logs to use custom value prefixes for regular expression values, matching the behavior of the application.

https://github.com/user-attachments/assets/92cf06fd-768a-48c3-b3dc-5374252fe389

Taken from:

https://github.com/grafana/logs-drilldown/blob/main/src/services/query.ts#L76

imagen

### DIFF
--- a/src/services/variableHelpers.ts
+++ b/src/services/variableHelpers.ts
@@ -106,6 +106,10 @@ export const operatorFunction = function (variable: AdHocFiltersVariable) {
     const meta: Record<string, string> = wip.meta;
     const type = meta.type;
 
+    if (type === 'int') {
+      return [...operators, ...numericOperators];
+    }
+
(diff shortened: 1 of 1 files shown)


gtk-grafana:
This should be good but we've got some bugs here if you take this same query to explore and then back to logs drilldown we see
image

gtk-grafana:
I think we probably want to split out into multiple pills but feel free to merge this and I'll follow up with a fix that solves the problem I mentioned above.

matyax:
At least with this combo of equality and regex and I'm being able to go back and forth, but definitely, let's follow this up.

https://github.com/user-attachments/assets/864fef97-d264-4ef8-b3f8-315866575295

Issue #1681, opened 2025-12-16
[FEAT]: Embedded logs drilldown - support field selectors


gtk-grafana:
**Description of the new feature**
Support field selectors in embedded logql queries.

**Acceptance Criteria**
A concise description of requirements.

PR #1683, opened 2025-12-16
feat(embedded): field filters support

gtk-grafana:
Fixes https://github.com/grafana/logs-drilldown/issues/1681

### DIFF
--- a/src/Components/Pages.tsx
+++ b/src/Components/Pages.tsx
@@ -77,7 +77,7 @@ function getEmbeddedScene() {
   const initialStart = 'now-15m';
   const initialEnd = 'now';
   const query =
-    '{service_name=~"tempo-distributor|tempo-ingester"} | user=~"[0][3]428" |~ "(?i)Error" | json | logfmt | drop __error__, __error_details__ |  msg=~"Failed to get keys from [m].*"';
+    '{service_name=~"tempo-distributor|tempo-ingester"} | user!="03428" |~ "(?i)Error" | json | logfmt | drop __error__, __error_details__ |  msg!="Failed to get keys from redis"';
 
   const from = dateTimeParse(initialStart);
   const to = dateTimeParse(initialEnd);


matyax:
I'm seeing a lot of errors in the console for this branch. I think there's a missing "" around some of the filter values or something like that, seeing the JSON parse exceptions.

Tested just by visiting the embed route.

imagen

imagen

src/services/logqlMatchers.ts
```diff
@@ -323,25 +323,33 @@ function parseFields(query: string, context?: PluginExtensionPanelContext, lokiQ
...
+        if (logFmtParser.length && jsonParser.length) {
+          parser = 'mixed';
+        } else if (logFmtParser.length) {
+          parser = 'logfmt';
+        } else if (jsonParser.length) {
+          parser = 'json';
+        } else {
+          // If there is no parser in the query so far, the field would have to be metadata or an invalid query?
+          labelType = labelType ?? LabelType.StructuredMetadata;
+          parser = 'structuredMetadata';
```
matyax, line 351: I understand a "mixed" parser, but a structuredMetadata parser is technically incorrect. How about "none"? Not a big issue, but it's potentially confusing.

matyax (commented):
Other than the comments I made, it works as expected.

src/services/logqlMatchers.ts
gtk-grafana: let's not muddy up this PR, the parser meta prop is only used within the logs drilldown application, and fields with `structuredMetadata` parser are only used in the structured metadata variable, it just happens to be that in logQL structured metadata is not parsed.

TL;DR, this is nuanced, structured metadata might not be a parser in logQL but at the current level of abstraction in we have a dedicated structured metadata variable and we don't really care that something is or isn't parsed, but we really do care that fields and metadata make it to the right variable.

gtk-grafana:
Error should be fixed, forgot to encode the field value. We might not need to merge this any more though, so maybe we hold onto this until we hear more from the k8 embedding folks

matyax:
I think it's fine to support this, and could be requested again at any point. Unless you see a chance for regressions, I can't find reasons to hold this.

src/Components/EmbeddedLogsExploration/EmbeddedLogs.tsx
```diff
@@ -63,10 +63,15 @@ export function buildLogsExplorationFromState({
   }));
 
   const initialFields: AdHocFiltersWithLabelsAndMeta[] | undefined = fields?.map((f) => {
+    const fieldValue: FieldValue = {
+      parser: f.parser ?? 'mixed',
+      value: f.value,
+    };
     return {
       key: f.key,
       operator: f.operator,
```
matyax, line 69: Should we also check for regexes and apply the custom prefix? I suspect we should.

gtk-grafana: Supported in https://github.com/grafana/logs-drilldown/pull/1683/commits/59781032db85b31cbdbad09bf798eab05d93de61

Issue #1682, opened 2025-12-16
[FEAT]: Embedded logs drilldown - break out multiple regex values into multiple pills


gtk-grafana:
**Describe the issue**
As of https://github.com/grafana/logs-drilldown/pull/1678, embedded queries with multiple values under a regex operator works, but combined values are harder to work with in Logs drilldown, let's break them out into multiple pills.

**Expected behavior**
queries of the form `{service_name=~"tempo-distributor|tempo-ingester"}` are broken out into 2 pills in the variable, one for each value.

PR #1684, opened 2025-12-16
fix(embedded): split combined pills into multiple

gtk-grafana:
Fixes: https://github.com/grafana/logs-drilldown/issues/1682


matyax (commented):
The problem with regexes is that they can be any arbitratry expression, not just a|b

`const query = '{place=~"luna|(m|n)oon"}';`

imagen

`const query = '{place=~"luna|[a-zA-Z]+oon"}';`

imagen

gtk-grafana:
Yeah good point, I was hoping we could improve the current UX without needing to support a whole new API that takes filters instead of a query string, but that's probably what we'll need to do

PR #1685, opened 2025-12-17
feat: Default columns table buttons

gtk-grafana:
Split out into a new PR since this will conflict with https://github.com/grafana/logs-drilldown/pull/1665


gtk-grafana:
Not worth fussing with in the old table until we either remove urlColumns or migrate to ng-table.

PR #1694, opened 2026-01-07
fix(Table): move panel wrap reference and improve size calculations

matyax:
This PR directly addresses the "inch-worm" bug that was reproducible with the docker mega menu, and also fixes width and height calculation for the table.

The fix was by changing the ref from the global wrapper to a div inside the PanelChrome, removing the need for magic numbers in the calculation. A minor overflow would cause a resize loop. Additionally, it fixed some overflows due to the referenced div and these calculations.

https://github.com/user-attachments/assets/e29f21f7-cfe5-4242-a611-4e2f4f8f571f

### DIFF
--- a/src/Components/Table/TableWrap.tsx
+++ b/src/Components/Table/TableWrap.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useState, useTransition } from 'react';
 
 import { css } from '@emotion/css';
 import { useResizeObserver } from '@react-aria/utils';
@@ -49,6 +49,7 @@ export const TableWrap = (props: TableWrapProps) => {
   const { logsFrame } = useQueryContext();
 
   const [panelWrapSize, setPanelWrapSize] = useState({ height: 0, width: 0 });
+  const [, startTransition] = useTransition();
 
   // Table needs to be positioned absolutely, passing in reference wrapping panelChrome from parent
   useResizeObserver({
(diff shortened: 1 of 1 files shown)


src/Components/Table/TableWrap.tsx
```diff
@@ -51,18 +49,33 @@ export const TableWrap = (props: TableWrapProps) => {
   const { logsFrame } = useQueryContext();
 
   const [panelWrapSize, setPanelWrapSize] = useState({ height: 0, width: 0 });
+  const [, startTransition] = useTransition();
```
gtk-grafana, line 52:
```suggestion
  const [isPending, startTransition] = useTransition();
```

src/Components/Table/TableWrap.tsx
```diff
@@ -51,18 +49,33 @@ export const TableWrap = (props: TableWrapProps) => {
...
 
   // Table needs to be positioned absolutely, passing in reference wrapping panelChrome from parent
   useResizeObserver({
     onResize: () => {
       const element = props.panelWrap.current;
-      if (element) {
-        if (panelWrapSize.width !== element.clientWidth || panelWrapSize.height !== element.clientHeight) {
+      if (!element) {
```
gtk-grafana, line 58:
```suggestion
      if (!element || isPending) {
```

gtk-grafana (approved):
LGTM, setting the width to 0 when inch-worming is a clever workaround

src/Components/Table/TableWrap.tsx
gtk-grafana: This can help reduce a few extra setstates

matyax:
Yeah, that definitely seems to be fixing it @L2D2Grafana . Still, we really need to change the reference element to set the table size because it seems to be causing this extra scroll that someone complained about a few weeks ago:

imagen

I'll see if I can do that on top of your changes.

matyax:
Let's see if https://github.com/grafana/logs-drilldown/pull/1695 can supersede this PR.

gtk-grafana:
the transition hook for calculating widths on resize might be a good performance enhancement in addition to the fix in https://github.com/grafana/logs-drilldown/pull/1695

matyax:
Yeah, I tried it and didn't like the results at all, specially with a throttled CPU.

matyax:
Closing this one to continue in the other one.

Issue #1631, opened 2025-11-12
Logs Breakdown: Check container sizes and unnecessary scrolls


matyax:
With enough screen height, and collapsed filters/logs volume, the logs view can fit without scrolls. Check heights and styles, and also overflow settings. See image below.

<img width="1916" height="846" alt="Image" src="https://github.com/user-attachments/assets/afc84f91-de88-4404-a869-f62450687275" />

PR #1695, opened 2026-01-07
fix(table): Improve table size container and fix resize behavior with docked Mega Menu

L2D2Grafana:
This PR directly addresses the "inch-worm" bug that was reproducible with the docked mega menu, and also fixes width and height calculation for the table.

The fix combines a CSS update and a change of the passed width/height ref from the global wrapper to a div inside the PanelChrome, removing the need for magic numbers in the calculation. A minor overflow would cause a resize loop. Additionally, it fixed some overflows due to the referenced div and these calculations. This change additionally fixes https://github.com/grafana/logs-drilldown/issues/1631

https://github.com/user-attachments/assets/e29f21f7-cfe5-4242-a611-4e2f4f8f571f

### DIFF
--- a/src/Components/Table/TableWrap.tsx
+++ b/src/Components/Table/TableWrap.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState, useTransition } from 'react';
+import React, { useCallback, useState } from 'react';
 
 import { useResizeObserver } from '@react-aria/utils';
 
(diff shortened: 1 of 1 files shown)


matyax:
@gtk-grafana I added the transition in the last commit. Wanna give it a try? I'm not skipping isPending because it has the risk to skip an actual valid value.

gtk-grafana:
hmm, that last commit seems to be hurting more then helping, seeing flickering for >5s after resize compared to <2s

PR #1697, opened 2026-01-08
chore(deps): add lint-staged as dep

L2D2Grafana:
Somehow lint-staged was never added as a dev dependency 😮‍💨 . Adding it now.


gtk-grafana (approved):
Nice catch. LGTM

PR #1698, opened 2026-01-08
chore(default-columns): upgrade API from alpha to beta

gtk-grafana:
Upgrading alpha API to beta

Core: https://github.com/grafana/grafana/pull/116035


matyax:
Can't get past 

imagen

And this is what I would like to prevent when we throw Errors from React.

matyax:
With the docker image (yarn server):

imagen

L2D2Grafana (approved):
Playwright.config.ts lgtm, I think matching anything above a range is good so we don't forget the EXACT version number again. All the tests are running against 12.3.1

Issue #216, opened 2024-04-10
Feature: Bookmark / Save feature similar to explore metrics


shantanualsi:
It would be useful to explore the possibility of having bookmarks or saved "trails" similar to what's in explore metrics

PR #1702, opened 2026-01-09
feat(SaveSearch): Add support to save filters (or queries) to local storage or to Saved Queries if available

matyax:
People without Saved Queries support (OSS users, older Grafana versions, etc) will use local storage for saving queries, while Cloud/Enterprise users with Saved Queries access will use it for storage.

When Saved Queries is available, we not only show Loki queries to use, but we also save new queries to it.

Closes https://github.com/grafana/logs-drilldown/issues/216

### How to test

- Requires [Grafana Enterprise](https://github.com/grafana/grafana-enterprise/).
- Requires [Grafana Core branch](https://github.com/grafana/grafana/pull/116622)

### Analytics

Save:

- grafana_logs_app_save_search_visited
- grafana_logs_app_save_search_search_saved

Load:

- grafana_logs_app_load_search_visited
- grafana_logs_app_load_search_search_toggled
- grafana_logs_app_load_search_search_deleted
- grafana_logs_app_load_search_search_loaded
- grafana_logs_app_load_search_saved_query_loaded

### Local storage demo:

https://github.com/user-attachments/assets/0207643b-5323-40f3-93b3-5fca56f5922b

### Saved queries demo:

https://github.com/user-attachments/assets/17867189-36f5-4e79-aa4c-71d787f60ae9

### Load query UI, inspired in Saved Queries:

imagen

### Save Search UI:

imagen

Pending:
- [x] POC validation and consensus
- [x] Disable in embedded apps
- [x] Tests
- [x] Analytics


src/services/saveSearch.ts
```diff
@@ -0,0 +1,215 @@
...
+    },
+    [editQueryTemplate]
+  );
+
+  const deleteSearch = useCallback(
+    async (uid: string) => {
+      if (backend === undefined) {
+        logger.error('[Save search]: Uninitialized');
+        return;
+      } else if (backend === 'local') {
```
L2D2Grafana: It's nice to use localStorage and imho we overused it because we didn't have a BE. If we're using the App Platform apis that are open source. I'd be fine dropping localStorage and only rely on the api.

matyax:
> Is there a feature flag I need to run to see this in core and logs-drilldown to see the api calls?

I forgot. to add to the description that this requires grafana-enterprise.

src/services/saveSearch.ts
matyax: The APIs are not always available, and in those cases we use local storage.

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -214,13 +215,16 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
...
-        new ToolbarScene({
-          isOpen: false,
-          key: CONTROLS_VARS_TOOLBAR,
-        })
-      );
+    if (getDrilldownSlug() === 'explore') {
+      controls.push(new LoadSearchScene());
+      if (config.featureToggles.exploreLogsAggregatedMetrics) {
+        controls.push(
+          new ToolbarScene({
```
gtk-grafana, line 222: I think we should put the button on the left side of the datasource, I never liked where I put the agg metrics button

gtk-grafana: Also if we don't have any saved searches I'm not sure we want to show the button?

xenartro: It appears as disabled when there are no searches, with a tooltip.

src/Components/SavedSearches/LoadSearchModal.tsx
```diff
@@ -0,0 +1,269 @@
...
+  selected?: boolean;
+}
+
+function SavedSearchItem({ onSelect, search, selected }: SavedSearchItemProps) {
+  const styles = useStyles2(getStyles);
+
+  const id = useId();
+  return (
+    // eslint-disable-next-line jsx-a11y/label-has-associated-control
+    <label className={styles.label} htmlFor={id}>
```
gtk-grafana: Add an aria-label instead of the eslint-disable

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: And we should find a single placement for this button instead of having a different location for serviceScene and serviceSelection

src/Components/SavedSearches/LoadSearchModal.tsx
```diff
@@ -0,0 +1,269 @@
...
+      backgroundColor: theme.colors.action.selected,
+    },
+
+    ':has(:focus-visible)': css({
+      backgroundColor: theme.colors.action.hover,
+      outline: `2px solid ${theme.colors.primary.main}`,
+      outlineOffset: '-2px',
+    }),
+
+    '.favoriteButton': {
```
gtk-grafana: Is this used?

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: @matyax @xenartro 

Actually, let's just nuke it from the serviceSelection scene and keep the current placement with the share buttons (although maybe actually put the saved searches on the left side of share)

> It appears as disabled when there are no searches, with a tooltip.

Yeah, but if users can't interact with it why is it taking up space?

Image

gtk-grafana: in this context it is more permissible: 

Image

src/lib/api-clients/queries/v1beta1/endpoints.gen.ts
```diff
@@ -0,0 +1,647 @@
+/* eslint-disable jsdoc/check-alignment */
+/* eslint-disable sort/string-unions */
+/* eslint-disable sort/type-properties */
+import { api } from './baseAPI';
```
gtk-grafana: we should check back in on app platform to see how they are progressing on externalizing rtkq, also maybe a comment in here stating that this is temporary

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: ahhh, but when you do have saved queries you kinda do want it visible on the service selection UI... I see

gtk-grafana: Actually it might be better placed with the label filters (in both views) as it has to do with filtering?

gtk-grafana:
To spin up local (docker) grafana running enterprise see: https://github.com/grafana/logs-drilldown/pull/1708

gtk-grafana (commented):
Overall very cool and useful, I love it and cannot wait.

But here come all the nits anyway: 

disable "save search" button when query is already added to library (or throw error on adding duplicate query)

The "Query template successfully edited" confuses me, and shows up behind the modal backdrop:
Image

We query the `/queries.grafana.app/v1beta1/namespaces/default/queries?limit=1000` on page load, and every time the button is hit. Let's only call once per session max (unless user saves a new query), if we can set browser cache headers even better. You might need the metadata service singleton to cache api results across scene routes. I'll look tomorrow if we can cache app platform GETs via browser without core changes.

src/Components/SavedSearches/LoadSearchModal.tsx
matyax: Good call, done.

matyax:
Thanks for the feedback @gtk-grafana 

> disable "save search" button when query is already added to library (or throw error on adding duplicate query)

The current Query Library doesn't have this limitation. We can show a notice if the query is already stored, but ultimately let the user decide.

imagen

> The "Query template successfully edited" confuses me, and shows up behind the modal backdrop:

Yes, unfortunately comes from the middleware definition in Grafana Enterprise, which we can't change right now.

> We query the /queries.grafana.app/v1beta1/namespaces/default/queries?limit=1000 on page load, and every time the button is hit.

For now, I don't think we need to get out of our way more than delegating that to the library. We are now keeping the results for 5 minutes. The Explore Saved Queries component is currently refetching on every mount.

> rename /src/lib/api-clients/v1beta1/ to /src/lib/api-clients/queries/v1beta1/

Done.

gtk-grafana:
@matyax 

> Yes, unfortunately comes from the middleware definition in Grafana Enterprise, which we can't change right now.

Can we create an issue for this in the proper repo?

> For now, I don't think we need to get out of our way more than delegating that to the library. We are now keeping the results for 5 minutes. The Explore Saved Queries component is currently refetching on every mount.

~IMO we should be deciding the cache rules in the application and not in the API, but it's fine if we want to address this later. Except caching the initial call across routes, that I'd like to see on the initial release especially if we always make another call when the modal is opened. I can show you how I did that with default columns or I can update in a child PR if you'd rather.~

Edit: NVM the app platform api seems to cache stuff in redux outside of scenes, so setting `refetchOnMountOrArgChange` appears to work and cache results across routes. Nice find!

src/Components/IndexScene/IndexScene.tsx
matyax: Mind making a quick mockup? I'm not sure I follow the proposed placement.

gtk-grafana: image

or

image

image

image

gtk-grafana: Or add a new row on the service selection scene and always leave them on the right (where they are on the service scene) might be a better use of space, but conceptually having these buttons preceding the filters makes sense as they overwrite the labels/filters

gtk-grafana: TL;DR let's not move the icons between selection/service scenes

matyax: Ah, perfect, thank you! I like the service selection alternative much more, let's use one of those.

For the service details, I would like to leave the header alone, not adding more stuff there, and use the Share section, as it's not only less cluttered, but I feel resemblance between the sharing and saving features.

gtk-grafana: Sounds good, next to share makes sense if you are running cloud, although maybe a bit misleading for OSS since they are not sharable.

But good enough for MVP, we need to come back and redesign the whole header anyway

matyax: Moved 

imagen

matyax:
There shouldn't be any pending things to address in this PR up until this point. @CollinFingar whenever you have time, I'd love to know what your team thinks of this approach, and if it looks good to you at least as an MVP of this feature.

matyax:
Please, if you can think of a better place for it in the landing page, please feel free to share.

CollinFingar (commented):
Looking good! I love the integration with Saved Queries!

What are your thoughts on opening Saved Queries (when enabled) for the save instead of via the modal? That way, the user can also add tags, we wouldn't have to update the modal when Saved Queries is further developed with new metadata, and the user would know that query has also been saved to Saved Queries?

If not, I think maybe a blob of text in the modal mentioning that it is saving to Saved Queries would be ideal. That way, users would know the resource they're actually adding to in case things start filling up for Saved Query users elsewhere haha

Side note, [here](https://github.com/grafana/grafana-enterprise/pull/9510/changes) is a change implemented for Query History in Explore. It causes the drawer to close after saving which could be helpful if we want the drawer to open here instead.
Image

gtk-grafana:
@CollinFingar is there a saved query component externalized for plugin consumption that we can use in an OSS repo?

CollinFingar:
@gtk-grafana Check out [this](https://github.com/grafana/grafana/blob/179d5f605e889edae0d052c20b9d234a22ebb68a/public/app/features/explore/QueryLibrary/QueryLibraryContext.tsx) file.

We added this to integrate with Explore, Panel-Edit, etc. in OSS. For example, check out the use of [opening the drawer and saving a query](https://github.com/grafana/grafana/blob/179d5f605e889edae0d052c20b9d234a22ebb68a/public/app/features/explore/QueryLibrary/QueryLibraryContext.tsx#L99) in [Query History](https://github.com/grafana/grafana/blob/179d5f605e889edae0d052c20b9d234a22ebb68a/public/app/features/explore/RichHistory/RichHistoryAddToLibrary.tsx#L44-L53)

gtk-grafana:
@CollinFingar awesome thank you! 
@matyax we just need to make sure that the drawer still works in the embedded contexts (i.e. knowledge graph/asserts) which are already in a drawer and we should be good to use the component from enterprise instead

matyax:
That sounds great, but unless I missed something really obvious, none of those are exposed components for plugins @CollinFingar @gtk-grafana

~~If that is the case, it would be interesting to consider a Saved Queries implementation that is exposed to plugins, supporting saving and retrieving. But instead of being just an externalization of the current components, it should be built with embedding in mind, supporting for example only showing queries of a given data source (for example Loki, for us). Here's an example of that, but for creating alerts: https://github.com/grafana/grafana/pull/116439~~

~~For now I will update the copy text to reflect what we discussed in Slack, and add Tags to the query creation form. Thanks!~~

matyax:
Allright @gtk-grafana / @CollinFingar with just a tiny exposed component from core I was able to get full access to Saved Queries, if available, from Logs Drilldown. See https://github.com/grafana/grafana/pull/116622

This PR has also been updated to use that instead. This is great!

src/Components/IndexScene/VariableLayoutScene.tsx
```diff
@@ -241,7 +241,7 @@ function getStyles(theme: GrafanaTheme2, height: number, headerCollapsed = false
       flexWrap: 'wrap',
       gap: theme.spacing(2),
       label: 'filtersWrap',
-      width: 'calc(100% - 450)',
+      width: 'calc(100% - 450px)',
```
matyax: Something I just noticed.

imagen

gtk-grafana: Nah delete this line instead

Without this line:
image

With this line:
image

src/Components/IndexScene/IndexScene.tsx
gtk-grafana: @matyax it's pretty confusing that the used saved queries button moves, can we duplicate the load query button, or always put one or both buttons next to the filters?

image

image

gtk-grafana: Or heck, make the tabs area narrower on index and put them in the same place as they are on the service selection, just without a share button

src/services/saveSearch.ts
```diff
@@ -0,0 +1,107 @@
...
+import { v4 as uuidv4 } from 'uuid';
+
+import { config } from '@grafana/runtime';
+import { DataQuery } from '@grafana/schema';
+
+import pluginJson from '../plugin.json';
+import { logger } from './logger';
+import { narrowSavedSearches } from './narrowing';
+
+const MIN_VERSION = '12.4.0-pre';
```
gtk-grafana: This doesn't work for cloud versions, when running `yarn server:enterprise` I'm getting a version that does support the change `v12.4.0-21342258703`, but the `isQueryLibrarySupported` returns false.

gtk-grafana:
```suggestion
const MIN_VERSION = '12.4.0-21256324731';
```

gtk-grafana: https://github.com/grafana/grafana/pull/116622#issuecomment-3785354545.

Looks like it was released in `12.4.0-21256324731`

matyax: Is the feature flag enabled? We can also ignore the versions and just use the feature flag status, like Saved Queries internally does.

src/Components/IndexScene/IndexScene.tsx
matyax: Yeah, definitely the latter.

src/Components/IndexScene/VariableLayoutScene.tsx
matyax: Yikes, that's not what I saw.

src/services/saveSearch.ts
gtk-grafana: > Is the feature flag enabled? 

Yup

> We can also ignore the versions and just use the feature flag status, like Saved Queries internally does.

But then folks that use saved queries in explore, but don't have the new grafana changes will probably get an error thrown.

But looks like this was addressed in https://github.com/grafana/logs-drilldown/pull/1702/commits/916d5588875a4d8d44172613d625da608cd9f390 so we should be good on this

src/Components/IndexScene/IndexScene.tsx
matyax: imagen

👌

src/Components/IndexScene/VariableLayoutScene.tsx
matyax: Removed.

matyax:
Addressed the button placement. Looking much better now!

Issue #1706, opened 2026-01-12
[BUG]: Copy short-links not working in ios/safari


gtk-grafana:
**Describe the bug**
Copy short link button returns success message, but fails to copy url.

PR #1707, opened 2026-01-12
fix(links): upgrade to clipboard.write for ios/safari

gtk-grafana:
Use clipboard.write which has better ios/safari support.
Also memoize short-links so we don't make unneeded calls.

Fixes: https://github.com/grafana/logs-drilldown/issues/1706

Note: Another use-case for better externalization of methods from core grafana, this bug was fixed in core Grafana but since the short link methods are not exposed, we have to manually copy changes into logs drilldown repo.

Also there is a new short-link API that we are not using because the rtkq handlers are not exposed for plugins.

### DIFF
--- a/src/services/text.ts
+++ b/src/services/text.ts
@@ -4,7 +4,7 @@ import { config, locationService } from '@grafana/runtime';
 import { logger } from './logger';
 
 /**
- * Copies text synchronously. If not executed in the same callstack as the user event this will fail in Safari!
+ * Copies text synchronously. If not executed in the same call stack as the user event this will fail in Safari!
  * @param string
  */
 export const copyText = (string: string) => {


src/services/links.ts
```diff
@@ -0,0 +1,82 @@
...
+// Copied from shortLinks.ts in core
+export const createAndCopyShortLink = async (path: string) => {
+  const appEvents = getAppEvents();
+
+  try {
+    if (typeof ClipboardItem !== 'undefined' && navigator.clipboard.write) {
+      await navigator.clipboard.write([createShortLinkClipboardItem(path)]);
+    } else {
+      const shortLink = await createShortLink(path);
+      copyText(shortLink);
```
xenartro, line 54: Shoudln't we fix copyText instead? If we inadvertedly use copyText without this fix, it would have the same problem.

gtk-grafana: copyText will work as long as it is invoked in the same callstack as a user interaction, it's only the async case that needs clipboard.write

xenartro: Ohhh, unexpected.

gtk-grafana: I followed the approach in core, but documented this behavior in copyText

L2D2Grafana (approved):
LGTM seems like core has updated createAndCopyShortLink() https://github.com/grafana/grafana/blob/main/public/app/core/utils/shortLinks.ts#L94
And in safari the link is copied now.

PR #1709, opened 2026-01-12
chore: promote aggregated metrics to GA

gtk-grafana:
Core PR: https://github.com/grafana/grafana/pull/116174

For the record I do not think this is a good idea, aggregated metrics requiring the query to be rewritten breaks how scenes expects queries to behave.

Also queries less then 1 minute on the service selection will always be empty despite having results:

image


gtk-grafana:
Closing: we will look into converting the feature flag to the new feature flag format because Loki doesn't have time to address the awfulness of the aggregated metrics API by the time we need to clean up the feature flags

Issue #1704, opened 2026-01-09
Remove Feature Toggles that are greater than 1 year old


cakennedy:
In an effort to reduce costs by reducing cardinality in our metrics and other instrumentation, and to facilitate migration off of the old feature flag system,[Dave Henderson](mailto:dave.henderson@grafana.com) put together [a list of all the feature toggles that are older than one year, for each squad](https://docs.google.com/spreadsheets/d/1Ww2ibkZ9ZQOx58QTQiLue5SqjH8jnHfhIGdgE1w7Vng/edit?gid=0#gid=0&fvid=856216879).  He’s asked that we take action on these feature flags:

- If the features are not going to progress beyond experiments, let's consider deleting them
- If they are going to progress, let's prioritize that
- If they should really just be settings (not feature flags), let's fix that
- 

For Logs Drilldown, there are several:  

- [x] [lokiExperimentalStreaming](https://github.com/grafana/grafana/pull/116205)
- [x] [lokiLogsDataplane](https://github.com/grafana/grafana/pull/116205)
- [x] logsExploreTableDefaultVisualization,
- [ ] exploreLogsShardSplitting (keep and migrate to the new feature flag system)
- [x] [exploreLogsLimitedTimeRange](https://github.com/grafana/grafana/pull/116177)
- [x] exploreLogsAggregatedMetrics, (Loki squad won't have time to promote aggregated metrics API to GA by deadline so we need to convert this to the new feature flag format. Trevor won't let Galen delete it.)
- [ ] lokiShardSplitting (keep and migrate to the new feature flag system)
- [ ] logsContext
- [ ] dataplane (oss-big-tent) meet with them turn on everywhere, etc data frame contact spec, time and line to timestamp and body. Only affects logs not metrics.
- [ ] lokiQueryLimitsContext (less then 1 year old, but needs to be long lived toggle, reassign to big tent?)

The task is to make a decision about them that allows us to remove the feature toggle, then implement the decision.  So, if the decision is that the functionality controlled by this feature toggle is no longer needed, the implementation will be to remove the toggle and the functionality.  Alternatively, the decision could be to promote the functionality to be unconditionally available and remove the toggle.  Or, as a third alternative, the toggle needs to be turned into a setting.  

Once the decision has been made and implemented, please also go back to the spreadsheet and update the status for the feature toggle.

# Docs and Example in Metric

[...]

PR #1711, opened 2026-01-13
chore(feature flags): add openfeature flags

L2D2Grafana:
- #1704 
- [Metrics Drilldown example](https://github.com/grafana/metrics-drilldown/pull/886/)
- [Docs](https://enghub.grafana-ops.net/docs/default/Component/grafana-backend-services-docs/openfeature/feature-flags-with-open-feature)

Adding the foundations for the new GOFF feature flag implementations and converted exploreLogsAggregatedMetrics to use the new stuff. A cache of Feature flags was added onload to use throughout the app just like the previous config.featureToggles. Users would need to reload the app if a feature flag changed to get the new value.

@NWRichmond what's the idea behind adding the feature flag to our tracking? Also I did like the simplified definitions of enabled, and disabled or boolean types over the [multi-variant type](https://gofeatureflag.org/docs/configure_flag/create-flags#-multi-variant-flags) where we look for treatment, what's the reasoning behind that?
```
display-banner:
  variations:
    enabled: true
    disabled: false
  defaultRule:
    variation: disabled
```
https://gofeatureflag.org/docs/configure_flag/create-flags#%EF%B8%8F-create-your-first-feature-flag-manually


NWRichmond:
> what's the idea behind adding the feature flag to our tracking?

This is just for feature flags that are used in A/B tests. I suppose there could be some other use cases where the feature flag value might be important information to accompany the rest of the payload, but I can't think of an example.

> Also I did like the simplified definitions of enabled, and disabled or boolean types over the [multi-variant type](https://gofeatureflag.org/docs/configure_flag/create-flags#-multi-variant-flags) where we look for treatment, what's the reasoning behind that?

For many feature flags, a boolean or `enabled`/`disabled` is great. There are cases where it helps to have more variants. One of those cases is A/B testing. Specifically, for A/B tests where it's useful to exclude certain users from the test altogether. Users who are excluded typically get the same experience as the `control` cell. But to have a nice 50-50 split of `treatment` and `control` populations in the experiment data, we need to draw a distinction between people who got the `control` treatment because they're part of the 50% of the experiment that got the control, and people who got the `control` treatment because they were excluded from the experiment. In this case, having multi-variant feature flags is super useful, and we can add a third variant, `excluded`, to keep our `treatment`/`control` proportions true to the [percentages](https://gofeatureflag.org/docs/configure_flag/rollout-strategies/experimentation#example) defined in an experiment-style flag definition.

There are some more sophisticated setups using JSON objects that help with some advanced use cases. [Here's an example](https://github.com/grafana/deployment_tools/blob/dbe2678f600c8d85c89b25df946854ba563a78d7/ksonnet/environments/hosted-grafana/waves/feature-toggles/goff/pathfinder/flags/experiment-variant.libsonnet#L20-L27) from the Pathfinder app.

src/Components/App.tsx
```diff
@@ -1,24 +1,39 @@
...
+// Initialize OpenFeature provider and populate flag cache
+const featureFlagsReady = initOpenFeatureProvider().then(() => initializeFeatureFlags());
+
+const App = (props: AppRootProps) => {
+  const [isReady, setIsReady] = useState(false);
+
+  useEffect(() => {
+    // Initialize and cache the feature flags for use in the app
+    featureFlagsReady.then(() => {
+      setIsReady(true);
```
gtk-grafana: If the call fails the entire app fails to load stuck in a loading state, can we catch any network errors and fall back?

gtk-grafana (changes requested):
Breaks the whole app in 11.6 - 12.1 as the API is not available, let's add some backward compatibility to attempt to load the flags from config if the API call fails

L2D2Grafana:
> Breaks the whole app in 11.6 - 12.1 as the API is not available, let's add some backward compatibility to attempt to load the flags from config if the API call fails

Nice catch! Updated waitForClientReady() for errors and to fallback to config.featureToggles.exploreLogsAggregatedMetrics; Tested with version 11.6.0

src/Components/App.tsx
matyax: This is a valid concern. This is missing a `.catch()`. We should account for it an minimally show an error message, and an invitation to reload/try again.

L2D2Grafana: Added a catch()

PR #1715, opened 2026-01-15
chore(ci): check grafanaVersion in the tests, turn on running dev version

L2D2Grafana:
https://github.com/grafana/plugin-ci-workflows/pull/206#issuecomment-3755249309

- Update our e2e workflow to check for the grafana version in the test
- Turn on running the dev image

Well it's cool right now that the dev image passes I suspect that won't always be the case and we might want to check for that in the isLatestGrafana()
Screenshot 2026-01-15 at 1 39 29 PM
Running most of the test on later versions
Screenshot 2026-01-15 at 1 37 07 PM
Skipping older versions
Screenshot 2026-01-15 at 1 36 39 PM

### DIFF
--- a/tests/config/grafana-versions-supported.ts
+++ b/tests/config/grafana-versions-supported.ts
@@ -1,4 +1,6 @@
+import { test } from '@grafana/plugin-e2e';
 import semver from 'semver';
+
 export const GRAFANA_LATEST_SUPPORTED_VERSION = '12.3.1';
 /**
  * Check if the current Grafana version is the latest supported version or newer.
@@ -25,7 +27,5 @@ export const isLatestGrafana = (grafanaVersion: string): boolean => {
  * });
  */
 export const skipUnlessLatestGrafana = ({ grafanaVersion }: { grafanaVersion: string }): void => {
-  if (!isLatestGrafana(grafanaVersion)) {
-    test.skip(`Skipping: requires Grafana >= ${GRAFANA_LATEST_SUPPORTED_VERSION}`);
-  }
+  test.skip(!isLatestGrafana(grafanaVersion), `Skipping: requires Grafana >= ${GRAFANA_LATEST_SUPPORTED_VERSION}`);
 };


tests/appNavigation.spec.ts
```diff
@@ -1,12 +1,14 @@
...
 import pluginJson from '../src/plugin.json';
+import { GRAFANA_LATEST_SUPPORTED_VERSION, isLatestGrafana } from './config/grafana-versions-supported';
 import { ExplorePage } from './fixtures/explore';
 
 test.describe('navigating app', () => {
   let explorePage: ExplorePage;
 
-  test.beforeEach(async ({ page }, testInfo) => {
+  test.beforeEach(async ({ page, grafanaVersion }, testInfo) => {
+    test.skip(!isLatestGrafana(grafanaVersion), `Skipping: requires Grafana >= ${GRAFANA_LATEST_SUPPORTED_VERSION}`);
```
sunker: I don't really understand the `GRAFANA_LATEST_SUPPORTED_VERSION` variable. Do you update this variable after each Grafana release? To me it would be more explicit to define exactly from what Grafana version the features that this section tests are available. E.g 
`test.skip(semver.lt(grafanaVersion, '12.3.1'), 'Feature x not supported before 12.2.2`)`

sunker (commented):
Looks good from my point of view, just a minor comment. Thanks @L2D2Grafana!

tests/appNavigation.spec.ts
L2D2Grafana: Basically the idea is to run our test suite on the highest grafana version we can, and we run smoke tests on all the other versions. Keeping GRAFANA_LATEST_SUPPORTED_VERSION made sense to me because then I only have to update on place instead of every file.  If we started writing test to match each grafana version I could see us doing what you are suggesting in the future.

gtk-grafana: Yeah basically since the selectors and test ids change regularly between grafana versions, this pins our e2e tests to run against something that we know will work until we have time to make updates to the selectors for the latest release.

sunker: Please note that the purpose of the `grafana/plugin-e2e` package is to provide cross-grafana version testing support, so you should not have to adapt plugin tests depending on Grafana version. All e2e-selectors in Grafana are [versioned](https://github.com/grafana/grafana/tree/main/packages/grafana-e2e-selectors/src/selectors#versioned-selectors), meaning their values may expand to a different value depending on the Grafana version. So if you use the e2e-selectors resolved by the `selectors` fixture in Playwright, plugin-e2e will ensure they are compatible with the version of Grafana that is currently being targeted in your test. 

e.g
```ts
test('annotation query should be OK', async ({ annotationEditPage, selectors }) => {
  // the following e2e selector resolves to different values for different version of Grafana,
  // but the e2e test code remains the same. 
  await annotationEditPage.getByGrafanaSelector(selectors.components.CodeEditor.container).fill('SELECT * FROM users');
  await expect(annotationEditPage.runQuery()).toBeOK();
});
```

You can read more about selecting element the plugin-e2e [docs](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/selecting-elements).

gtk-grafana:
@L2D2Grafana is this PR still the desired approach or were you working on something else?

L2D2Grafana:
> @L2D2Grafana is this PR still the desired approach or were you working on something else?

This is the desired approach from Plugins Platform.

tests/exploreServicesBreakDown.spec.ts
```diff
@@ -37,8 +39,8 @@ test.describe('explore services breakdown page', () => {
   });
 
   test.afterEach(async ({ page }) => {
-    await explorePage.unroute();
-    explorePage.echoConsoleLogsOnRetry();
+    await explorePage?.unroute();
```
gtk-grafana: This is confusing that the type of `explorePage` doesn't allow for undefined but we're still optionally running these methods.

Does the afterEach get run if the test is skipped? If so we should update the type of explorePage or gate afterEach methods to return early if the test was skipped.

Updating the explorePage type might introduce a bunch of changes, so I would guess it's better to avoid running stuff.

tests/appNavigation.spec.ts
gtk-grafana: Sounds good and thanks for the info, last time we looked at the e2e package it had some incorrect/missing selectors and was in alpha so we figured we'd wait as at the time we only had a few tests that needed selectors from core.

Looks like it's still in [alpha](https://github.com/grafana/grafana/tree/main/packages/grafana-e2e-selectors/), but we can create an issue to take another look to see if it will work now.

sunker: Oh, that's not intentional - it's been used in the grafana packages for many years. Let me know if I can help to onboard you with the plugin-e2e package. :)

L2D2Grafana:
> tests/exploreServicesBreakDown.spec.ts

This makes sense because we did not write the tests to support all the versions. We run the full suite on the latest grafana version. Then we run the smoke tests on all the other versions. Going forward we can write tests for each version if we want. Locally the tests pass for me because we're running main.

tests/exploreServicesBreakDown.spec.ts
L2D2Grafana: Yes this code does get run and I cannot bypass it at a higher level in a beforeAll(). I've updated the tests to exit early.

tests/exploreServices.spec.ts
```diff
@@ -16,7 +17,8 @@ test.describe('explore services page', () => {
   let explorePage: ExplorePage;
 
   test.describe('parallel', () => {
-    test.beforeEach(async ({ page }, testInfo) => {
+    test.beforeEach(async ({ page, grafanaVersion }, testInfo) => {
+      test.skip(!isLatestGrafana(grafanaVersion), `Skipping: requires Grafana >= ${GRAFANA_LATEST_SUPPORTED_VERSION}`);
       explorePage = new ExplorePage(page, testInfo);
```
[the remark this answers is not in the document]
matyax: Here.

tests/savedSearches.spec.ts
```diff
@@ -1,19 +1,22 @@
...
 import { testIds } from '../src/services/testIds';
+import { GRAFANA_LATEST_SUPPORTED_VERSION, isLatestGrafana } from './config/grafana-versions-supported';
 import { E2EComboboxStrings, ExplorePage, serviceSelectionPaginationTextMatch } from './fixtures/explore';
 
 test.describe('saved searches', () => {
   let explorePage: ExplorePage;
 
-  test.beforeEach(async ({ page }, testInfo) => {
+  test.beforeEach(async ({ page, grafanaVersion }, testInfo) => {
+    test.skip(!isLatestGrafana(grafanaVersion));
```
matyax:
```suggestion
    test.skip(!isLatestGrafana(grafanaVersion), `Skipping: requires Grafana >= ${GRAFANA_LATEST_SUPPORTED_VERSION}`);
```

L2D2Grafana:
Refactored out the check to a function and added a check in the serialize tests.

PR #1716, opened 2026-01-16
chore(cp): update create-plugin

L2D2Grafana:
### Manually update create-plugin to latest 6.7.5.
The cp-update.yml doesn't seem to be triggering a PR at the beginning of ever month now. We wonder if it was because it breaks locally. I manually ran `npx @grafana/create-plugin@latest update` and fixed the eslint conversion to .mjs then fixed the other more minor errors. Thanks @shelldandy for chatting about it!

### CVE
Added CVE fixes react-router-dom and @grafana/plugin-types-bundler
https://github.com/grafana/logs-drilldown/security/dependabot/63
https://github.com/grafana/logs-drilldown/security/dependabot/60

### DIFF
--- a/docs/sources/shared/index.md
+++ b/docs/sources/shared/index.md
@@ -1,0 +1,6 @@
+---
+headless: true
+description: |
+  A headless page sets the render and list build options to never, creating a bundle of page resources.
+  These page resources are accessible to the `docs/shared` shortcode.
+---
--- a/docs/sources/shared/troubleshoot-logs-drilldown.md
+++ b/docs/sources/shared/troubleshoot-logs-drilldown.md
@@ -1,0 +1,309 @@
+---
+headless: true
+description: |
+  A headless page sets the render and list build options to never, creating a bundle of page resources.
+  These page resources are accessible to the `docs/shared` shortcode.
+---
+
+[//]: # 'This file documents logs drilldown error messages and troubleshooting'
+[//]: #
+[//]: # 'This shared file is included in these locations:'
+[//]: # '/logs-drilldown/docs/logs-drilldown/latest/troubleshooting/_index.md'
+[//]: # 'If you make changes to this file, verify that the meaning and content are not changed in any place where the file is included.'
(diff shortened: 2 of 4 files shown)


src/Components/IndexScene/LevelsVariableScene.tsx
```diff
@@ -130,6 +130,7 @@ export class LevelsVariableScene extends SceneObjectBase<LevelsVariableSceneStat
   };
 
   static Component = ({ model }: SceneComponentProps<LevelsVariableScene>) => {
+    /* eslint-disable react-hooks/rules-of-hooks -- This is a functional component, not a class component */
```
gtk-grafana: uhhhh is there a way we can disable this globally instead of marking up everything? We've never used class components in this repo and we never will

src/Components/IndexScene/LayoutScene.tsx
```diff
@@ -48,16 +48,16 @@ export class LayoutScene extends SceneObjectBase<LayoutSceneState> {
   }
 
   static Component = ({ model }: SceneComponentProps<LayoutScene>) => {
+    /* eslint-disable react-hooks/rules-of-hooks -- This is a functional component, not a class component */
     const indexScene = sceneGraph.getAncestor(model, IndexScene);
     const { contentScene } = indexScene.useState();
     const { interceptDismissed, variableLayout } = model.useState();
+    const styles = useStyles2(getStyles);
```
gtk-grafana, line 54: good catch, is there a new rule that keeps hooks from getting called differently on each render?

src/Components/IndexScene/LevelsVariableScene.tsx
gtk-grafana: Also can we open an issue to get these rules to play nice with scenes, this is a common and supported pattern

L2D2Grafana: I did some digging s🪏 I think that the rule was always there because I can't see much changing in the eslint-plugin-react-hooks, but maybe the rule was not being applied right in previous grafana-eslint that used recommended-latest in the config. Since its not a grafana package, I don't think theres much to do here beside override it.

PR #1720, opened 2026-01-20
docs: Make local build consistent so docs/shared lookups work

jdbaldry:
We sync docs to the legacy explore-logs directory and only to the "next" version so this updates `make docs` to match that.

### DIFF
--- a/docs/sources/troubleshooting/_index.md
+++ b/docs/sources/troubleshooting/_index.md
@@ -15,8 +15,4 @@ weight: 900
 
 # Troubleshooting
 
-[//]: # 'Shared content for error messages and troubleshooting'
-[//]: # 'This content is located in /logs-drilldown/docs/sources/shared/troubleshoot-logs-drilldown.md'
-
 {{< docs/shared lookup="troubleshoot-logs-drilldown.md" source="explore-logs" version="next" >}}
-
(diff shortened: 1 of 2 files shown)


docs/sources/troubleshooting/_index.md
```diff
@@ -15,8 +15,4 @@ weight: 900
 
 # Troubleshooting
 
-[//]: # 'Shared content for error messages and troubleshooting'
-[//]: # 'This content is located in /logs-drilldown/docs/sources/shared/troubleshoot-logs-drilldown.md'
```
JStickler, line 19: I'd rather leave breadcrumbs for the next person who updates these docs.

JStickler (approved):

docs/sources/troubleshooting/_index.md
jdbaldry: I was a bit confused by the comment so I removed it. The `shared` shortcode also has the same information I guess. Happy for you to put it back in :)

JStickler: I think people start with the topic they want to edit, and if it's "empty" it's helpful top know where to look for the content.

jdbaldry: Yeah makes sense, sorry for the extra work caused by me making more than the required change, tired brain last night.

PR #1723, opened 2026-01-29
fix(assistant): add `instructions` for structured metadata for Assistant

svennergr:
Just a nit. We noticed that Assistant will use these structured metadata fields as streamselectors. Adding additional instruction to use label filters.

https://raintank-corp.slack.com/archives/C08GW7KDCJ0/p1769704646612339?thread_ts=1769703344.591589&cid=C08GW7KDCJ0


matyax:
Also using this opportunity to pitch for https://github.com/grafana/grafana-assistant-app/issues/2597 , if you think it's worth to continue investing in the context we provide from apps.

svennergr:
> Also using this opportunity to pitch for [grafana/grafana-assistant-app#2597](https://github.com/grafana/grafana-assistant-app/issues/2597) , if you think it's worth to continue investing in the context we provide from apps.

Good reminder. With [this PR](https://github.com/grafana/grafana-assistant-app/pull/4387) and version `0.1.13` of the `@grafana/assistant` SDK will support `operators` in those label context attributes. Did not yet at structured metadata as it felt a bit too specific, but will think about it again.

Thanks Matias!

PR #1737, opened 2026-02-03
fix: set correct datasource when embedding log drilldown component

vpadi:
## Why                                                                                                                                                                                                                                                                                                                                                                                                                         
When embedding the Logs Drilldown component, there was no way to explicitly specify which datasource to use. The initialization logic would skip straight to checking localStorage or the default datasource, ignoring any datasourceUid passed in the component's state.                                                                                                                                                                                                                                                                                                                                                                                                                                                                             

## What
Added `state?.datasourceUid` to the datasource initialization precedence chain for `IndexScene`.

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -133,6 +133,8 @@ import {
 } from 'services/variables';
 export const showLogsButtonSceneKey = 'showLogsButtonScene';
 
+const DEFAULT_LOGS_DATASOURCE_UID = 'grafanacloud-logs';
+
 interface EmbeddedIndexSceneConstructor {
   datasourceUid?: string;
   hideTimePicker?: boolean;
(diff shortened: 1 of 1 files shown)


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -146,6 +146,7 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     const { jsonData } = plugin.meta as AppPluginMeta<JsonData>;
     const datasourceUid =
       jsonData?.dataSource ??
+      state?.datasourceUid ??
```
gtk-grafana, line 150: We probably want this to precede the jsonData?.dataSource as well

vpadi: Makes sense. I've updated it!

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -133,6 +133,8 @@ import {
 } from 'services/variables';
 export const showLogsButtonSceneKey = 'showLogsButtonScene';
 
+const DEFAULT_LOGS_DATASOURCE_UID = 'grafanacloud-logs';
```
gtk-grafana, line 136: I don't think we want to hardcode a dsUID that doesn't exist

vpadi: You want me to remove it then? I just move the hardcoded string to a constant

gtk-grafana: Oh my bad, I didn't see that this was already in use in main, you can leave it as is.

gtk-grafana (approved):
LGTM, but @vpadi you'll need to sign the [CLA](https://github.com/grafana/logs-drilldown/pull/1737#issuecomment-3841022700) for this repo.

https://cla-assistant.io/grafana/logs-drilldown?pullRequest=1737

PR #1739, opened 2026-02-05
fix(Load Search): respect relative time ranges in the time picker

matyax:
Part of https://github.com/grafana/logs-drilldown/issues/216

Fixes the link generation to respect relative time ranges if they are in use. Note that time ranges are not stored with saved searches; this is about respecting what the user has currently set in their time picker in Logs Drilldown.

### How to test

- With a relative time selected, load a search. The loaded search should use relative time.
- With an absolute time selected, load a search. The loaded search should use absolute time.

### DIFF
--- a/src/Components/SavedSearches/LoadSearchScene.tsx
+++ b/src/Components/SavedSearches/LoadSearchScene.tsx
@@ -69,7 +69,6 @@ export class LoadSearchScene extends SceneObjectBase<LoadSearchSceneState> {
       usePluginComponent<OpenQueryLibraryComponentProps>('grafana/query-library-context/v1');
 
     const indexScene = useMemo(() => sceneGraph.getAncestor(model, IndexScene), [model]);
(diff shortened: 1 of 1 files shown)


L2D2Grafana:
https://github.com/user-attachments/assets/09fad3a9-77f1-4e8a-8722-43b0f6196cdc

Not sure I tested this right. With a relative time selected, load a search. The loaded search should use relative time. But with an absolute time selected, load a search. The loaded search uses a relative time.

matyax:
@L2D2Grafana Time ranges are not stored. What we're respecting is what you have currently selected in the time picker.

matyax:
I think I need to make sure the time range is not stuck with a memoized value that is outdated. I'll take a look at that and fix accordingly. Thanks for the tip @L2D2Grafana !

matyax:
Should be fixed now.

PR #1744, opened 2026-02-10
chore(LoadSearchScene): pass drilldown as context

matyax:
Passing a custom context allows us be better identified and, for example, customize the experience for Drilldown users and don't show the "Open in Drilldown" button when you're already there.

Requires github.com/grafana/grafana-enterprise/pull/10877


L2D2Grafana (approved):
Are there namespacing issues with explore?

matyax:
I thought I wrote a description for this one. @L2D2Grafana Is to be better identified by Saved Queries, and for example don't show the "Drilldown" button when you're already in Drilldown.

PR #1746, opened 2026-02-11
fix(contextToLink): filter out labels generated by label_format

matyax:
Not only any label generated by using label_format would end up as a filter and producing no results, but this was also affecting traces-to-logs, where the "trick" to get logs with trace_id as part of the log line (with a line filter) used label_format, producing no results:

invalid

With these changes, we are excluding from processing as filters all labels coming from labe_format.

valid

Additionally, when we identify that the query contains traces-to-logs fields, we use fields instead of metadata, to allow these to be added after the parsers in the underlying query.

### How to test with traces

- Have a local Loki and Tempo data source configured
- Start local tempo and Loki `make devenv sources=tempo`
- Go to Traces Drilldown
- Open a trace
- Click to see logs
- The generated filters in Logs Drilldown should not contain label_format labels

### DIFF
--- a/src/services/extensions/links.test.ts
+++ b/src/services/extensions/links.test.ts
@@ -112,7 +112,7 @@ describe('contextToLink', () => {
     });
   });
 
-  it('should not include label_format labels as label filters ', () => {
+  it('should not include label_format labels as label filters', () => {
     const target = getTestTarget({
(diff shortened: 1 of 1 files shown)


matyax:
Maybe mine still works because I haven't re-built the image, but it's definitely super unstable. I need to do a hard restart after a few minutes of using it.

PR #1748, opened 2026-02-12
chore(drilldown): run drilldown apps together locally

L2D2Grafana:
🤝 Run All Drilldown Apps Locally
In the spirit of collaboration 🫂, you can run all drilldown apps (except profiles 🥲 ) together locally to test cross-app integration.

🚀 Getting Started
```
yarn
yarn start:all
yarn server:down 
```

> [!NOTE]
> Assumes the plugins are in adjacent directories:
> ../metrics-drilldown/dist
> ../traces-drilldown/dist

@joey-grafana I was having troubles running `make devenv sources=tempo` in grafana. Also I'm not sure how well this is working between traces and logs as I had to change time ranges to get things working. I still can't see logs for traces but it's something.
Screenshot 2026-02-12 at 8 52 16 AM
Screenshot 2026-02-12 at 9 51 43 AM
Screenshot 2026-02-12 at 9 51 55 AM

Fake Data Tour 🎥 

https://github.com/user-attachments/assets/31321984-d311-49f6-aee9-4394eded0736

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,7 @@
     "react-router-dom": "^7.12.0",
     "@remix-run/router": "1.23.2",
     "uuid": "^10.0.0",
+    "undici": "^7.24.0",
     "immutable": "5.1.5",
     "dompurify": "3.3.2",
     "serialize-javascript": "7.0.4",
--- a/src/Components/AppConfig/DefaultColumns/Config.test.tsx
+++ b/src/Components/AppConfig/DefaultColumns/Config.test.tsx
@@ -66,6 +66,12 @@ jest.mock('@grafana/runtime', () => ({
   },
 }));
 
+jest.mock('./isSupported', () => ({
+  isDefaultColumnsSupported: () => true,
+  isDefaultColumnsFlagsSupported: () => true,
+  isDefaultColumnsVersionSupported: true,
+}));
+
 jest.mock('@grafana/api-clients/rtkq/logsdrilldown/v1beta1', () => ({
   ...jest.requireActual('@grafana/api-clients/rtkq/logsdrilldown/v1beta1'),
   useGetLogsDrilldownDefaultColumnsQuery: jest.fn(),
--- a/src/Components/AppConfig/DefaultColumns/Config.tsx
+++ b/src/Components/AppConfig/DefaultColumns/Config.tsx
@@ -20,7 +20,7 @@ const Config = () => {
   if (!dsUID) {
     return <NoLokiSplash />;
   }
-  if (!isDefaultColumnsSupported) {
+  if (!isDefaultColumnsSupported()) {
     return <Unsupported />;
   }
 
--- a/src/Components/AppConfig/DefaultColumns/Unsupported.tsx
+++ b/src/Components/AppConfig/DefaultColumns/Unsupported.tsx
@@ -6,7 +6,7 @@ export function Unsupported() {
   return (
     <section>
       <h2>Default columns</h2>
-      {!isDefaultColumnsFlagsSupported && (
+      {!isDefaultColumnsFlagsSupported() && (
         <p>
           Default columns requires <code>kubernetesLogsDrilldown</code> feature flag to be enabled.
         </p>
--- a/src/Components/AppConfig/DefaultColumns/isSupported.ts
+++ b/src/Components/AppConfig/DefaultColumns/isSupported.ts
@@ -2,7 +2,9 @@ import semver from 'semver/preload';
 
 import { config } from '@grafana/runtime';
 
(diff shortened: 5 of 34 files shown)


joey-grafana:
This is pretty cool! I know that Tempo in core Grafana is out of data i.e. we have the image set to latest but the tempo.yaml config is not up to data so there are config options that are no longer used/possibly new config options that are not present which is causing the Tempo service to fail when run in Docker. 

I sent a message last week to the #tempo squad to take a look :)

generator/log/app-logger.go
```diff
@@ -9,9 +9,10 @@ import (
 )
 
 type AppLogger struct {
-	labels model.LabelSet
-	levels map[model.LabelValue]model.LabelSet
-	logger Logger
+	labels  model.LabelSet
+	levels  map[model.LabelValue]model.LabelSet
+	logger  Logger
+	sleepFn func()
```
trevorwhitney, line 15: what's the motivation of adding the sleepFn? does this change the varied sleep duration services like noisyTempo used to have, and replace it with a static one? if so, is that desired?

docker-compose-defaults.yaml
```diff
@@ -3,6 +3,11 @@ services:
     extends:
       file: .config/docker-compose-base.yaml
       service: grafana
+    ports: !override
+      - 3001:3000/tcp
+    volumes:
```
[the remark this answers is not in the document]
trevorwhitney: I'm curious about this as well, or do we want the children files/definitions to win?

generator/log/values.go
```diff
@@ -1,15 +1,58 @@
...
+
+// IsFullDataMode returns true when GENERATOR_FULL_DATA=1 (CI/E2E uses old/full dataset).
+func IsFullDataMode() bool {
+	return os.Getenv("GENERATOR_FULL_DATA") == "1"
+}
+
+// isFullDataMode returns true when GENERATOR_FULL_DATA=1 (CI uses old/full dataset).
+func isFullDataMode() bool {
+	return IsFullDataMode()
+}
```
trevorwhitney: why do we need the private version here? looks like it just calls the public one?

generator/log/values.go
```diff
@@ -1,15 +1,58 @@
...
+	time.Sleep(time.Duration(500+rand.Intn(1500)) * time.Millisecond)
+}
+
+// LogSleepOriginal sleeps 0–5s. Used when GENERATOR_FULL_DATA=1 (E2E) to match pre-refactor behavior.
+func LogSleepOriginal() {
+	time.Sleep(time.Duration(rand.Intn(5000)) * time.Millisecond)
+}
+
+// IsFullDataMode returns true when GENERATOR_FULL_DATA=1 (CI/E2E uses old/full dataset).
+func IsFullDataMode() bool {
```
trevorwhitney: the purpose of this function is confusing to me? is it true that IsFullData being false is what causes the traces to be emitted? That's confusing to me, as "full" implies "more" to me, and logs and traces would be "more" data. Should this be renamed to better reflect the CI gating it's there for?

docker-compose-defaults.yaml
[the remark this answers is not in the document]
gtk-grafana: without override prop in docker compose it will merge definitions with anything that extends instead of overwriting, so it tries to run it on port 3000 and 3001, so if you're running grafana from source on port 3000 it will throw an error and not spin up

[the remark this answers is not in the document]
gtk-grafana: oh nvm this comment is on volumes, uhh I'll just see myself out

generator/log/app-logger.go
L2D2Grafana: Logs used to be generated and there were calls to sleep mixed in. This abstracted the sleeps into a reusable function. Basically I created a version that sleeps more and takes up less of my docker resources if you run all the apps together with `yarn start:all`

docker-compose-defaults.yaml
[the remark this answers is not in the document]
L2D2Grafana: Removed this from defaults, it's all inherited from docker-compose-base.yaml and only docker-compose-local-all.yaml overrides it.

generator/log/values.go
L2D2Grafana: D'oh removed

generator/log/values.go
L2D2Grafana: Updated the name and comments that this is a flag for the ci data which has a faster sleep cycle.

L2D2Grafana:
@matyax updated this per all the comments. All this stuff is for local dev only and allows us to run the apps together with fake data 🧹

matyax (approved):
🚀 ☄️

PR #1757, opened 2026-02-23
fix(augurs): fallback if wasm fails

L2D2Grafana:
https://raintank-corp.slack.com/archives/C075BDBTX96/p1771844319780149
This will unblock the plugin to load if there is an error in wasm, but is not the long-term fix. 

#### Problem
Thank @sd2k for figuring this one out 🙏 
The Proxy issue comes from OpenTelemetry’s @opentelemetry/instrumentation-fetch, which Faro wires up when tracing is enabled. OpenTelemetry's @opentelemetry/instrumentation-fetch wraps globalThis.fetch via this._wrap(globalThis, 'fetch', ...) Grafana instances with Faro/OpenTelemetry tracing enabled have the fetch instrumentation active, which wraps globalThis.fetch. Instances without tracing use the native fetch and work fine.

### DIFF
--- a/src/module.tsx
+++ b/src/module.tsx
@@ -26,7 +26,11 @@ const App = lazy(async () => {
   initRuntimeDs();
 
   if (wasmSupported()) {
-    await Promise.all([initChangepoint(), initOutlier()]);
+    try {
+      await Promise.all([initChangepoint(), initOutlier()]);
+    } catch (e) {
+      console.warn('grafana-lokiexplore-app: WebAssembly init failed, ML sorting disabled.', e);
+    }
   }
 
   return import('Components/App');


matyax (approved):

src/module.tsx
```diff
@@ -26,7 +26,11 @@ const App = lazy(async () => {
   initRuntimeDs();
 
   if (wasmSupported()) {
-    await Promise.all([initChangepoint(), initOutlier()]);
```
matyax, line 29: This is fine, but I think the app will still be broken because if wasm is supported, we will try to use outlier detection and changepoint detection, assuming they were initialized.

matyax: Okay, I tried commenting the initialization and trying to sort, and it seems that we properly handle errors from Augurs, and it doesn't break the app, so this should work as a temporary fix 👍 

Image

L2D2Grafana: Yes it does fallback in the app. I will merge this and keep working on a better fix!

sd2k:
FYI this was fixed upstream in https://github.com/open-telemetry/opentelemetry-js/issues/6518.

PR #1762, opened 2026-02-23
fix(wasm): check if wasm enabled and no errors

L2D2Grafana:
Follow up to https://github.com/grafana/logs-drilldown/pull/1757 & https://raintank-corp.slack.com/archives/C075BDBTX96/p1771844319780149

This is a more thoughtful implementation for a fallback if there is an error with the augurs library. isWasmInit checks if webassembly is supported and if there was an error initialization grafana/augurs. Updated augurs to version 0.10.2 which should include the [fallback](https://github.com/grafana/augurs/pull/445) when instantiateStreaming fails.

### Testing
1. Don't change anything and setWasmInit is true
2. Observe the sorting modes
Screenshot 2026-02-23 at 1 09 13 PM

1. Throw and error or set setWasmInit(false);
4. Observe only fallback sorting modes
Screenshot 2026-02-23 at 1 08 49 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "@grafana/i18n": "^12.4.0",
     "@grafana/lezer-logql": "^0.3.0",
     "@grafana/runtime": "^12.4.0",
-    "@grafana/scenes": "^6.39.0",
+    "@grafana/scenes": "^6.57.0",
     "@grafana/scenes-react": "^6.52.12",
     "@grafana/schema": "^12.4.0",
     "@grafana/ui": "^12.4.0",
--- a/provisioning/datasources/default.yaml
+++ b/provisioning/datasources/default.yaml
@@ -2,11 +2,11 @@ apiVersion: 1
 
 datasources:
   - name: gdev-testdata
-    isDefault: true
     type: testdata
   - name: gdev-loki
     type: loki
     uid: gdev-loki
+    isDefault: true
     access: proxy
     url: http://host.docker.internal:3100
     jsonData:
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -133,7 +133,7 @@ import {
 } from 'services/variables';
 export const showLogsButtonSceneKey = 'showLogsButtonScene';
 
-const DEFAULT_LOGS_DATASOURCE_UID = 'grafanacloud-logs';
+const FALLBACK_DATASOURCE_UID = 'grafanacloud-logs';
 
 interface EmbeddedIndexSceneConstructor {
   datasourceUid?: string;
@@ -151,7 +151,7 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
       jsonData?.dataSource ??
       getLastUsedDataSourceFromStorage() ??
       getDefaultDatasourceFromDatasourceSrv() ??
-      DEFAULT_LOGS_DATASOURCE_UID;
+      FALLBACK_DATASOURCE_UID;
 
     const { unsub, variablesScene } = getVariableSet(
       datasourceUid,
@@ -811,6 +811,7 @@ function getVariableSet(
     name: VAR_LABELS,
     onAddCustomValue: onAddCustomAdHocValue,
     filters: initialLabelFilters ?? [],
+    inputPlaceholder: 'Filter by labels',
   });
 
   labelVariable._getOperators = function () {
@@ -863,6 +864,7 @@ function getVariableSet(
     name: VAR_FIELDS_AND_METADATA,
     onAddCustomValue: onAddCustomFieldValue,
     skipUrlSync: true,
+    inputPlaceholder: 'Filter by fields',
   });
 
   const levelsVariable = new AdHocFiltersVariable({
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTextSearch.test.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTextSearch.test.tsx
@@ -1,0 +1,157 @@
+import React, { useMemo, useState } from 'react';
+
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import { sceneGraph } from '@grafana/scenes';
+
+import { PatternTextSearchComponent, PatternsViewTextSearch } from './PatternsViewTextSearch';
+
+const NON_INDEXED_FILTERS_ALERT_TEXT =
+  /Parsed fields, structured metadata, and string filters are not supported for the pattern list/;
+
+const mockGetAncestor = jest.spyOn(sceneGraph, 'getAncestor');
+const mockGetFieldsVariable = jest.fn();
+const mockGetMetadataVariable = jest.fn();
+const mockGetLineFiltersVariable = jest.fn();
+
+jest.mock('../../../../services/variableGetters', () => ({
+  getFieldsVariable: (...args: unknown[]) => mockGetFieldsVariable(...args),
+  getMetadataVariable: (...args: unknown[]) => mockGetMetadataVariable(...args),
+  getLineFiltersVariable: (...args: unknown[]) => mockGetLineFiltersVariable(...args),
+}));
+
+function createVariableMock(filters: Array<{ key: string; operator: string; value: string }>) {
+  return {
+    useState: () => ({ filters }),
+  };
+}
+
+function TestWrapper({ model, onSetState }: { model: PatternsViewTextSearch; onSetState: jest.Mock }) {
+  const [patternFilter, setPatternFilter] = useState('');
+  const breakdownScene = useMemo(
+    () => ({
+      setState: (update: { patternFilter?: string }) => {
+        if (update.patternFilter !== undefined) {
+          setPatternFilter(update.patternFilter);
+          onSetState(update);
+        }
+      },
+      useState: () => ({ patternFilter }),
+    }),
+    [patternFilter, onSetState]
+  );
+  mockGetAncestor.mockReturnValue(breakdownScene);
+  return <PatternTextSearchComponent model={model} />;
+}
+
+describe('PatternsViewTextSearch', () => {
+  let setStateSpy: jest.Mock;
+  let model: PatternsViewTextSearch;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    setStateSpy = jest.fn();
+    mockGetFieldsVariable.mockReturnValue(createVariableMock([]));
+    mockGetMetadataVariable.mockReturnValue(createVariableMock([]));
+    mockGetLineFiltersVariable.mockReturnValue(createVariableMock([]));
+    model = new PatternsViewTextSearch();
+  });
+
+  describe('PatternTextSearchComponent', () => {
+    it('renders search input with placeholder', () => {
+      mockGetAncestor.mockReturnValue({
+        setState: jest.fn(),
+        useState: () => ({ patternFilter: '' }),
+      });
+      render(<PatternTextSearchComponent model={model} />);
+
+      expect(screen.getByPlaceholderText('Search patterns')).toBeInTheDocument();
+    });
+
+    it('does not show alert when no non-indexed filters are present', () => {
+      mockGetAncestor.mockReturnValue({
+        setState: jest.fn(),
+        useState: () => ({ patternFilter: '' }),
+      });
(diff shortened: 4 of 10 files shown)


src/module.tsx
```diff
@@ -28,8 +28,10 @@ const App = lazy(async () => {
   if (wasmSupported()) {
     try {
       await Promise.all([initChangepoint(), initOutlier()]);
+      setWasmInit(true);
     } catch (e) {
       console.warn('grafana-lokiexplore-app: WebAssembly init failed, ML sorting disabled.', e);
+      setWasmInit(false);
     }
```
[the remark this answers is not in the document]
L2D2Grafana: It is false by default.

L2D2Grafana:
> @L2D2Grafana, here is some (LLM) feedback. It seemed reasonable.
> 
Okay I updated per some of the comments, mostly the memoize one. It's not actually a problem since the app is only initialized once, but I suppose it helps with testing.

nicwestvold:
@L2D2Grafana I'm ready to approve, but it looks like there is a test failure with 12.4.0. Here is what claude had to say:

---

CI Failure Analysis

Why only 12.4.0?

`skipUnlessLatestGrafana` skips tests on versions below `GRAFANA_LATEST_SUPPORTED_VERSION = '12.3.1'`. So 11.6, 12.0, and 12.1 skip the savedSearches tests entirely — they don't pass, they skip. The tests actually run on 12.3.4, 12.4.0, and dev only.

Root cause

The PR changed two assertions in savedSearches.spec.ts:

Before (both test assertions, lines ~104 and ~249):

```ts
page.getByRole('button', { name: /Edit filter with key service_name/ })
  .filter({ hasText: 'tempo-ingester' })
```

After:

```ts
page.getByLabel('Edit filter with key service_name')
// ...toContainText('tempo-ingester')
```

The failing assertion receives `"service_name = tempo-distributor"` instead of `"tempo-ingester"`. The element is found — it's just the wrong service.

In Grafana 12.4.0 the ad-hoc filter chip structure appears to have changed: `getByLabel('Edit filter with key service_name')` matches the chip element, but the chip's text contains "tempo-distributor". This indicates the label format `'Edit filter with key service_name'` may now match something other than the active filter chip — possibly an input, a label element, or a chip rendered differently in 12.4.0 that happens to show an earlier filter value.

The old selector was resilient to this because it combined role + regex name + `filter({ hasText: 'tempo-ingester' })`, meaning it would only match if the element also contained the expected service name. With the new approach, the first matching element is found regardless of its value, and then `toContainText` fails.

Fix

Revert the two filter chip assertions back to the original `getByRole + filter` style which is inherently self-asserting:

// tests/savedSearches.spec.ts ~line 104

```ts
await expect(
  page.getByRole('button', { name: /Edit filter with key service_name/ })
    .filter({ hasText: 'tempo-ingester' })
).toBeVisible();
```

The `deleteButton` scoping change (line ~155, scoped to modal) was a valid improvement and should be kept.

L2D2Grafana:
> @L2D2Grafana, here is some (LLM) feedback. It seemed reasonable.
> 
> ## PR Review: `l2d2/wasm-fallback`
> **Overall:** The approach is sound — the shift from `wasmSupported()` (browser check only) to `isWasmInit()` (captures actual initialization success) is the right fix. The fallback to `stdDev`, UI filtering of ML options, and constructor preference override are all well-handled. A few issues worth addressing before merge.
> 
> ### Bugs / Correctness
> **[Medium] `memoize` cache is blind to WASM state — stale result risk** `src/services/sorting.ts:75-91`
> 
> The cache key for `sortSeries` includes `sortBy` and `direction` but **not** `wasmInitSucceeded`. If `sortSeries(series, 'changepoint', 'desc')` were ever called before WASM initialized, the try-catch fallback on line 49 silently mutates `sortBy` to `ReducerID.stdDev` and the stdDev-sorted result gets cached under a key containing `'changepoint'`. A later call with the same args after WASM succeeds returns the stale result.
> 
> In the current flow this can't happen (WASM is fully initialized before the `App` component loads), but the invariant is unprotected. Recommend appending `_${isWasmInit()}` to the cache key on line 90.
> 
> **[Low] `isWasmInit()` in Component render is not reactive** `src/Components/ServiceScene/Breakdowns/SortByScene.tsx:113`
> 
> ```tsx
> const wasmInit = isWasmInit();
> 
> This reads a plain module-level boolean — not Grafana Scenes state — so no state change will trigger a re-render. Safe today because WASM is initialized exactly once before any scene renders, but the ordering
> invariant is implicit. A comment documenting it would prevent a future regression if retry/re-init logic is ever added.
> 
> ---
> Missing Test Coverage
> 
> [Medium] No test for stored-preference override in constructor
> src/Components/ServiceScene/Breakdowns/SortByScene.test.tsx
> 
> The most important regression path of this PR — "user has 'changepoint' stored in localStorage, WASM fails on load, constructor should fall back to stdDev" — has no test. Suggested case:
> 
> test('Falls back to stdDev when stored preference is changepoint but WASM init failed', () => {
>   setSortByPreference('fields', 'changepoint', 'desc');
>   setWasmInit(false);
>   scene = new SortByScene({ target: 'fields' });
>   expect(scene.state.sortBy).toBe(ReducerID.stdDev);
> });
> 
> [Low] New sorting.ts exports have no unit tests
> src/services/sorting.test.ts
> 
> setWasmInit, isWasmInit, and getDefaultSortBy are untested. At minimum getDefaultSortBy should be tested for both branches.
> 
> ---
> Nits
> 
> Duplicate test names — SortByScene.test.tsx:46,56
> Both tests are named 'Reports criteria changes'. Line 56 tests direction change and should be renamed to 'Reports direction changes'.
> 
> Hardcoded 'changepoint' in test assertion — SortByScene.test.tsx:63
> expect(eventSpy).toHaveBeenCalledWith(new SortCriteriaChanged('fields', 'changepoint', 'asc'), true);
> Should use the DEFAULT_SORT_BY constant to stay in sync if the value ever changes.
> 
> setWasmInit(false) in catch is redundant — module.tsx:34
> The flag defaults to false and is only set to true on success, so this call is a no-op. Harmless, but a comment explaining "explicit reset in case of partial initialization" would clarify intent.
> 
> defaultOptions re-allocated every render — SortByScene.tsx:114-121
> When WASM is not initialized, .map(...).filter(...) creates new arrays on every render. Since isWasmInit() is stable after init, a useMemo([wasmInit]) would avoid the repeated allocation.
> 
> console.warn vs. logger — module.tsx:33
> Rest of the codebase uses the logger service for errors. console.warn may be intentional here given the module entry-point bundle size constraint, but worth a comment.
> 
> ---
> What's Well Done
> 
> - Default wasmInitSucceeded = false means the fallback is automatic if setWasmInit(true) is never reached (e.g., unsupported browser) — no extra guard needed.
> - All 4 breakdown scenes (FieldValues, Fields, Label, LabelValues) consistently updated — no missed call sites.
> - Moving SortBy type and constants to services/sorting.ts removes the circular dependency risk from importing them via SortByScene.
> - E2E selector fixes (getByLabel, modal-scoped getByRole) are strictly better and avoid false matches.
> - The constructor finalSortBy guard correctly handles the persisted-preference-but-WASM-failed scenario at the point of object creation.
> 
> ---
> Summary
> 
> | Severity |            File            |                            Issue                            |
> | Medium   | sorting.ts:75-91           | memoize key missing WASM state                              |
> | Medium   | SortByScene.test.tsx       | Missing test: stored preference override in constructor     |
> | Low      | SortByScene.tsx:113        | isWasmInit() not reactive — document the ordering invariant |
> | Low      | sorting.test.ts            | No unit tests for new exports                               |
> | Nit      | SortByScene.test.tsx:46,56 | Duplicate test names                                        |
> | Nit      | SortByScene.test.tsx:63    | Hardcoded 'changepoint' instead of DEFAULT_SORT_BY          |
> | Nit      | module.tsx:34              | Redundant setWasmInit(false)                                |
> | Nit      | SortByScene.tsx:114        | defaultOptions re-allocated each render                     |
> | Nit      | module.tsx:33              | console.warn vs. logger inconsistency                       |
> ```



> @L2D2Grafana I'm ready to approve, but it looks like there is a test failure with 12.4.0. Here is what claude had to say:
> 
> CI Failure Analysis
> 
> Why only 12.4.0?
> 
> `skipUnlessLatestGrafana` skips tests on versions below `GRAFANA_LATEST_SUPPORTED_VERSION = '12.3.1'`. So 11.6, 12.0, and 12.1 skip the savedSearches tests entirely — they don't pass, they skip. The tests actually run on 12.3.4, 12.4.0, and dev only.
> 
> Root cause
> 
> The PR changed two assertions in savedSearches.spec.ts:
> 
> Before (both test assertions, lines ~104 and ~249):
> 
> ```ts
> page.getByRole('button', { name: /Edit filter with key service_name/ })
>   .filter({ hasText: 'tempo-ingester' })
> ```
> 
> After:
> 
> ```ts
> page.getByLabel('Edit filter with key service_name')
> // ...toContainText('tempo-ingester')
> ```
> 
> The failing assertion receives `"service_name = tempo-distributor"` instead of `"tempo-ingester"`. The element is found — it's just the wrong service.
> 
> In Grafana 12.4.0 the ad-hoc filter chip structure appears to have changed: `getByLabel('Edit filter with key service_name')` matches the chip element, but the chip's text contains "tempo-distributor". This indicates the label format `'Edit filter with key service_name'` may now match something other than the active filter chip — possibly an input, a label element, or a chip rendered differently in 12.4.0 that happens to show an earlier filter value.
> 
> The old selector was resilient to this because it combined role + regex name + `filter({ hasText: 'tempo-ingester' })`, meaning it would only match if the element also contained the expected service name. With the new approach, the first matching element is found regardless of its value, and then `toContainText` fails.
> 
> Fix
> 
> Revert the two filter chip assertions back to the original `getByRole + filter` style which is inherently self-asserting:
> 
> // tests/savedSearches.spec.ts ~line 104
> 
> ```ts
> await expect(
>   page.getByRole('button', { name: /Edit filter with key service_name/ })
>     .filter({ hasText: 'tempo-ingester' })
> ).toBeVisible();
> ```
> 
> The `deleteButton` scoping change (line ~155, scoped to modal) was a valid improvement and should be kept.

Thanks, yeah the savedSearches e2e have been flaky and the compatibility was also failing!

src/module.tsx
```diff
@@ -28,8 +29,10 @@ const App = lazy(async () => {
   if (wasmSupported()) {
     try {
       await Promise.all([initChangepoint(), initOutlier()]);
+      setWasmInit(true);
```
matyax: This is technically incorrect, because there's no WASM initialization. It's either supported or not, and what's initialized is Augurs. 

```suggestion
      setWasmSortInit(true);
```

package.json
```diff
@@ -105,7 +105,7 @@
     "node": ">=24"
   },
   "dependencies": {
-    "@bsull/augurs": "^0.6.0",
+    "@bsull/augurs": "^0.10.2",
```
matyax, line 108: Awesome. No breaking changes?

matyax (changes requested):
Looking good! I'd appreciate renaming the initialization function because it's not properly reflecting the underlying situation.

src/module.tsx
matyax: Even better, since today we're using WASM but tomorrow it might be something else, and we don't need to update these functions just because the internal implementation changed.

```suggestion
      setSortingInit(true);
```

package.json
L2D2Grafana: Nope!

matyax (approved):
🚀

src/services/sorting.ts
```diff
@@ -9,23 +9,37 @@ import { getLabelValueFromDataFrame } from './levels';
...
+
+export type SortBy = typeof DEFAULT_SORT_BY | typeof SORT_BY_OUTLIERS | ReducerID | '';
+
+let wasmInitSucceeded = false;
+
+export const setWasmSortInit = (succeeded: boolean) => {
+  wasmInitSucceeded = succeeded;
+};
+
+export const isWasmInit = () => wasmInitSucceeded;
```
matyax, line 23:
```suggestion
export const setSortingInit = (succeeded: boolean) => {
  wasmInitSucceeded = succeeded;
};

export const isSortingInit = () => wasmInitSucceeded;
```

Worst case:

```suggestion
export const setWasmSortingInit = (succeeded: boolean) => {
  wasmInitSucceeded = succeeded;
};

export const isWasmSortingInit = () => wasmInitSucceeded;
```

Apologies, multitasking a lot.

PR #1766, opened 2026-02-24
chore(goff): update exploreLogsShardSplitting to goff

L2D2Grafana:
- #1704
Updates exploreLogsShardSplitting to the new GOFF paradigm.

#### Testing
- Add the feature flag exploreLogsShardSplitting
- Choose last 7 days as a time range
 
Screenshot 2026-02-24 at 2 31 40 PM


matyax:
Is there a deployment_tools PR that sets the feature flags value to the current state in the old system?

L2D2Grafana:
> Is there a deployment_tools PR that sets the feature flags value to the current state in the old system?

Here is a PR to add lokiShardSplitting into deployment tools. Since exploreLogsShardSplitting was already there, I copied what it was doing. Once lokiShardSplitting rolls out, we can remove the depreciated exploreLogsShardSplitting.
[Deployment tools](https://github.com/grafana/deployment_tools/pull/497781). I wasn't sure if this needed to be in deployment tools, because may it was turned on ad-hoc for users like aggregatedMetrics. In core lokiShardSplitting is used without anything in deployment tools.

matyax:
No, we should not enable lokiShardSplitting anywhere. It's opt-in and experimental.

What we need to make sure is that the new feature flag system has `exploreLogsShardSplitting`. We need to keep `exploreLogsShardSplitting` for people that needs this behavior to be disabled, like we do with `lokiQuerySplitting`. I should have clarified that before.

L2D2Grafana:
> No, we should not enable lokiShardSplitting anywhere. It's opt-in and experimental.
> 
> What we need to make sure is that the new feature flag system has `exploreLogsShardSplitting`. We need to keep `exploreLogsShardSplitting` for people that needs this behavior to be disabled, like we do with `lokiQuerySplitting`. I should have clarified that before.

My bad I thought we wanted to shift people off the depreciated exploreLogsShardSplitting and use lokiShardSplitting. So I've updated this PR to continue using exploreLogsShardSplitting, however with the new GOFF. Someone in core has piped all the old feature toggles to the new /flags endpoint. So we do not add them. 
Screenshot 2026-02-27 at 8 22 07 AM

matyax:
Oh, fantastic then!

matyax:
I'm thinking this may also affect retrocompatibility. We should probably sync this one with React 19.

PR #1792, opened 2026-03-06
chore(DefaultColumns): replace local rtkq with @grafana/api-clients

gtk-grafana:
Delete local rtkq copy!!!!


L2D2Grafana (approved):
Sharing code 🥳 !

Image

PR #1795, opened 2026-03-09
chore(cve): update immutable, go, dompurify, serialize-javascript

L2D2Grafana:
https://github.com/grafana/logs-drilldown/security/dependabot
Screenshot 2026-03-09 at 9 10 43 AM


package.json
```diff
@@ -139,6 +139,9 @@
     "react-router-dom": "^7.12.0",
     "@remix-run/router": "1.23.2",
     "uuid": "^10.0.0",
+    "immutable": "5.1.5",
```
gtk-grafana, line 142: immutable and dompurify are deps of grafana/data and grafana/ui, instead of pinning a downstream dep can we just update the version of those packages?

L2D2Grafana: Latest published versions right now are @grafana/ui@12.4.1 and @grafana/data@12.4.1, and they still declare:

@grafana/ui -> immutable: 5.1.4 (flagged; fix is 5.1.5+)
@grafana/data -> dompurify: 3.3.0 (flagged; fix is 3.3.2+)

gtk-grafana: 😢

PR #1809, opened 2026-03-19
fix(ServiceSelection): Reset scene when changing tabs

matyax:
Specially reproducible with slow data sources, when changing between tabs in the Service Selection, loading of new values depends on the Volume endpoint, so the previous values remains on-screen, with potentially confusing effects.

We're now resetting the scene on tab change. Additional minor fixes have been added as well.

> [!IMPORTANT]  
> We need to double check changing data sources, time range, tabs, etc, to make sure there's no regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 2.0.0
+
+##### Chores
+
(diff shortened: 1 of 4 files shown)


matyax:
That should be enough, thanks for testing!

PR #1819, opened 2026-03-26
feat: add Knowledge Graph insights annotations to timeseries

rdubrock:
## Summary
- Adds KG annotation data layers (critical/warning/info) that overlay assertion health states as colored time regions on timeseries panels
- Gated behind the `kgAnnotationsInLokiExplore` feature flag and requires the KG datasource (`grafanacloud-knowledgegraph`) to be present
- Adds an "Insights" toggle control to the EmbeddedScene controls bar

Depends on this PR in the asserts plugin: https://github.com/grafana/asserts-app-plugin/pull/3038

Tracking issue: https://github.com/grafana/asserts-app-plugin/issues/2921 (part of epic https://github.com/grafana/asserts-app-plugin/issues/2919)

Screenshot 2026-03-27 at 14 40 27
Screenshot 2026-03-27 at 14 43 57

## Test plan
- [ ] Enable `kgAnnotationsInLokiExplore` feature flag and verify the "Insights" toggle appears
- [ ] Verify annotation layers render colored time regions on timeseries panels for critical (red), warning (orange), info (blue)
- [ ] Verify toggling the switch on/off enables/disables all annotation layers
- [ ] Verify no annotations or toggle appear when feature flag is disabled
- [ ] Verify no annotations or toggle appear when KG datasource is not configured

### DIFF
--- a/docker-compose-defaults.yaml
+++ b/docker-compose-defaults.yaml
@@ -12,5 +12,5 @@ services:
       GF_SERVER_SERVE_FROM_SUB_PATH: true
       GF_SERVER_ROOT_URL: http://localhost:3001/grafana
       GF_PLUGINS_PREINSTALL_DISABLED: true
-      GF_FEATURE_TOGGLES_ENABLE: accessControlOnCall lokiLogsDataplane kubernetesLogsDrilldown ${GF_FEATURE_TOGGLES_ENABLE}
+      GF_FEATURE_TOGGLES_ENABLE: accessControlOnCall lokiLogsDataplane kubernetesLogsDrilldown localizationForPlugins ${GF_FEATURE_TOGGLES_ENABLE}
       GRAFANA_CLOUD_LOGS_KEY: ${GRAFANA_CLOUD_LOGS_KEY}
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,9 +1,11 @@
 import { FlatCompat } from '@eslint/eslintrc';
 import { defineConfig } from 'eslint/config';
 import importPlugin from 'eslint-plugin-import';
-// Plugins
 import jsxa11y from 'eslint-plugin-jsx-a11y';
 import sortPlugin from 'eslint-plugin-sort';
+
+// eslint-disable-next-line import/no-unresolved -- package.json exports field not recognized by import plugin
+import grafanaI18nPlugin from '@grafana/i18n/eslint-plugin';
 
 import baseConfig from './.config/eslint.config.mjs';
 
@@ -122,6 +124,26 @@ export default defineConfig([
       'react-hooks/set-state-in-effect': 'off',
       'react-hooks/immutability': 'off',
       'react-hooks/refs': 'off',
+    },
+  },
+
+  {
+    name: 'logs-drilldown/i18n',
+    files: ['src/**/*.{ts,tsx}'],
+    ignores: ['**/*.test.{ts,tsx}', '**/__tests__/**', '**/test/**', '**/mocks/**'],
+    plugins: {
+      '@grafana/i18n': grafanaI18nPlugin,
+    },
+    rules: {
+      '@grafana/i18n/no-untranslated-strings': [
+        'warn',
+        {
+          basePaths: ['src'],
+          namespace: 'grafana-lokiexplore-app',
+          calleesToIgnore: ['^css$', 'use[A-Z].*', '^get.*Styles$'],
+        },
+      ],
+      '@grafana/i18n/no-translation-top-level': 'warn',
     },
   },
 ]);
--- a/i18next.config.ts
+++ b/i18next.config.ts
@@ -1,0 +1,19 @@
+import { defineConfig } from 'i18next-cli';
+
+export default defineConfig({
+  locales: ['en-US'],
+  extract: {
+    input: ['src/**/*.{ts,tsx}'],
+    output: 'src/locales/{{language}}/{{namespace}}.json',
+    defaultNS: 'grafana-lokiexplore-app',
+    keySeparator: '.',
+    nsSeparator: false,
+    contextSeparator: '_',
+    pluralSeparator: '_',
+    defaultValue: '',
+    sort: true,
+    indentation: 2,
+    functions: ['t', '*.t'],
+    transComponents: ['Trans'],
+  },
+});
--- a/package.json
+++ b/package.json
@@ -32,7 +32,9 @@
     "sign": "npx --yes @grafana/sign-plugin@latest",
     "prepare": "husky",
     "docs": "cd docs && make docs",
-    "spellcheck": "cspell -c cspell.config.json \"**/*.{ts,tsx,js,go,md,mdx,yml,yaml,json,scss,css}\""
+    "spellcheck": "cspell -c cspell.config.json \"**/*.{ts,tsx,js,go,md,mdx,yml,yaml,json,scss,css}\"",
+    "i18n:extract": "i18next-cli extract -c i18next.config.ts",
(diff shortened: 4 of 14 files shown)


rdubrock:
Tracking issue: https://github.com/grafana/asserts-app-plugin/issues/2921 (part of epic https://github.com/grafana/asserts-app-plugin/issues/2919)

L2D2Grafana:
The code looks fine. I tried to get it running locally by adding the feature flag and asserts app, but could quite get the calls to the datasource to work. It adds an insights toggle at the top of the page. 
Screenshot 2026-03-27 at 11 55 54 AM

matyax:
Thanks for adding this integration to Logs Drilldown! One thing that we need to improve is the position of the toggle, which currently looks out of place.

imagen

It could be added as a new toggle to this dropdown:

imagen

We could also add a tooltip to it, explainig what it is. It's also pretty of out place in the filters section, but unfortunately I can't think of an alternative place for it.

imagen

rdubrock:
Good call @matyax! I went ahead and added to that ToolbarScene, also added an info tooltip:

Screenshot 2026-03-30 at 13 31 57

Also I made the annotations multilane so that when there are overlapping annotations of differing severities they all show up (it's clamped to one lane per severity, so a max of 3)

Screenshot 2026-03-30 at 13 34 02

I've created a new API to help with finding insights based on labels. It seems like it's working in the cases I know to check for, but there will probably need to be a few iterations required to make it work more often. The goal is that when it works, it's correct. If it's not working, it'll just fail silently and no insights will show up.

matyax:
I migrated the feature flag to openFeature in this PR, I hope you don't mind.

Just to be on the same page, should this toggle appear when Logs Drilldown is embedded in Asserts, or only in regular (not embedded) Logs Drilldown?

rdubrock:
> I migrated the feature flag to openFeature in this PR, I hope you don't mind.
> 
> Just to be on the same page, should this toggle appear when Logs Drilldown is embedded in Asserts, or only in regular (not embedded) Logs Drilldown?

I'm totally easy/flexible on basically everything. If openFeature works best for you then that's great. I was thinking it'd be cool if the toggle/insights showed up all the time, the goal was to have it only really show up for users where it's useful. But I'm also fine keeping just in the asserts drawer if that feels better. These annotations are getting added all over the place, mostly with the intent of integrating the graph more deeply without requiring users to actually visit the graph UI. But logs drilldown is a different use case than some of the others (like say DB o11y, which is already more tightly tied to o11y and asserts to begin with)

matyax:
Also made sure disabling aggregated metrics didn't prevent access to this new toggle:

Toggle

matyax:
@rdubrock I was asking because we can't get it to appear, even if forcing it to be enabled, in our testing embedded route. 

imagen

If you confirm that this works when Logs Drilldown is embedded then it's all good 👍

src/Components/ServiceScene/LogsVolume/InsightsWidgetScene.tsx
```diff
@@ -1,11 +1,9 @@
-import React from 'react';
+import { sceneGraph, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
 
-import { SceneComponentProps, sceneGraph, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
-
-import { isOperatorInclusive } from '../../../services/operatorHelpers';
-import { getLabelsVariable } from '../../../services/variableGetters';
-import { SERVICE_NAME } from '../../../services/variables';
-import { InsightsTimelineWidget } from '../../AddedComponents/InsightsTimelineWidget';
+// import { isOperatorInclusive } from '../../../services/operatorHelpers';
```
matyax: Do we need to keep the commented code in this component?

rdubrock: Ah yes good catch. I'm not really sure, there was a previous attempt to integrate insights into the logs drilldown via importing a component from the asserts app and rendering it in certain situations. I think this new approach of using built in annotations is cleaner, but I didn't want to delete this without running it by you first

rdubrock:
Screenshot 2026-03-31 at 08 21 40

Yeah it's showing up for me in the drawer. It won't show up if there isn't a datasource with UID `grafanacloud-knowledgegraph`.

rdubrock:
Feature flag PR here: https://github.com/grafana/deployment_tools/pull/535232/changes

src/Components/ServiceScene/LogsVolume/InsightsWidgetScene.tsx
matyax: Yeah, I think this is safe to remove.

matyax: Removed! Again, given that you have a working set up, please double check that the KG annotations are working as expected. Thanks!

src/services/kgAnnotations.ts
```diff
@@ -0,0 +1,163 @@
...
+    return false;
+  }
+  return Object.values(config.datasources).some((d) => d.uid === KG_DATASOURCE_UID);
+}
+
+function createAnnotationLayers(labels: Record<string, string>, datasourceUid: string) {
+  const severities = [
+    { value: 'critical', color: 'red', label: 'Critical' },
+    { value: 'warning', color: 'orange', label: 'Warning' },
+    { value: 'info', color: 'blue', label: 'Info' },
```
matyax, line 38: Just to be sure. Do these values need to match the levels in the logs, or are these just generic? Because otherwise some users might have crit, warn, info, etc, which we would need to read from the data frames.

rdubrock: These are the "Insight" severities, not log severities. The knowledge graph only has these three levels, and they are consistent. We list them here so we can fire off one annotation request per severity and have a lane for each

matyax: Perfect, thanks!

matyax:
From my perspective this looks ready! I think you haven't yet signed the [Contributor License Agreement](https://cla-assistant.io/grafana/logs-drilldown?pullRequest=1819), and I just approved the feature flag PR.

rdubrock:
The CLA assistant is very unhappy 😂 
Screenshot 2026-04-01 at 07 12 59

rdubrock:
@matyax Claude feels like the e2e failures aren't tied to these code changes. This was a suggested plan to resolve. Does this look appealing, or do you think it's wrong in this case? 



```

  The page snapshot shows the Grafana home page is loaded in beforeEach. The home page has a "Latest from the blog" section with a News panel that shows "Error loading RSS feed". The text "Error
  loading RSS feed" matches the /loading/i regex in assertNotLoading().

  This is the Grafana 12.0.10 home page's RSS news panel failing to fetch its feed (expected in a CI/Docker environment with no outbound internet). The text "Error loading RSS feed" contains the
  word "loading" which triggers the assertion failure.

  But why did this pass before? The beforeEach runs assertNotLoading() on whatever page the browser starts at. Previously, the page likely started at about:blank, but something about this PR or the
  test environment now causes the page to start at the Grafana home page. Or more likely — this has always been borderline/flaky on 12.0.10, and a timing change (slightly larger bundle from the new
  files) now causes the home page to consistently render the error before assertNotLoading() runs.

  This is NOT caused by the KG annotations code itself — it's a pre-existing fragility in the assertNotLoading() helper. The /loading/i regex is too broad and matches "Error loading RSS feed" on the
   Grafana home page. The fix should be in the test helper, not in the feature code.

  Possible fixes:
  1. Make assertNotLoading() more specific — match a plugin-specific loading indicator rather than any text containing "loading"
  2. Navigate to about:blank in beforeEach before asserting
  3. Add an exclusion for the RSS feed error text
  
  ```

PR #1830, opened 2026-04-06
feat(errors): expose AttributeDistribution component via UI extension

bohandley:
[app-o11y-kwl companion PR](https://github.com/grafana/app-o11y-kwl/pull/3108)

Screenshot 2026-04-20 at 3 12 39 PM

https://github.com/user-attachments/assets/4b1b7197-9fa6-46ea-a890-31fe224f949c

## Problem

Understanding what a set of log events have in common requires writing custom queries or building manual breakdowns. There is no way for a consumer app to give users an automatic, interactive breakdown of their data by every detected field.

## What this adds

A generic `AttributeDistribution` component and a Loki adapter (`LokiFieldDistribution`) that together expose a full-featured attribute breakdown sidebar. The component is exposed via Grafana's UI Extensions API so app plugins can embed it with no build-time dependency on logs-drilldown.

Extension ID: `grafana-lokiexplore-app/errors-analysis/v1`

## UX behaviors

**Attribute breakdown**
- Detects all logfmt fields in the current dataset via Loki's `detected_fields` API and shows each as a section with percentage bars per value
- Each section shows the value name, event count, and percentage
- Sections load and stream in independently as results arrive

**Priority attributes**
- The consumer specifies an ordered list of fields to pin at the top
- Priority fields are always shown even when absent from `detected_fields` - absence is itself informative (e.g. an error that never occurs on iOS)
- Remaining detected fields follow below, revealed in batches via expand/collapse chevrons

**Filtering**
- Clicking any value opens a menu: "Filter for value" (include) or "Filter out value" (exclude)
- Selected values are highlighted in the sidebar; the same filters appear in the page-level filter bar
- Removing a filter via the page-level filter bar clears the highlight in the sidebar
- Distributions re-evaluate against the filtered subset after every selection
- Consumer is notified of the full filter set via `onFiltersChange`

**Sticky values**
- On first filter selection, all current values are snapshotted per field
- Values that drop out of the filtered distribution remain visible at 0% and dimmed while any filter is active, so users never lose the full option set

**External filter sync**
- Filters applied outside the component (e.g. from a page-level filter bar) are reflected in the sidebar on load and when they change
- Filters removed externally clear from the sidebar

**Pinning attributes**
- A search combobox lets users find any already-detected field and pin it to the top of the list
- Only fields surfaced by auto-detection are available. The combobox does not accept arbitrary field names

**"Show all" link**
- Optional `showAllLink` prop renders a navigation link at the bottom of the sidebar, built by the consumer

## Links

- Companion PR (app-o11y-kwl): https://github.com/grafana/app-o11y-kwl/pull/3108

## Checklist

- [x] Unit tests added for reducer, `orderByPriority`, and `mergeWithSnapshot`
- [ ] UI tested with both Light and Dark theme
- [ ] Documentation updated if needed


src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
```diff
@@ -0,0 +1,178 @@
...
+  // Set this when the underlying query caps the number of events so users understand
+  // the distributions are based on a sample. Example: "Last 1000 logs"
+  // The consuming app sets this -- it knows what limit its query applies.
+  queryLimitLabel?: string;
+  // See AttributeDistributionProps.showAllLink -- note the OR-query limitation there.
+  showAllLink?: { href: string; title: string };
+  timeRange: { from: number; to: number };
+}
+
+export default function ErrorsAnalysis({
```
matyax: If I'm reading this right, there's nothing specific for errors in these components, right? These could be ErrorAnalysis as much as DebugAnalysis, depending on the data provided by the consumer. If that is the case, the exposed component would actually be AttributeDistribution.

Just making sure I understand the intention here.

bohandley: Yes! Nothing specific for errors, correct 😄. It just takes a LogQL query, which could be what you mentioned, ErrorAnalysis or DebugAnalysis.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,959 @@
...
+    if (values.length === 1) {
+      q += ` | ${field}${operator}"${values[0]}"`;
+    } else {
+      const regexOp = operator === '=' ? '=~' : '!~';
+      const pattern = values.map(escapeRegex).join('|');
+      q += ` | ${field}${regexOp}"${pattern}"`;
+    }
+  }
+  return q;
+}
```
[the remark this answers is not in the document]
bohandley: Will add tests after testing with frontend o11y squad

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,959 @@
...
+  const visibleValues = expanded ? allValues.slice(0, MAX_VALUES_EXPANDED) : allValues.slice(0, MAX_VALUES_COLLAPSED);
+  const hasMore = allValues.length > MAX_VALUES_EXPANDED;
+  const isExpandable = allValues.length > MAX_VALUES_COLLAPSED;
+
+  return (
+    <div className={styles.section}>
+      <button className={styles.sectionHeader} onClick={isExpandable ? onToggle : undefined}>
+        <span className={styles.sectionLabel}>{config.label}</span>
+        {isExpandable && <Icon name={expanded ? 'angle-up' : 'angle-down'} size="sm" />}
+      </button>
```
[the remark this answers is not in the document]
bohandley: Fixed

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,959 @@
...
+                <div
+                  className={cx(
+                    styles.valueRow,
+                    isSelected && styles.valueRowSelected,
+                    item.retained && styles.valueRowRetained
+                  )}
+                  onClick={openMenu}
+                  role="button"
+                  tabIndex={0}
+                  onKeyDown={(e) => e.key === 'Enter' && openMenu(e as unknown as React.MouseEvent<HTMLElement>)}
```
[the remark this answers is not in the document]
bohandley: Removed

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,959 @@
...
+
+  let q = baseQuery;
+  for (const [field, { operator, values }] of byField) {
+    if (values.length === 1) {
+      q += ` | ${field}${operator}"${values[0]}"`;
+    } else {
+      const regexOp = operator === '=' ? '=~' : '!~';
+      const pattern = values.map(escapeRegex).join('|');
+      q += ` | ${field}${regexOp}"${pattern}"`;
+    }
```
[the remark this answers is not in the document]
bohandley: A user would have to deliberately type a value with a " in the free-text field input to trigger this, and the query comes from frontend o11y. Not implementing this.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,959 @@
...
+
+    run();
+
+    return () => {
+      cancelled = true;
+    };
+    // fetchAttributes is a stable module-level function; context fields and
+    // priorityAttributes are the true dependencies.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [context.query, context.datasourceUid, context.timeRange.from, context.timeRange.to, priorityAttributes]);
```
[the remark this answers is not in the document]
bohandley: Implementing useMemo for fetchAttributes.

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
matyax: In that case, I'd appreciate if you rename this to something closer to what it actually is, maybe `LokiAttributeDistribution`. Even more, since this is Loki and it's using the detected_fields endpoint, `LokiFieldDistrubiution`.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+  field: string;
+  operator: '!=' | '=';
+  value: string;
+}
+
+// A value entry extended with a `retained` flag used for the sticky values pattern.
+interface DisplayValue extends LabelValueCount {
+  // True for values from the pre-filter snapshot that are absent from the current
+  // filtered distribution. Shown at 0% and dimmed so the user can still see and
+  // interact with them after filtering narrows the result set.
```
matyax:
```suggestion
  // filtered distribution.
```
If the implementation changes, this comment will be unlikely to be updated and become obsolete.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+export interface AttributeConfig {
+  field: string;
+  label: string;
+}
+
+export interface LabelValueCount {
+  count: number;
+  percentage: number;
+  value: string;
+}
```
matyax:
```suggestion
export interface AttributeConfig {
  attribute: string;
  attribute_name: string; // or display_name
}

export interface AttributeValueCount {
  count: number;
  percentage: number;
  value: string;
}
```

Not only it's already mixing concepts, but since we've decided to call this component AttributeDistribution, to show the distribution of attributes, we should continue referring to what is being displayed as attributes for consistency. I wouldn't suggest to unify it to "labels" because it makes it specific for a certain type of attributes. For example, Loki has both labels and fields, different from each other.

In general:
- Refer to the data as attributes
- Avoid using `label`
- Avoid using `field`

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+  attributes: AttributeConfig[];
+  data: Record<string, AttributeState>;
+  detecting: boolean;
+  selectedFilters: ActiveFilter[];
+  // Snapshot of value lists per field, taken the moment the first filter is applied.
+  // Retained until all filters are cleared. null when no filters are active.
+  valueSnapshot: Record<string, LabelValueCount[]> | null;
+}
+
+type Action =
```
matyax: I'd appreciate it if you extract this to it's own module, as there's no need to have the whole implementation in a single file. It should make adding unit tests for all these reducer code simpler too.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+      return { ...state, selectedFilters: newFilters, valueSnapshot };
+    }
+    case 'CLEAR_FILTERS':
+      return { ...state, selectedFilters: [], valueSnapshot: null };
+    default:
+      return state;
+  }
+}
+
+function orderByPriority(detected: AttributeConfig[], priority: AttributeConfig[]): AttributeConfig[] {
```
matyax:
```suggestion
function orderByPriority(attributes: AttributeConfig[], priority: AttributeConfig[]): AttributeConfig[] {
```

Or

```suggestion
function orderByPriority(current: AttributeConfig[], priority: AttributeConfig[]): AttributeConfig[] {
```

For the Loki use case these would be detected fields, but generically these are just attributes, right?

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+
+// Escapes special RE2 regex characters in a literal value string.
+function escapeRegex(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+// Appends LogQL filter pipeline stages for active filters to the base query.
+// Multiple values for the same field and operator are combined via regex (=~ / !~).
+// Filters across different fields stack as AND (separate pipeline stages).
+function buildEffectiveQuery(baseQuery: string, filters: ActiveFilter[]): string {
```
matyax: It was unexpected to see a LogQL function here. Given that the implementation has been dettached between querying and displaying, exposing props like `fetchAttributes` and `fetchDistribution`, I would also expect this to come from what is currently called `ErrorAnalysis`

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+  if (!filters.length) {
+    return baseQuery;
+  }
+
```
matyax: This should be already implemented in https://github.com/grafana/logs-drilldown/blob/main/src/services/ExpressionBuilder.ts

You can find examples in https://github.com/grafana/logs-drilldown/blob/main/src/services/ExpressionBuilder.test.ts

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+}
+
+export interface AttributeDistributionProps {
+  context: DatasetContext;
+  // Adapter function -- provided by the consumer. Returns detected fields for
+  // the given context. The consumer is responsible for filtering, label mapping,
+  // and any dataset-specific field discovery logic.
+  fetchAttributes: (context: DatasetContext) => Promise<AttributeConfig[]>;
+  // Adapter function -- provided by the consumer. Returns value counts for a
+  // single field within the dataset described by context.
```
matyax:
```suggestion
  // Returns detected fields for the given context. The consumer is responsible for filtering, label mapping,
  // and any dataset-specific field discovery logic.
  fetchAttributes: (context: DatasetContext) => Promise<AttributeConfig[]>;
  // Returns value counts for a single attribute within the dataset described by context.
```

There's a lot of redundancy and bloat in these comments.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+            const values = await fetchDistribution(ctx, attr.field);
+            dispatch({ type: 'LOADED', field: attr.field, values });
+          } catch {
+            dispatch({ type: 'ERROR', field: attr.field });
+          }
+        })
+      );
+    },
+    // fetchDistribution is a stable module-level function passed from the adapter.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
matyax:
```suggestion
```

This should be an exception, and unless some very specific cases, this is probably hiding a bug or a non-ideal implementation. It also hard to know what this is hiding, as there is no explanation of why was it disabled.

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,959 @@
...
+          } catch {
+            dispatch({ type: 'ERROR', field: attr.field });
+          }
+        })
+      );
+    },
+    // fetchDistribution is a stable module-level function passed from the adapter.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
```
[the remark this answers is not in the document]
matyax: Yep.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
matyax: This would also delegate escaping values and other common source of issues, which should be handled there.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+      }
+      if (cancelled) {
+        return;
+      }
+      const ordered = orderByPriority(detected, priorityAttributes);
+      dispatch({ type: 'SET_ATTRIBUTES', configs: ordered });
+      const activeFilters = selectedFiltersRef.current;
+      const effectiveContext =
+        activeFilters.length > 0
+          ? { ...context, query: buildEffectiveQuery(context.query, activeFilters) }
```
matyax: To reinstate a comment above. buildEffectiveQuery is eventually passed back to fetchDistribution() which is already in ErrorAnalysis.tsx, so fetchDistribution() should be the one that builds the effective query.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+    }
+
+    let cancelled = false;
+
+    async function run() {
+      dispatch({ type: 'DETECTING' });
+      let detected: AttributeConfig[] = [];
+      try {
+        detected = await fetchAttributes(context);
+      } catch {
```
matyax:
```suggestion
      } catch (e) {
        logger.error(e);
```

Otherwise this will silently swallow any error.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+    }
+
+    run();
+
+    return () => {
+      cancelled = true;
+    };
+    // fetchAttributes is a stable module-level function; context fields and
+    // priorityAttributes are the true dependencies.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
matyax:
```suggestion
```

Again I would very appreciate not disabling exhausting dependencies unless it's strictly necessary.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,959 @@
...
+      return;
+    }
+    // Use the raw field name as the label -- the consumer's label mapping only
+    // applies to the detected field list returned by fetchAttributes.
+    const config: AttributeConfig = { field, label: field };
+    dispatch({ type: 'ADD_ATTRIBUTE', config });
+    setNewFieldInput('');
+    const effectiveContext = { ...context, query: buildEffectiveQuery(context.query, state.selectedFilters) };
+    loadDistributions([config], effectiveContext);
+  }
```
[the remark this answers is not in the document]
matyax: What is the purpose of allowing the user to add any arbitrary attribute?

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+  const nonPriorityAttributes = state.attributes.filter((a) => !priorityFieldSet.has(a.field));
+  // Only split if there are both priority and non-priority fields. If no priority
+  // attributes are configured, all fields are shown without a button.
+  const hasSplit = priorityFieldSet.size > 0 && nonPriorityAttributes.length > 0;
+  const priorityAttributes_ = hasSplit ? state.attributes.filter((a) => priorityFieldSet.has(a.field)) : state.attributes;
+  const visibleNonPriority = nonPriorityAttributes.slice(0, extraFieldsShown);
+  const visibleAttributes = hasSplit ? [...priorityAttributes_, ...visibleNonPriority] : state.attributes;
+  const remainingCount = nonPriorityAttributes.length - extraFieldsShown;
+  const nextBatch = Math.min(10, remainingCount);
+
```
matyax, line 309: All of this will be re-executed on every render and cause a cascade of re-renders in the whole component tree.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+      <div className={styles.header}>
+        <div className={styles.title}>{t('errors-analysis.title', 'Attribute Explorer')}</div>
+        {queryLimitLabel && <div className={styles.queryLimit}>{queryLimitLabel}</div>}
+        <div className={styles.description}>
+          {t(
+            'errors-analysis.description',
+            'Spot patterns and narrow down root causes by exploring how your data breaks down across key attributes. Click any value to filter your results.'
+          )}
+        </div>
+      </div>
```
matyax:
```jsx
      <div className={styles.header}>
        <div className={styles.title}>{t('errors-analysis.title', 'Attribute Explorer')}</div>
        {queryLimitLabel && <div className={styles.queryLimit}>{queryLimitLabel}</div>}
        <div className={styles.description}>
          {t(
            'errors-analysis.description',
            'Spot patterns and narrow down root causes by exploring how your data breaks down across key attributes. Click any value to filter your results.'
          )}
        </div>
      </div>
```

It'd be nice to make this header an optionally passed prop from the consumer, to disable, or support any custom header for the particular implementation.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+        />
+        <button className={styles.addButton} type="button" onClick={handleAddField}>
+          {t('errors-analysis.add-field-button', '+ Add')}
+        </button>
+      </div>
+
+      {state.detecting && (
+        <div className={styles.detectingRow}>
+          <Spinner size="sm" />
+          <span>{t('errors-analysis.discovering-fields', 'Discovering fields\u2026')}</span>
```
matyax:
```suggestion
          <span>{t('errors-analysis.discovering-attributes', 'Discovering attributes\u2026')}</span>
```

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,949 @@
...
+        {isExpandable && <Icon name={expanded ? 'angle-up' : 'angle-down'} size="sm" />}
+      </button>
+
+      {loading && (
+        <div className={styles.loadingRow}>
+          <Spinner size="sm" />
+        </div>
+      )}
+
+      {!loading && error && <div className={styles.emptyRow}>{t('errors-analysis.no-data', 'No data')}</div>}
```
matyax: If it's an actual error, we should let the user know.

```suggestion
      {!loading && error && <div className={styles.emptyRow}>{t('errors-analysis.error', 'Failed to load')}</div>}
```

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
```diff
@@ -0,0 +1,188 @@
...
+  // list because it has domain knowledge of which fields are noise for its dataset.
+  // If not provided, all detected fields are shown.
+  fieldsToExclude?: string[];
+  // See AttributeDistributionProps.initialSelectedFilters.
+  initialSelectedFilters?: Array<{ field: string; operator: '!=' | '='; value: string }>;
+  // Display name overrides for raw field names. The consuming app owns this mapping
+  // because it knows what its fields mean to users.
+  // Unknown fields fall back to their raw field name.
+  // If not provided, all fields display with their raw name.
+  labelMap?: Record<string, string>;
```
matyax:
```suggestion
  attributeMap?: Record<string, string>;
```

I would suggest to move this to AttributeDistribution, and using it at visualization time. So instead of creating a new data structure with `attribute` and `atrtribute_name` (current field, and label), you just check if there is a value in the map, or fall back to the attribute. Like:

```
<div className="attribute">{labelMap?.[attribute] ?? attribute}</div>
```

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
```diff
@@ -0,0 +1,188 @@
...
+  // boundaries (different webpack bundles) causes cross-realm prototype errors.
+  const excludeSet = new Set(fieldsToExclude);
+
+  return async function fetchAttributes(context: DatasetContext): Promise<AttributeConfig[]> {
+    const ds = (await getDataSourceSrv().get(context.datasourceUid)) as LokiDatasource;
+
+    const start = dateTime(context.timeRange.from).utc().toISOString();
+    const end = dateTime(context.timeRange.to).utc().toISOString();
+
+    const response = (await (ds as any).getResource(
```
matyax: I'd appreciate it if `any` and type assertions are avoided as much as possible, otherwise using TypeScript becomes meaningless.

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
```diff
@@ -0,0 +1,188 @@
...
+      'detected_fields',
+      { end, query: context.query, start },
+      { requestId: 'errors-detected-fields' }
+    )) as { fields?: Array<{ label: string }> };
```
matyax: In Logs Drilldown we use narrowing functions not only to avoid type assertions, but also to guarantee that we're working with the type of data that we expect. 

https://github.com/grafana/logs-drilldown/blob/main/src/services/narrowing.ts

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (!labelField || !valueField) {
+      return;
+    }
+
+    function getValue(values: unknown, i: number): unknown {
+      if (typeof (values as any)?.get === 'function') {
+        return (values as any).get(i);
+      }
+      return (values as any)[i];
+    }
```
matyax: Given that above you know that:

```tsx
const labelField = frame.fields.find((f) => f.type === FieldType.string && f.name === field);
const valueField = frame.fields.find((f) => f.type === FieldType.number);
```

Then you know you'll be dealing with either strings or numbers.

```suggestion
    function getValue(values: string[] | number[], i: number): string | number {
      // Not really sure what is this for, I haven't seen this before
      if ('get' in values) {
        return values.get(i);
      }
      return values[i];
    }
```

Or even better:

```suggestion
    function getStringValue(values: string[], i: number): string {
      return String(values[i]);
    }
    function getNumberValue(values: string[], i: number): string {
      return String(values[i]);
    }
     // Later:    
     const labelValue = getStringValue(labelField.values, i);
     const count = getNumberValue(valueField.values, i);
```

matyax (commented):
Adding a few comments and suggestions on the implemenation. There wasn't a simple way of testing this component in Logs Drilldown, so I haven't tried it yet, but I would like to.

In general, I would appreciate it if:
- We split the code in a couple of modules
- Rename ErrorAnalysis
- Leave all data and query related code in what is currently called ErrorAnalysis
- Minimize `any`, type assertions, and exhaustive dependency disables
- Add some unit tests

Thank you so much for working on this, I love this idea!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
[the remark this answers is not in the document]
bohandley: Investigating this! This was an ask in the design stage 🤔 @pklam @codecapitano

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Will do!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Yep! These are just attributes.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: I'll update this!

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
bohandley: Good catch! I'll remove the `any`

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
bohandley: Nice! I'll implement this

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
bohandley: Cool! I'll update this

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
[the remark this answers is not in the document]
pklam: Hi there! So right now on first view, this component shows a subset of default, important labels like Version or Browser. Teams will also have their own custom labels (like: team_label or for some internal teams, stackId) that they want to always see to contextualize their errors.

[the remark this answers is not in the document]
matyax: Ohh, I see. If we expect those labels to come from detected_fields (otherwise they wouldn't exist?), can we use a Select/ComboBox component instead? As arbitrary user input shouldn't be allowed.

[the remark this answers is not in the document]
bohandley: Cool! I am reading this as we have two controls serve different navigation patterns for the same data to show detected fields:

- "Show more" is for browsing (show the next 10 fields)
- "Add field" is for when you know what you're looking for
  - when a user adds a field, they select from hidden fields and this is shown, or sorted up so it can be seen.

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Updated!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Bloat reduced!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: fixed!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Updated!

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
bohandley: Renamed!

src/Components/AttributeDistribution/AttributeDistribution.tsx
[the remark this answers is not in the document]
bohandley: Fixed!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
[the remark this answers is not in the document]
bohandley: Updated with functionality: user can click to see hiden fields, selects one and it is shown in the component.
Screenshot 2026-04-15 at 6 56 12 PM

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Updated

src/Components/ErrorsAnalysis/ErrorsAnalysis.tsx
bohandley: Done!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Updated!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: Logged!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: un-disabled!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
bohandley: updated!

src/Components/AttributeDistribution/AttributeDistribution.tsx
bohandley: memoized!

src/Components/AttributeDistribution/AttributeDistribution.tsx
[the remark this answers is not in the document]
bohandley: Not sure why this is not shown as outdated, but it's fixed.

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,203 @@
+import React, { useMemo } from 'react';
```
matyax, line 1: Super ultra minor issue, but the main folder is still called `ErrorAnalysis`. I think renaming to `AttributeDistribution` should work 👍

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,203 @@
...
+  fieldsToExclude = [],
+  initialSelectedFilters,
+  attributeMap = {},
+  onFiltersChange,
+  priorityAttributes,
+  query,
+  queryLimitLabel,
+  showAllLink,
+  timeRange,
+}: LokiFieldDistributionProps) {
```
matyax, line 194: Lovely <3

src/Components/ErrorsAnalysis/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,203 @@
...
+  queryLimitLabel,
+  showAllLink,
+  timeRange,
+}: LokiFieldDistributionProps) {
+  const fetchAttributes = useMemo(
+    () => makeFetchAttributes(fieldsToExclude, attributeMap),
+    [fieldsToExclude, attributeMap]
+  );
+
+  if (!getFeatureFlag('drilldown.logs.attributeExplorer')) {
```
matyax: Do you think the feature flag is really needed? I don't know from Frontend o11y's perspective, but from the Logs POV it doesn't seem be necessary.

matyax:
Thank you for addressing the comments! It's looking great.

I tried to test it, but it got into a loop of cancelled and re-executed requests.

https://github.com/user-attachments/assets/6f07cf40-c352-4250-a065-399e220fc585

I tried to test it in a hacky way, with:

```
<LokiFieldDistribution datasourceUid={'ferb7bt38kykge'} query={'{service_name="mimir/ingeter"}'} timeRange={sceneGraph.getTimeRange(model).state.value} />
```

Let me know if there's anything incorrect on my end and I'll try again.

src/Components/ErrorsAnalysis/LokiFieldDistribution.tsx
bohandley: I'm happy to remove the feature flag!

src/Components/ErrorsAnalysis/AttributeDistribution.tsx
```diff
@@ -0,0 +1,706 @@
...
+      nonPriorityAttributes: nonPriority,
+      priorityAndPinned: pinned,
+    };
+  }, [priorityAttributes, state.attributes, state.userPinnedAttributes]);
+
+  const { visibleAttributes, remainingCount, nextBatch } = useMemo(() => {
+    const visibleNonPriority = nonPriorityAttributes.slice(0, extraFieldsShown);
+    const remaining = nonPriorityAttributes.length - extraFieldsShown;
+    return {
+      nextBatch: Math.min(10, remaining),
```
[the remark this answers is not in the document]
bohandley: Addressed 👍🏻

src/Components/ErrorsAnalysis/attributeDistributionState.ts
```diff
@@ -0,0 +1,200 @@
...
+      if (newFilters.length === 0) {
+        valueSnapshot = null;
+      }
+
+      return { ...state, selectedFilters: newFilters, valueSnapshot };
+    }
+    case 'CLEAR_FILTERS':
+      return { ...state, selectedFilters: [], valueSnapshot: null };
+    case 'SET_FILTERS':
+      return { ...state, selectedFilters: action.filters, valueSnapshot: action.filters.length === 0 ? null : state.valueSnapshot };
```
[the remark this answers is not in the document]
bohandley: Addressed 👍🏻

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,706 @@
...
+      )}
+
+      {comboboxOptions.length > 0 && (
+        <Combobox
+          options={comboboxOptions}
+          value={null}
+          placeholder={t('errors-analysis.field-input-placeholder', 'Search to add more attributes')}
+          onChange={(option) => option && handlePinAttribute(option.value)}
+        />
+      )}
```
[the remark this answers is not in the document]
bohandley: Addressed 👍🏻, PR description updated.

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
bohandley: Renamed!

src/Components/ErrorsAnalysis/LokiFieldDistribution.tsx
bohandley: Removed!

bohandley:
> Thank you for addressing the comments! It's looking great.
> 
> I tried to test it, but it got into a loop of cancelled and re-executed requests.
> 
>  Test.mov 
> I tried to test it in a hacky way, with:
> 
> Let me know if there's anything incorrect on my end and I'll try again.

@matyax Thank you for the feedback! Your test was good, but I had some bugs with the default props and the timeRange. Here's more details and setup suggestions

something like this should get it working!

```typescript
const { value: datasourceUid } = getDataSourceVariable(model).useState();
const query = sceneGraph.interpolate(model, LOG_STREAM_SELECTOR_EXPR);
const timeRange = sceneGraph.getTimeRange(model).state.value;

<LokiFieldDistribution
  datasourceUid={String(datasourceUid)}
  query={query}
  timeRange={timeRange}
/>
```

With this I was able to see high cardinality in the component!

Screenshot 2026-04-16 at 2 58 42 PM


https://github.com/user-attachments/assets/f61afaea-746d-4697-a606-4242ae81cbd9

matyax:
Amazing progress! @pklam and @bohandley , some thoughts after testing the component:

- Changing the time range should reset the component and cancel in-flight requests.

- This is how we identify fields with values that failed to load. We should let the user know why they can't see values (too many to load).

imagen

Also, if I remember correctly, by setting the supporting query type to Logs Drilldown, it will return partial results.

imagen

But I'm not sure we should use it here because the cardinality can be extreme (20k+).

- With many values, it gets very challenging to discriminate between attribute names and values, that is, what is an attribute and what is a value.

imagen

A little bit more padding could help:

imagen

And also a some extra space between the last value and the line. It's certainly too close right now:

Before:

imagen

After:

imagen

- With queries like `{service_name="grafana/grafana"}`, last 15 minutes, there's a lot of missing values and I'm not sure why. For some, it might depend on having parsers (the parser for a given field is available in the detected_fields response) in the metric query; for others max_series_reached error. For example `{service_name="grafana/grafana"} | logfmt | json | drop __error__, __error_details__ `

imagen

- In this demo it can be seen a bit of what has been mentioned before + some extra details. Reloading (refresh, timerange change) should cancel requests and reset state to loading again. Additionally, the component is firing (expensive) requests for fields that are not visible. For example, "msg" is the first attribute, but it's loading for "orgId" which is not visible. It should definitely be lazy and only request for what is being displayed, otherwise it will waste a lot of resources and consumption for things that will potentially not be displayed. But beyond that, it delays by a lot the loading time of some of the fields that are visible.

https://github.com/user-attachments/assets/4c9d68fa-c2ed-45b5-9285-03f492588b79

- Regarding field ordering, does Frontend O11Y have any requirement or suggestion? E.g. alphabetical, cardinality (tricky, might prioritize extreme-cardinality fields), type of field (parsed or metadata), etc.

- I'm not super convinced about keeping the active filters in the top of the component. Not only it needs to be synced with the header filters (if the user removes them there, the AttributeDistribution component should update), but it also has absolutely no effect in the AttributeDistribution component.

Additionally, once a filter is applied the user can't continue seeing the other values? Or filter by more than one value on a given field?

imagen

I think it would be better to highlight the selected values and fields, not show active filters on top, and also allow to filter by more than one value.

- If you like the idea, it would be trivial to send the user to Logs Drilldown to see the full distribution of values for a given field.

bohandley:
@matyax thank you for the detailed feedback! 
Fixed:
  - Time range changes now cancel in-flight distribution requests and reset all fields to loading
  - Fields that fail to load now show the datasource error message inline (e.g. "maximum number of series (20000) reached" truncated message but full message in tooltip hover)
  - Padding increased between values and section dividers to better distinguish attribute names from values
  - Lazy loading implemented via IntersectionObserver. Only fields visible in the viewport trigger distribution fetches
  - Filter chip strip removed; selected values are highlighted inline
  - Per-field link to Logs Drilldown added as an interactive tooltip on each field name

On field ordering: [the component has a priorityAttributes prop](https://github.com/grafana/app-o11y-kwl/pull/3108/changes#r3105310795) that lets the consumer define which fields appear first and in what order. For Faro errors we pin Browser, Version, View/Page, OS/Device, and Location to the top. Additional detected fields appear below in Loki's returned order, giving the app team full control over ordering without the component needing domain knowledge.

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,812 @@
...
+    } else if (existingForField && existingForField.operator !== operator) {
+      newFilters = [...state.selectedFilters.filter((f) => f.field !== field), { field, value, operator }];
+    } else {
+      newFilters = [...state.selectedFilters, { field, value, operator }];
+    }
+
+    dispatch({ type: 'TOGGLE_FILTER', field, value, operator });
+
+    loadDistributions(visibleAttributes, context, newFilters);
+    onFiltersChange?.(newFilters);
```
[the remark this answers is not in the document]
bohandley: Copilot's solution has a real bug on the SET_FILTERS path. Trying another solution

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const ds = (await getDataSourceSrv().get(context.datasourceUid)) as LokiDatasource;
+
+    const start = dateTime(context.timeRange.from).utc().toISOString();
+    const end = dateTime(context.timeRange.to).utc().toISOString();
+
+    const response = await (ds as unknown as LokiLike).getResource(
+      'detected_fields',
+      { end, query: context.query, start },
+      { requestId: 'errors-detected-fields' }
+    );
```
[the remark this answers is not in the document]
bohandley: not removing this, just making it more specific. Willthere ever be two of these on the same page, probably not, but the request id needs to stay for cancelling requests.

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,236 @@
...
+  // Display name overrides for raw field names. Unknown fields fall back to their raw name.
+  attributeMap?: Record<string, string>;
+  datasourceUid: string;
+  // Fields excluded from the distribution sidebar.
+  fieldsToExclude?: string[];
+  onFiltersChange?: (filters: Array<{ field: string; operator: '!=' | '='; value: string }>) => void;
+  // Attributes pinned to the top of the list.
+  priorityAttributes?: AttributeConfig[];
+  // The full Loki log query for this error group.
+  query: string;
```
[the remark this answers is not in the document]
bohandley: Passing a string means the consumer cannot rename fields. This breaks the consumer contract and forces them into using attribute map.

src/Components/AttributeDistribution/attributeDistributionState.ts
```diff
@@ -0,0 +1,230 @@
...
+    return currentFilters.filter((_, i) => i !== existingIndex);
+  } else if (existingIndex >= 0) {
+    // Operator switch for this value: replace in place
+    return currentFilters.map((f, i) => (i === existingIndex ? { ...f, operator } : f));
+  } else if (existingForField && existingForField.operator !== operator) {
+    // Different operator already active for this field: clear and add new
+    return [...currentFilters.filter((f) => f.field !== field), { field, value, operator }];
+  } else {
+    return [...currentFilters, { field, value, operator }];
+  }
```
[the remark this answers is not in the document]
bohandley: Implementing this

src/Components/AttributeDistribution/attributeDistributionState.ts
```diff
@@ -0,0 +1,230 @@
...
+    case 'DETECTING': {
+      const resetData: Record<string, AttributeState> = {};
+      for (const [field, attrState] of Object.entries(state.data)) {
+        resetData[field] = { error: false, expanded: attrState.expanded, loading: true, values: [] };
+      }
+      return { ...state, data: resetData, detecting: true, valueSnapshot: null };
+    }
+    case 'SET_ATTRIBUTES': {
+      const detectedFields = new Set(action.configs.map((c) => c.attribute));
+      const userAdded = state.attributes.filter((a) => !detectedFields.has(a.attribute));
```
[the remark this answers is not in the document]
bohandley: Priority attributes are a const passed in so this is not an issue

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,802 @@
...
+
+
+      {comboboxOptions.length > 0 && (
+        <Combobox
+          options={comboboxOptions}
+          value={null}
+          placeholder={t('errors-analysis.field-input-placeholder', 'Search to add more attributes')}
+          onChange={(option) => option && handlePinAttribute(option.value)}
+        />
+      )}
```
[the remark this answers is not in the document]
bohandley: We are not creating a custom value for this. It is only a search.

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,237 @@
...
+    return [];
+  }
+  return response['fields'].filter((f): f is { label: string } => isRecord(f) && typeof f['label'] === 'string');
+}
+
+// Appends a per-field metric aggregation around the base log query.
+// The base query must already include logfmt and any hash filters so that
+// [$__range] counts only events in this error group.
+function buildDistributionQuery(baseQuery: string, field: string): string {
+  return `sum by (${field}) (count_over_time(${baseQuery} | keep ${field} [$__range]))`;
```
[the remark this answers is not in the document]
bohandley: This is too strict. Query comes from consumers.

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,802 @@
...
+                      isExcluded && styles.valueRowExcluded,
+                      item.retained && styles.valueRowRetained
+                    )}
+                    onClick={openMenu}
+                    onKeyDown={(e) => {
+                      if (e.key === 'Enter' || e.key === ' ') {
+                        e.preventDefault();
+                        e.currentTarget.click();
+                      }
+                    }}
```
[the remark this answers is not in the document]
bohandley: Fixing PR description

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,802 @@
...
+        <div className={styles.detectingRow}>
+          <Spinner size="sm" />
+          <span>{t('errors-analysis.discovering-fields', 'Discovering attributes\u2026')}</span>
+        </div>
+      )}
+
+      {!state.detecting && state.attributes.length === 0 && (
+        <div className={styles.emptyState}>{t('errors-analysis.no-fields-detected', 'No fields detected for this error group.')}</div>
+      )}
+
```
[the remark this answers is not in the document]
bohandley: Fixing PR description

src/services/extensions/exposedComponents.tsx
```diff
@@ -4,6 +4,7 @@ import { Trans } from '@grafana/i18n';
 import { LinkButton } from '@grafana/ui';
 
 import pluginJson from '../../plugin.json';
+import { LokiFieldDistributionProps } from 'Components/AttributeDistribution/LokiFieldDistribution';
 import { EmbeddedLogsExplorationProps } from 'Components/EmbeddedLogsExploration/types';
 import { OpenInLogsDrilldownButtonProps } from 'Components/OpenInLogsDrilldownButton/types';
```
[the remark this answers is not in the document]
bohandley: Using `type`

src/Components/AttributeDistribution/attributeDistributionState.ts
```diff
@@ -0,0 +1,230 @@
...
+      const merged = [...action.configs, ...userAdded];
+      const data: Record<string, AttributeState> = {};
+      for (const c of merged) {
+        data[c.attribute] = state.data[c.attribute] ?? { error: false, expanded: false, loading: true, values: [] };
+      }
+      return { ...state, attributes: merged, data, detecting: false };
```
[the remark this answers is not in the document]
bohandley: This is only half right. This would drop userPinnedAttributes, from the search box.Coming up with a better fix.

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,237 @@
...
+        to: dateTime(context.timeRange.to).utc().toISOString(),
+      },
+      to: dateTime(context.timeRange.to),
+    },
+    requestId: `errors-breakdown-${field}`,
+    scopedVars: {},
+    startTime: Date.now(),
+    targets: [target],
+    timezone: 'browser',
+  };
```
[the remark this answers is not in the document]
bohandley: Need to keep request Id but make it more specific.

matyax:
Could you please show a custom message for this case?

imagen

The reason is that people don't know what "max series" is. Something around "Too many values to display" should be ok.

Love the "open in Drilldown" link, btw!

imagen

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,823 @@
...
+        <button
+          className={cx(styles.sectionHeader, hasActiveFilter && styles.sectionHeaderActive)}
+          type="button"
+          onClick={isExpandable ? onToggle : undefined}
+        >
+          <span className={styles.sectionLabel}>{config.attribute_name}</span>
+        </button>
+        {fieldLink && (
+          <a
+            aria-label={t('errors-analysis.field-link-label', 'View {{name}} in Logs Drilldown field tab', { name: config.attribute_name })}
```
matyax, line 490:
```suggestion
            aria-label={t('errors-analysis.field-link-label', 'View {{name}} in Logs Drilldown field tab', { name: config.attribute_name })}
            title={t('errors-analysis.field-link-label', 'View {{name}} in Logs Drilldown field tab', { name: config.attribute_name })}
```

A title would make this visible when hovering the link, or alternatively a tooltip.

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,823 @@
...
+    }
+
+    run();
+
+    return () => {
+      cancelled = true;
+      subscriptionsRef.current.forEach((sub) => sub.unsubscribe());
+      subscriptionsRef.current = [];
+    };
+  }, [context.query, context.datasourceUid, context.timeRange.from, context.timeRange.to, loadDistributions]);
```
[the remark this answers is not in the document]
bohandley: Priority attributes does not change order. Added fields are only appended. Disregarding this.

src/Components/AttributeDistribution/AttributeDistribution.tsx
```diff
@@ -0,0 +1,823 @@
...
+  userPinnedRef.current = state.userPinnedAttributes;
+  // Read via refs so unstable consumer references don't trigger re-detection.
+  // Only genuine dataset changes (query, datasource, time range) should re-detect.
+  const attributeLabelsRef = useRef(attributeLabels);
+  attributeLabelsRef.current = attributeLabels;
+  const priorityAttributesRef = useRef(priorityAttributes);
+  priorityAttributesRef.current = priorityAttributes;
+  const fetchAttributesRef = useRef(fetchAttributes);
+  fetchAttributesRef.current = fetchAttributes;
+
```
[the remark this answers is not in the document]
bohandley: attribteLabels is static in practice so won't be updated. This would be good for a future refactor but I am rejecting this suggestion.

src/Components/AttributeDistribution/LokiFieldDistribution.tsx
```diff
@@ -0,0 +1,237 @@
...
+      return String(values[i]);
+    }
+
+    function getNumberValue(values: number[], i: number): number {
+      return Number(values[i]);
+    }
+
+    for (let i = 0; i < frame.length; i++) {
+      const labelValue = getStringValue(labelField.values as string[], i);
+      const count = getNumberValue(valueField.values as number[], i);
```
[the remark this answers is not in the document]
bohandley: False positive. This codebase uses @grafana/data 13.0.0. The type definition at node_modules/@grafana/data/dist/types/types/dataFrame.d.ts confirms: values: T[]. Bracket access is correct and the cast labelField.values as string[] is safe.

matyax (approved):
Thanks for iterating over this! I look forward to use this in Logs Drilldown soon!

PR #1831, opened 2026-04-06
fix(insights): use namespaced goff flag for KG annotations

rdubrock:
## Summary
- The `kgAnnotationsInLokiExplore` feature flag was registered without the `drilldown.logs.` namespace prefix required by goff, so the flag never resolved to `true` even when enabled in the goff backend
- Renamed all references from `kgAnnotationsInLokiExplore` to `drilldown.logs.kgAnnotationsInLokiExplore` (flag definition, availability check, toolbar guard)
- The `config.featureToggles` fallback still checks the un-namespaced key since that's the Grafana-side toggle name

## Test plan
- [ ] Enable `drilldown.logs.kgAnnotationsInLokiExplore` in goff and verify the Insights toggle appears on service detail pages (logs, labels, patterns, fields)
- [ ] Verify the toggle does not appear on the explore/service selection page
- [ ] Verify annotations render on the timeseries panel when the toggle is enabled


rdubrock:
@L2D2Grafana nice, the main updates fixed it. I don't have permissions to merge, so if it looks good then merge away!

PR #1844, opened 2026-04-09
fix(Feature Flags): remove logsPanelControls

matyax:
Part of https://github.com/grafana/grafana/issues/117687

Since we're still compatible with 11.6, we'll continue using the version check, without looking at the feature toggle state.


L2D2Grafana (approved):
🧹

PR #1859, opened 2026-04-20
build: migrate to pnpm

nicwestvold:
### ✨ Description

Switches the package manager from Yarn 4 to pnpm. No runtime code changes — just tooling, lockfile, and CI updates.

### 📖 Summary of the changes

- Replaced `yarn.lock` with `pnpm-lock.yaml` and added `pnpm-workspace.yaml`.
- Updated `package.json` scripts and `packageManager` field for pnpm.
- Removed `.yarn/releases/` and `.yarnrc.yml`.
- Updated `.gitignore` for pnpm.
- CI/CD workflows (`ci.yml`, `cd.yml`, `is-compatible.yml`) now install and run with pnpm.
- Tweaked `lint-staged.config.js`, `README.md`, and provisioning configs where commands were yarn-specific.

### 🧪 How to test?

- `pnpm install` then `pnpm dev` / `pnpm build` / `pnpm test` locally.
- Confirm CI is green on this PR (build, test, compatibility checks).
- Verify the plugin still loads in the provisioned Grafana via `pnpm server`.


.github/workflows/ci.yml
```diff
@@ -34,6 +35,7 @@ jobs:
       # Same settings as publish
       run-playwright: ${{ github.event_name == 'pull_request' }} # Run playwright tests for PRs only
       run-playwright-with-grafana-dependency: '>=11.6 <13.0.0 || >13.0.0'
+      run-playwright-with-skip-grafana-react-19-preview-image: true # Plugin not yet React 19 ready; tracked separately
```
matyax: What is the meaning of this comment?

nicwestvold: 🙂 you got to it before I could adjust. ~~I don't think this whole line should be here.~~ I am fixing now.

I have updated the comment - it was vibed 🙈 

v7 added the React 19 image to the matrix. We are temporarily disabling it, to focus only on the `pnpm` migration.

package.json
```diff
@@ -124,43 +126,40 @@
...
-    "tar": "7.5.11",
-    "js-yaml": "4.1.1",
-    "@grafana/data": "13.0.0",
-    "@grafana/schema": "12.4.2"
+  "pnpm": {
+    "overrides": {
+      "react-router-dom": "^7.12.0",
+      "uuid": "^10.0.0",
+      "@grafana/data": "~13.0.0",
+      "@grafana/schema": "~12.4.2",
```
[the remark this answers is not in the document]
nicwestvold: This is intentional. Having a stale override could block future patches. We should only pin when necessary.

package.json
```diff
@@ -115,6 +116,7 @@
     "@grafana/api-clients": "^13.0.0-22766927077",
     "@grafana/assistant": "0.1.13",
     "@grafana/data": "^13.0.0",
+    "@grafana/faro-web-sdk": "^1.19.0",
     "@grafana/i18n": "^13.0.0",
```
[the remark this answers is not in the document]
nicwestvold: These aren't new runtime deps. pnpm's stricter resolution just surfaced what yarn was hoisting implicitly. Same import graph, now explicit. "No runtime code changes" still holds.

.github/workflows/ci.yml
L2D2Grafana: We were running against the dev image, can we bring this back?

L2D2Grafana:
The PR works well locally running pnpm 👍 I'd be interested to know if the dev image is failing due to library updates.

L2D2Grafana (approved):

.github/workflows/ci.yml
nicwestvold: yep! I will bring this back and see. 

from what I could tell previously, it had to do with react 19 strict mode causing hooks/effects to run twice, which doubled the expectation. We'll know shortly, though 😉. I'll remove this and push.

nicwestvold: ~~Okay, I think I _finally_ got to the bottom on the react 19 test issues.~~

Correction: I got to the bottom of **some** of the test failures.

React 19 causes the `IntersectionObserver` to see more grid items during initial mount. I confirmed this, but haven't identified why React 19 does this. Instead of hardcoding a viewport-dependent count, the test ensures the correct grouping of queries (2x).

tests/exploreServices.spec.ts
```diff
@@ -377,7 +391,8 @@ test.describe('explore services page', () => {
         // Noticed that the below assertions were flaking when not running the trace, we need to wait a tiny bit to let the last requests fire
         await page.waitForTimeout(50);
         expect(logsVolumeCount).toEqual(4);
-        expect(logsQueryCount).toEqual(16);
+        // 1 initial load + 3 refreshes = 4× the initial panel query count
+        expect(logsQueryCount).toEqual(initialQueryCount * 4);
```
nicwestvold, line 395: This test previously hardcoded `logsQueryCount == 4`, but due to how many services were in-view (and thus fired) with React 19, the test was failing - there were 4 instead of 2 services that were visible.

The updated assertion captures what we expect: greater than zero queries fired, with each service producing 2 queries, and refreshes multiply the count correctly.

tests/exploreServices.spec.ts
```diff
@@ -479,8 +496,9 @@ test.describe('explore services page', () => {
           // Dropdown should be open
           await expect(selectNewLabelSelect).toContainText('Search labels');
 
-          // Add "namespace" as a new tab
-          await page.getByText('namespace', { exact: true }).click();
+          // Add "namespace" as a new tab — type to filter the dropdown so the option is in viewport
+          await page.keyboard.type('namespace');
+          await page.getByRole('option', { name: 'namespace', exact: true }).click();
```
nicwestvold, line 501: The "namespace" option was outside the 600px test viewport in CI. Typing into the searchable Select filters the list so the option is visible before clicking. `exact: true` avoids matching "k8s_namespace_name".

tests/exploreServices.spec.ts
```diff
@@ -57,8 +57,18 @@ test.describe('explore services page', () => {
...
+
+      // Clear the search filter — the clear button uses aria-label (old Select) or title (new Combobox)
+      await page.getByRole('button', { name: 'Clear value' }).first().click();
+
+      // Assert there is more than one result now
+      await expect(explorePage.getPanelHeaderLocator().nth(1)).toBeVisible();
+
+      // Assert that the first element is nginx with the Remove filter action
+      await expect(explorePage.getPanelHeaderLocator().first().getByText('nginx', { exact: true })).toBeVisible();
+      await expect(explorePage.getPanelHeaderLocator().first().getByText('Remove', { exact: true })).toBeVisible();
```
nicwestvold, line 71: Previously this section used a conditional based on `textContent()` of the full panel header, which returned different values under React 18 vs 19 due to action button render timing.

.github/workflows/ci.yml
nicwestvold: @L2D2Grafana, there were a few test changes that I made in order to get the e2e test suite to pass on both React 18 and 19. I ran it locally to confirm and am waiting for the CI to run so that I can verify that it _actually_ worked. I mention it to note that those test files should get a very careful look to ensure that I didn't change behavior/intent of what was under test.

L2D2Grafana:
Thanks for turning on the react dev image. It looks like I have an error from my combobox changes. We've decided to turn off the dev image and I will create another PR to fix combobox and turn it back on.

PR #1860, opened 2026-04-20
feat(line filters): move to the top, update tests

L2D2Grafana:
## Summary 📝

- **Line filters moved to the top:** Line filters are edited in the index variable strip (with the other variables) instead of inline above the logs/JSON/table panel; `LineFilterVariablesScene` is hidden until the service drilldown activates, then shown via `ServiceScene.showVariables()` with `LINE_FILTER_VARIABLES_SCENE_KEY`.
- **Unified filter rows:** Single `filters.map` path, optional remove control, “Add line filter” when the last row has text, `ensureAtLeastOneEmptyLineFilter` when visible with no filters, and defaults pulled from stored regex/case/exclusive preferences.

Screenshot 2026-04-17 at 2 33 02 PM

## Test 🧪

- [x] Review `LineFilterVariablesScene` + `LogsListScene` integration: inline `LineFilterScene` removed from logs/JSON/table layouts; line limit row layout updated (`LineLimitScene` sizing).
- [x] Manually verify after selecting a service: line filters appear in the top controls, “Add line filter” appears when the last row has content, remove/add behavior and URL/history updates behave as before.
- [x] **E2E updates:** Submit line filters with Enter where appropriate, assert the logs panel stays closed via the Logs panel header test id, fix regex re-enable assertion (one “Disable regex” control), and adjust expectations for the new layout (e.g. line filter count / migration assertions).


matyax:
- We should remove the (x) from Line Filter now that's shown by default.

imagen

- I thought I made this comment in this PR but I can't find it. We should use the available space instead of continue growing the header, which we know many users find uncomfortable, even with the collapse function.

imagen

- We should either remove or tweak the debounced apply of new line filters, which I forgot they were working this way. Either by introducing an apply button to them, or making sure we're giving users enough time to write before sending a new request. There's also an alert that appears when typing that I don't know where it's coming from.


https://github.com/user-attachments/assets/006d1978-263b-40ac-8711-595f2ad20a72

L2D2Grafana:
- Removed the (x) from the default line filter
- Moved the line filters inline with the fields 
Screenshot 2026-04-24 at 3 47 43 PM

- I tweaked the debounce. It might be too long now. I could explore adding an apply button. It does update on keypress enter. The rest of the query variables do not have apply buttons so it's a bit weird.

- I'm not seeing that error message you posted. I do see this error message, but it's not tied to the line filter
Screenshot 2026-04-24 at 4 23 51 PM

matyax:
Love it! One outstanding issue. When using the popover menu, it adds a new filter after leaving an empty filter before it:

https://github.com/user-attachments/assets/80216caa-876c-4951-8cec-7c7bd76ca910

You can find the code that handles it in the Logs scene.

L2D2Grafana:
Updated to handle the LogsPanelScene add to line filters!

matyax:
Let's hold this one until after the next release, to give it some time to catch any unexpected regression.

In the mean time, could you give it a try to see if we can move the line limit inside the visualization, so we can reclaim the whole row of space where the line filter used to be?

Line limit

L2D2Grafana:
Updated to put the Line Limit in the Logs panel header. 
Screenshot 2026-04-28 at 2 30 29 PM
It's a bit tricky in small, but users can scroll through it
Screenshot 2026-04-28 at 2 30 17 PM

matyax (approved):
👏 👏 👏  amazing!

Let's wait until after the release to "dogfood" for a little bit.

matyax:
A potential follow up for this issue is probably going to be to add an apply button to this filter. I imagine it being applied after typing might be considered non ideal for some users.

PR #1863, opened 2026-04-22
chore(deps): downgrade data and i18n for compatibility

L2D2Grafana:
The compatibility check is failing a lot as the grafana dependencies get republished this week. This is to unblock ci, but I expect it will switch back to requiring 13.0.1. very soon.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,8 +114,8 @@
     "@emotion/css": "11.10.6",
     "@grafana/api-clients": "^13.0.0-22766927077",
     "@grafana/assistant": "0.1.13",
-    "@grafana/data": "^13.0.0",
-    "@grafana/i18n": "^13.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -158,9 +158,7 @@
     "qs": "6.14.2",
     "strip-ansi": "6.0.1",
     "tar": "7.5.11",
-    "js-yaml": "4.1.1",
-    "@grafana/data": "13.0.0",
-    "@grafana/schema": "12.4.2"
```
matyax, line 163: Are we removing schema? Data was moved above, but schema is gone.

L2D2Grafana: This is removing it from the resolutions. It should still be in dependencies

matyax (approved):

PR #1867, opened 2026-04-24
fix(TagProviders): add missing time range to keys and values providers

### LINKED ISSUE #1811: Label filter autocomplete dropdown does not respect the current page time range — services with no recent logs are missing
## Summary

The "Add label filter" combobox in Logs Drilldown fetches label values using a fixed 
internal time window instead of the currently selected page time range. This causes 
label values that haven't appeared in logs recently to be silently missing from the 
autocomplete dropdown, even though they exist within the user's selected time range.

## Steps to Reproduce

1. Set the page time range to a wide window e.g. **Last 7 days**
2. Click **+ Add label filter** and select any label (e.g. `service_name`)
3. Start typing a value in the combobox
4. Observe that some values are missing from the dropdown

## Expected Behavior

The label value autocomplete should use the **same time range as the current page** 
when querying Loki's label values API, so any value visible in the current view is 
also discoverable in the filter dropdown.

## Actual Behavior

The dropdown uses a fixed short time window to fetch label values. Any label value 
with no log activity within that window is silently omitted from the autocomplete — 
with no indication to the user that values are missing.

The only workarounds are:
- Type the full value manually and select **"Use custom value: ..."**
- Switch to Grafana Explore, which respects the page time range for label autocomplete

## Why This Matters

This is particularly confusing during incident investigation — a service that has 
crashed and stopped logging is exactly when you need to filter for it, but it 
disappears from the dropdown. It also affects low-volume or batch services that 
log infrequently.

## Proposed Fix

Use the current page `timeRange` when calling Loki's label values API for the filter 
combobox. If performance is a concern for large time ranges, cap the autocomplete 
window at the page time range so it never queries more than needed, while ensuring 
it never queries less than what the user is currently viewing.

matyax:
Fixes https://github.com/grafana/logs-drilldown/issues/1811

It turns out, getTagKeys and getTagValues were using the default time range instead of the currently selected one.

https://github.com/user-attachments/assets/3cc74442-0ecc-40ae-a4da-74ca6bb17e2f


L2D2Grafana (approved):
🦋 Nice fix, seeing it update the time for the labels and values call for last 7 days

Image

PR #1869, opened 2026-04-24
fix(combobox): open numeric popover react19, run against react19 dev image

L2D2Grafana:
## Summary 📝

- **`NumericFilterPopoverScene`:** Tighten the document `click` listener that closes the parent popover so clicks on the breakdown **filter button group** (`data-testid` …`filter-button-group`) are ignored—avoids the popover closing as soon as you use the **Add to filter** trigger after the Combobox portal work. Cleanup now removes the same handler reference reliably.
- **`lint-staged`:** Extend the cspell glob negation with **`.github/**/*`** so workflow YAML under `.github` is not spellchecked on pre-commit (fewer false positives on pins, comments, etc.).
- **`.github/workflows/ci.yml`:** Simplify shared CI inputs for the Playwright job—**`run-playwright-with-grafana-dependency: '>=11.6'`** and drop the prior skip / narrow range / React 19 preview image flags so the matrix matches the intended **React 19 dev** coverage from the branch commits.

## Test 🧪

- [x] Manually on a PR build: **Fields breakdown** → **Add to filter** → numeric popover stays open when using the header controls and Comboboxes; dismiss still works via outside click / Cancel as before
- [x] Confirm **CI workflow** reflects the new `run-playwright-with-grafana-dependency` value and Playwright still runs for PRs as expected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,9 +34,7 @@ jobs:
 
       # Same settings as publish
       run-playwright: ${{ github.event_name == 'pull_request' }} # Run playwright tests for PRs only
-      run-playwright-with-grafana-dependency: '>=11.6 <13.0.0 || >13.0.0'
-      # v7 added the React 19 preview image to the matrix; v4 doesn't have it
-      run-playwright-with-skip-grafana-react-19-preview-image: true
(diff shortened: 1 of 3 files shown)


matyax (approved):
Tested the main comboboxes and couldn't find any regression.

PR #1881, opened 2026-05-06
chore(e2e): playwright use mocked data

L2D2Grafana:
## Description ✨ 
This is a WIP. I'm putting it up to show direction and try it against the real ci. e2e tests running about 4min in ci vs the old 13min

## Summary 📝

 - Closes #1836 ([CHORE]: e2e tests mock api calls). Replaces every Loki API call in the e2e suite with static, per-service fixtures so tests run offline against a Grafana-only Docker stack (`docker-compose.playwright.yaml`) instead of requiring `gdev-loki`.
- Layered mocking model: `mockExploreApi(page)` registers default `_global` + `tempo-ingester` routes; per-test scenario loaders under `tests/mocks/scenarios/` shadow the defaults via `page.route()` (with `route.fallback()` for layered handlers). All navigation URLs use a fixed snapshot time (`tests/mocks/snapshotTime.ts`) so frames always sit inside the active range and nothing depends on `Date.now()`.
- Recorder pipeline: `pnpm mocks:record` captures live Loki responses via `tests/recordExploreMocks.spec.ts` → `pnpm mocks:split` shards them into `tests/mocks/labels/<group>-<name>/`. Captured raw output is gitignored (single `ds_query.json` exceeds GitHub's 100 MB limit); only the per-service slices are committed. Full architecture docs in `tests/README.md` and `tests/mocks/README.md`.

## Test 🧪

- [x] `pnpm e2e`
- [x] `pnpm e2e:matrix`
- [x] `pnpm e2e:fast`

> [!NOTE]
> 13 tests are skipped via `test.fixme` and tagged with inline comments explaining the follow-up Loki query whose response wasn't captured. Each can be re-enabled by adding a dedicated dynamic scenario under `tests/mocks/scenarios/` (see `loadBreakdownTempoDistributorWithLevelInfo.ts` for the canonical pattern).


L2D2Grafana:
Going with the snapshot version.

PR #1892, opened 2026-05-11
fix(LogsPanelScene): set timestamp resolution from viz options

matyax:
Setting a missing option.

### How to test

Set the timestamp resolution to ns, or ms. After refreshing, your selection must be respected.


L2D2Grafana (approved):
👍

PR #1752, opened 2026-02-18
feat(LogsTablePanel): Add support for the Logs Table native visualization

gtk-grafana:
This PR adds support to use the new Logs Table visualization.

Requires the `logsTablePanelNG` feature flag.

### What to test

- Displayed fields persistence
- Backend configured displayed fields
- Add filters from the table
- Sort order updates
- Wrapping controls

PR #1893, opened 2026-05-12
fix(LevelsVariableScene): fix negative values in queries and the displayed value in the combobox

matyax:
Fixed:
- Produced queries: negative level queries were being sent as `level = "!level"` instead of `level != "level"`
- Fixed visible value for negative filters, including the ! symbol
  imagen

Fixes https://github.com/grafana/logs-drilldown/pull/1752#pullrequestreview-4266778713

### DIFF
--- a/src/Components/IndexScene/LevelsVariableScene.tsx
+++ b/src/Components/IndexScene/LevelsVariableScene.tsx
@@ -72,12 +72,19 @@ export class LevelsVariableScene extends SceneObjectBase<LevelsVariableSceneStat
         return [];
       }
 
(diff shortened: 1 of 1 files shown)


L2D2Grafana (approved):
Thanks for fixing this! I remember co-pilot mentioning this once before when I updated combobox. I took a look at it but the old code always had equals = true. It worked in ops with the old table as well. 

Ops behaving as expected.
Image

Image

Local seems good with the new fixes

Image

One tiny thing if the user unchecks a box the old !value shows but if you select it again it's positive. I think you are right it might be good to update this component in the future to something with an operator selector.

Image

PR #1899, opened 2026-05-19
fix(lineFilters): cancel debounced requests when onToggleExclusive changes

L2D2Grafana:
### Description ✨
Fixes https://github.com/grafana/support-escalations/issues/22394

## Summary 📝

- Fixes #22394: switching line filter **Include** / **Exclude** immediately after typing no longer resets the filter when the 2s text debounce is still pending.
- Cancels the pending debounced line-filter update in `onToggleExclusive` so the include/exclude operator change is not overwritten by a stale callback.
- Adds a unit regression test covering type → toggle exclude → advance debounce timer.

## Test 🧪

- [x] `npx jest src/Components/IndexScene/LineFilter/LineFilterVariablesScene.test.ts` (7 tests passed)
- [x] Manual: open Logs Drilldown → Logs tab → type in line filter → switch Include/Exclude before ~2s elapses → filter text and exclude/include mode stay in sync; query reflects exclude (`!>`) without reverting to include
- [x] Manual: blur line filter or wait for debounce, then toggle Include/Exclude — behavior unchanged (still works as before)

### DIFF
--- a/src/Components/IndexScene/LineFilter/LineFilterVariablesScene.test.ts
+++ b/src/Components/IndexScene/LineFilter/LineFilterVariablesScene.test.ts
@@ -1,11 +1,20 @@
+import { ChangeEvent } from 'react';
+
 import { AdHocFilterWithLabels } from '@grafana/scenes';
 
 import {
   applyLogSelectionToLineFilters,
   lineFilterRowIsEmpty,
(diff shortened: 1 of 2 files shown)


matyax (approved):
👌

Issue #1542, opened 2025-09-18
Max lines error state: clear filters should reset "max lines" input


matyax:
When max_lines errors, we correctly show an error state, but the action is to "clear filters", which doesn't really help the user.

<img width="1051" height="702" alt="Image" src="https://github.com/user-attachments/assets/9d7c111e-8237-4237-ab07-2cac8b9447a9" />

matyax:
Recently improved. No further changes planned.

PR #1902, opened 2026-05-20
fix(lineLimit): show invalid for values over the max line limit

L2D2Grafana:
### Description
- Fixes https://github.com/grafana/support-escalations/issues/22408
- Right now I show invalid if the value goes over the limit. I could allow users to go over the limit see the error and clear filters, but it seemed better to preempt the error state.

## Summary 📝

- Fixes #22408: entering a line limit above the Loki max (e.g. 6000 via custom value) no longer persists to `localStorage` or runs queries; the field stays visible with invalid styling so users can correct it.
- **Clear filters** now removes the saved line-limit override (`clearMaxLines`), resets the line limit UI to datasource defaults, and re-runs queries so the max-entries error state is recoverable without manual storage edits.
- Adds `getMaxLinesLimit` (Loki config → datasource default → highest preset) and excludes `*.test` / `*.spec` files from lint-staged cspell to avoid false hook failures.

## Test 🧪

- [x] `pnpm test src/Components/ServiceScene/LineLimitScene.test.tsx src/services/variableHelpers.test.ts` (17 tests passed)
- [x] Line limit combobox: enter a value above the stack max (e.g. 6000) → field shows invalid, no new network request, value not saved after reload from that action
- [x] Line limit combobox: enter a valid value (e.g. 2000) → persists and logs query runs
- [x] With a bad line limit in `localStorage` and extra filters: **Clear filters** → line limit resets to default, error clears, queries run again

Screenshot 2026-05-20 at 10 08 52 AM

### DIFF
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -3,6 +3,6 @@ const PRETTIER_WRITE = 'prettier --write --ignore-path .prettierignore';
 module.exports = {
   'src/**/*.{js,jsx,ts,tsx}': 'pnpm run lint',
   '**/*': ['eslint --fix', 'bash -c tsc-files --noEmit', PRETTIER_WRITE],
-  'src/**/!(*locales)/**/*.{ts,tsx,js,go,md,mdx,yml,yaml,json,scss,css}':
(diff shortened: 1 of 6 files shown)


matyax:
This should also fix https://github.com/grafana/logs-drilldown/issues/1542

matyax (approved):
LGTM and works as expected. Good job!

PR #1895, opened 2026-05-15
feat(LogsListScene): Allow logs panel to fit the screen height with an option to use full screen size

matyax:
After many complaints about scrolls, we're removing the need for scrolling to see the bottom of the logs panel, and the panel height will adapt to the available space. Demo:

https://github.com/user-attachments/assets/950cc3f7-d86f-4e72-8984-31c7fec76152

Additionally, it adds a new option to optionally use full screen height for logs:

https://github.com/user-attachments/assets/486342f2-030e-4a9b-b993-baecf663cd01

PR #1903, opened 2026-05-20
fix(BreadownView): remove hardcoded min height

matyax:
Follow up to https://github.com/grafana/logs-drilldown/pull/1895, removing a hardcoded height.

Before:

Before

After:

After


L2D2Grafana (commented):
👍

L2D2Grafana (approved):

PR #1932, opened 2026-06-08
feat(ExpandCondenseLosgsView): Move to Panel Menu

matyax:
Follow up to https://github.com/grafana/logs-drilldown/pull/1895

The UI is already complex enough, and this shouldn't be an option that needs to be permanently visible to be toggled (and I didn't really like it there anyway), so moving to Panel Menu.

Demo:

https://github.com/user-attachments/assets/69922f6d-e0e5-4377-a882-d82c7337e514


L2D2Grafana (approved):
👍 

Suggestions
- Add to docs/sources/view-logs/_index.md that expand/condense is under the panel menu (⋯) → Interface, since this is a discoverability change.

matyax:
Oh yes, great idea! There's no mention of that triple-dot menu, so I'll open a PR to work with @knylander-grafana

PR #1937, opened 2026-06-09
fix(ServiceSelectionTabsScene): rename add tab action

matyax:
Before:

imagen

After:

imagen

Making sure the purpose of the button is self-explainatory to new users.


L2D2Grafana (dismissed):
👍 
This broke the e2e tests looking for the old name 

Image

L2D2Grafana (approved):

PR #1953, opened 2026-06-12
fix(ServiceSelection): fix broken context queries in due to trailing commas

matyax:
Follow up to https://github.com/grafana/logs-drilldown/pull/1940, fixing broken queries with trailing commas in the Service Selection page, when Log Context was used.


L2D2Grafana (approved):
👍 Nice, not seeing the logs context error anymore. Some tests could be a nice addition.

PR #1969, opened 2026-06-22
fix(generateLinkFromFilters): fix potentially duplicated filters

matyax:
Before these changes, filters (specially level) could hold duplicated values because we were building urlParams from the current URL params. Example of the bug:

https://github.com/user-attachments/assets/7c2b766d-49dc-48bc-b236-090187ac864b


L2D2Grafana (approved):
LGTM, thanks for adding tests.

PR #1978, opened 2026-06-26
chore: add warnOnConflicts to i18next config

shelldandy:
## Summary
- Add `warnOnConflicts: 'error'` to `i18next.config.ts` to align with core Grafana's config
- This will catch translation key conflicts during extraction rather than silently overwriting

## Test plan
- [x] `pnpm i18n-extract-check` passes with no changes


L2D2Grafana (approved):
👍

PR #1985, opened 2026-07-03
chore: create new plugin info menu

joey-grafana:
### ✨ Description

Related issue(s): https://github.com/grafana/profiles-drilldown/issues/973

Create new hew plugin info menu to better align with other Drilldown apps.

### 🧪 How to test?

Before 
Screenshot 2026-07-03 at 09 16 38

After
Screenshot 2026-07-03 at 09 16 12


package.json
```diff
@@ -106,6 +106,7 @@
     "webpack-bundle-analyzer": "4.10.2",
     "webpack-cli": "6.0.1",
     "webpack-livereload-plugin": "3.0.2",
+    "webpack-merge": "6.0.1",
```
L2D2Grafana, line 109: Do we need webpack-merge I don't see it used in this PR?

tests/breakdowns/jsonView.spec.ts
```diff
@@ -387,11 +387,12 @@ test.describe('JSON view breakdown (nginx-json)', () => {
...
-      // Short default viewport (see setDefaultViewportSize) leaves the linked row below the fold;
-      // scroll before asserting visibility in the viewport.
+      // JSON viz scrolls within its own overflow container (not the page viewport), so assert
+      // the app scrolled to the linked line via the selectedLine URL param instead of
+      // Playwright's toBeInViewport(), which fails with nested scroll + sticky headers.
       const restoredLine = page.getByText(selectedLogLineText ?? '').first();
-      await restoredLine.scrollIntoViewIfNeeded();
-      await expect(restoredLine).toBeInViewport();
+      await expect(restoredLine).toHaveAttribute('data-scrolled', 'true', { timeout: 30_000 });
+      await expect(restoredLine).toBeVisible();
```
L2D2Grafana, line 395: Ah yes this does seem to be failing on main or flaky. This fix works, but think we can remove the timeout, since there is already a global timeout.

src/Components/App/header/PluginInfo.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+const getStyles = (theme: GrafanaTheme2) => ({
+  menuHeader: css`
+    padding: ${theme.spacing(0.5, 1)};
+    white-space: nowrap;
+  `,
+  subTitle: css`
+    color: ${theme.colors.text.secondary};
+    font-size: ${theme.typography.bodySmall.fontSize};
+  `,
```
L2D2Grafana: We've been using Object syntax for css 
```
css({
  color: theme.colors.text.secondary,
  ...(disabled && { opacity: 0.5 }),
});
```

src/Components/Header/LokiLogo.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
L2D2Grafana, line 1: I'm fine with this just being `Header` instead of `App/header`

src/Components/IndexScene/VariableLayoutScene.tsx
```diff
@@ -135,12 +136,24 @@ export class VariableLayoutScene extends SceneObjectBase<VariableLayoutSceneStat
...
+                  {!indexScene.state.embedded && (
+                    <PluginHeaderToolbar>
+                      {controls.map((control) => {
+                        return control instanceof ToolbarScene ? (
+                          <control.Component key={control.state.key} model={control} />
+                        ) : null;
+                      })}
+                      <PluginInfo variant="canvas" />
+                    </PluginHeaderToolbar>
+                  )}
```
L2D2Grafana, line 156: Logs also shows a gear icon for `exploreLogsAggregatedMetrics`. I think they look pretty good together Image

package.json
joey-grafana: Good catch — it's not new usage from this PR. webpack.config.ts already imports merge from webpack-merge to extend the Grafana base webpack config. It was previously only available transitively via webpack-cli; this just makes that dependency explicit.

It also errors if I don't explicitly import the package
Screenshot 2026-07-09 at 10 17 22

L2D2Grafana (approved):
LGTM, thanks @joey-grafana !

PR #1987, opened 2026-07-09
fix(parsePrometheusDuration): fix regexp builder.

matyax:
While working on something else I noticed an error in the console.

imagen

Investigating, I found that the group names were being lost at some point.

Regexes:

Regexes

Output:

Result

Causing errors when trying to parse durations. I'm still not sure if this is a result from minification/compilation.


L2D2Grafana (approved):
Nice find 👍 ! I was able to see this by running `pnpm build` it does not show up in local dev.

Image

PR #1991, opened 2026-07-15
feat(store): refactor expanded/collapsed logs storage

matyax:
This PR fixes two things:

1. When logs are embeded, default to the expanded view. Otherwise it results on a very tiny Logs Drilldown that might confuse users and it's not trivial to fix.
2. Remove the exploration prefix from the stored expanded/collapsed preferences. Otherwise users need to set this property every time they browse a given set of logs.

Can be tested combining the [default URL](http://localhost:3000/a/grafana-lokiexplore-app/explore) with the [embedded URL](http://localhost:3000/a/grafana-lokiexplore-app/embed).

### DIFF
--- a/src/services/store.ts
+++ b/src/services/store.ts
@@ -7,6 +7,7 @@ import { CollapsablePanelText, TimeSeriesPanelType } from '../Components/Panels/
 import { FieldsPanelsType } from '../Components/ServiceScene/Breakdowns/FieldsAggregatedBreakdownScene';
 import { SortDirection } from '../Components/ServiceScene/Breakdowns/SortByScene';
 import pluginJson from '../plugin.json';
+import { isEmbeddedLogs } from './extensions/embedding';
 import { escapePrimaryLabel } from './extensions/links';
 import { isDedupStrategy } from './guards';
 import { logger } from './logger';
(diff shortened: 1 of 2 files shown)


src/services/store.ts
```diff
@@ -570,16 +571,21 @@ export function setTableLogLine(value: LogLineState) {
 
 // Logs view size
 export function getExpandedLogsView(sceneRef: SceneObject) {
-  const PREFIX = getExplorationPrefix(sceneRef);
+  const embedded = isEmbeddedLogs(sceneRef);
+  const stored = localStorage.getItem(`${pluginJson.id}.logs${embedded ? '.embedded' : ''}.expanded`);
```
L2D2Grafana, line 575: I would call this something like localStorageIsEmbedded

L2D2Grafana (approved):
LGTM 
Seeing [expanded in the embedded view](http://localhost:3001/grafana/a/grafana-lokiexplore-app/embed?ld-patterns=%5B%5D&pageSlug=logs&ld-var-all-fields=&ld-urlColumns=%5B%5D&ld-visualizationType=%22logs%22&ld-displayedFields=%5B%5D&ld-userDisplayedFields=false&ld-var-filters=file%7C%3D%7CC:%5CGrafana%5Clogs%5Cgateway.txt&ld-sortOrder=%22Ascending%22&ld-wrapLogMessage=true&ld-prettifyLogMessage=false&ld-var-fields=&ld-var-metadata=&ld-var-jsonFields=&ld-var-lineFilters=&ld-var-levels=&ld-var-ds=gdev-loki&ld-var-patterns=) 
I'm also okay with overriding the user's preference always in the embedded view.

PR #1997, opened 2026-07-17
feat(DefaultColumns): add support for stored multiple values of the same label as a regex

matyax:
When these kind of filters are saved to the query library:

imagen

The underlying query which is saved is:

imagen

Which, when reconstructing the state on load changes to:

imagen

This PR adds suppor for these type of queries and default columns.


L2D2Grafana (approved):
LGTM

Image

PR #1998, opened 2026-07-20
fix(LogsTablePanelScene): fix missing title in Table NG

matyax:
Not only the title was missing, but also "Title" was the hardcoded string that it displayed. Now shows the count, like the Logs panel.

B:

imagen

A:

imagen


L2D2Grafana (approved):
🧹

PR #2019, opened 2026-07-30
feat(LogsVolumePanel): update color for the visible range

matyax:
Since multi-lane annotations were enabled, the visible range could no longer be displayed as a region, making it much less visible.

imagen

We're now using a higher contrast color, that is also neutral, so it's not confused to any type of log level or other type of situation.

Light theme:

Light

Dark theme:

Dark

### DIFF
--- a/src/Components/ServiceScene/LogsVolume/LogsVolumePanel.tsx
+++ b/src/Components/ServiceScene/LogsVolume/LogsVolumePanel.tsx
@@ -143,7 +143,6 @@ export class LogsVolumePanel extends SceneObjectBase<LogsVolumePanelState> {
     const isCollapsed = getLogsVolumeOption('collapsed');
     // Overrides are defined by setLogsVolumeFieldConfigOverrides, any overrides added here will be overwritten!
     const viz = PanelBuilders.timeseries()
-      .setOption('annotations', { multiLane: true })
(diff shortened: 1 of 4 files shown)


src/Components/ServiceScene/LogsVolume/LogsVolumePanel.tsx
```diff
@@ -143,7 +143,6 @@ export class LogsVolumePanel extends SceneObjectBase<LogsVolumePanelState> {
     const isCollapsed = getLogsVolumeOption('collapsed');
     // Overrides are defined by setLogsVolumeFieldConfigOverrides, any overrides added here will be overwritten!
     const viz = PanelBuilders.timeseries()
-      .setOption('annotations', { multiLane: true })
```
matyax, line 146: This was also set in line 152.

L2D2Grafana (dismissed):
🎨

L2D2Grafana (dismissed):

09jvilla:
Looks much better - thank you Matias!

L2D2Grafana (approved):

gtk-grafana:
You can re-enable annotation opacity regions by disabling the default the hide lines and areas panel option:
image

Just make sure you have a finite number of overlapping regions or the graph could get too many overlays to be legible

PR #2020, opened 2026-07-31
chore(ui-refresh): fix patterns table overflow and replace hardcoded colors with theme colors

L2D2Grafana:
Let me know if there is anything else I should fix!

 ## Summary 📝

- I think this has nothing to do with the refresh, but did it anyway. Fixes the pattern table spilling out of its container on the Patterns tab, reported on #2015. Its scroll container declared `height: calc(100vh - 450px)` with `minHeight: 470px`, while `SceneCSSGridLayout` gave every row a shared `200px` via `autoRows`. The container overflowed its cell by exactly 270px (470 − 200) with nothing clipping it, so its 11px scrollbar gutter painted **outside** the panel — reading as a stray vertical line down through the table. Replaced with a column `SceneFlexLayout` carrying per-child heights (200px chart / 470px table) so the two declarations can't disagree.
- Replaces hardcoded colors with theme colors: the log-level hexes in `getFieldMappings` and `setLevelColorOverrides` become Grafana **named** colors, the JSON search highlight uses `theme.components.textHighlight` instead of an orange with `color: 'black'`, the favourite star moves to the palette and `text.secondary`, and a `borderRadius: '5px'` becomes `theme.shape.radius.default`. Named colors matter because they resolve **per theme** — `semi-dark-red` is `#E02F44` in dark and `#C4162A` in light — so they adapt where a fixed hex can't, and `#705da0` was a v6-era value no longer in the palette at all.
- Two adjacent bugs found while in here: a module-scope `const theme = config.theme2` froze the theme at import (so those styles wouldn't follow a theme switch — exactly what the refresh does), and `overrideToFixedColor` indexed the visualization palette unbounded, so any pattern past the end of the palette got `fixedColor: undefined`. It now reads the palette per call and wraps by `palette.length`. Also drops the sticky header's `zIndex` from `navbarFixed` (1000) to `1`, which was painting over the nav and breadcrumbs.

## Test 🧪

https://github.com/user-attachments/assets/53854661-d9d7-49a5-a15d-daef9284776e

Screenshot 2026-07-31 at 8 51 34 PM

### DIFF
--- a/src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
+++ b/src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
@@ -1,6 +1,7 @@
-import React, { useMemo } from 'react';
+import React, { useLayoutEffect, useMemo, useRef, useState } from 'react';
 
 import { css, cx } from '@emotion/css';
+import { useResizeObserver } from '@react-aria/utils';
 import { CellProps } from 'react-table';
 
 import { DataFrame, GrafanaTheme2, LoadingState, PanelData, scaledUnits } from '@grafana/data';
@@ -205,7 +206,6 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
       columns.splice(1, 0, {
         header: 'Levels',
         id: 'levels',
-        // @todo custom sort method?
         cell: (props: CellProps<PatternsTableCellData>) => {
           props.cell.row.original.levels.sort();
           return props.cell.row.original.levels.map((level) => {
@@ -277,7 +277,9 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
   }
 }
 
-export const PATTERNS_TABLE_HEIGHT = '470px';
+// Small guard only — a large floor pushes the wrapper past the viewport bottom on short windows,
+// stacking a page scrollbar on top of the table's own.
+const PATTERNS_TABLE_MIN_HEIGHT = '200px';
 
 const getTableStyles = (theme: GrafanaTheme2) => {
   return {
@@ -289,8 +291,7 @@ const getTableStyles = (theme: GrafanaTheme2) => {
       '> div': {
         height: '100%',
       },
-      height: '100%',
-      minHeight: 0,
+      minHeight: PATTERNS_TABLE_MIN_HEIGHT,
       overflow: 'hidden',
       // Make table headers sticky
       th: {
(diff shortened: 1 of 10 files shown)


matyax:
I definitely think we should fix the height spilling, but we need the list to use all the available space without extra scrolls, instead of having a fixed height:

imagen

src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
```diff
@@ -208,24 +208,24 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
         // @todo custom sort method?
```
matyax:
```suggestion
```
Now that we're at it, we should clean up these todos, which should have been issues instead.

src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
```diff
@@ -208,24 +208,24 @@ export class PatternsViewTableScene extends SceneObjectBase<SingleViewTableScene
...
-              fill={'outline'}
-              className={styles.levelWrap}
-              onClick={() => {
-                props.cell.row.original.togglePatternLevel(level);
-              }}
-            >
-              {level}
-            </Button>
-          ));
+          return props.cell.row.original.levels.map((level) => {
```
matyax, line 211: I really like this! 👏

src/services/levels.ts
```diff
@@ -86,6 +86,70 @@ export function normalizeLevelName(level: string) {
   return level;
 }
 
+export const getFieldMappings = (): ValueMap => {
```
matyax, line 106: We should double check the colors match those defined here and unify the definition.

https://github.com/grafana/logs-drilldown/blob/main/src/services/panel.ts#L55

matyax (commented):
Looking good! Left a few comments and suggestions.

src/Components/ServiceScene/Breakdowns/Patterns/PatternsViewTableScene.tsx
L2D2Grafana: Removed both todos (this one and the one in `PatternsFrameScene.updatePatterns`) and filed issues to track them instead: #2024 (custom sort for the Levels column) and #2025 (children subscribing to state).

L2D2Grafana:
Fixed the fixed-height issue: the table now measures its rendered offset and sizes itself to `calc(100vh - offset)` — the same approach `LogsListScene` uses — with a small 200px floor so it never collapses entirely. A larger floor turned out to cause double scrollbars on short windows (the wrapper got pushed past the viewport bottom, adding a page scrollbar on top of the table's own), so the floor is intentionally minimal.

Verified locally:
- 1600px viewport: table grows to fill down to 16px above the viewport bottom — no page scroll, no inner scroll when the patterns fit
- 1200px viewport: fills to the bottom, overflow scrolls inside the table with the scrollbar contained (no gutter spill from #2015)
- 930px viewport (short window, 113-pattern-scale data): single scrollbar — the table's own — with no page scroll

Not sure what else to fix here. The double scroll is gone unless you have a very small screen. There has to be some kind of min height on the table.
Screenshot 2026-08-04 at 8 04 45 AM
Screenshot 2026-08-04 at 8 05 22 AM

I don't think using the flexbox grow will work through the scenes components because there is no definite height anywhere.

matyax (approved):
Looks much better!

PR #2041, opened 2026-08-05
fix(assistant): update to latest and fix context bugs

L2D2Grafana:
## Summary 📝

- Fixes assistant context provisioning: filter operators are passed on context items instead of encoded into values, initial context is registered on activation (not after the first filter change), metadata/line-filter/pattern/time-range changes now refresh the context, and `assistantInitialized` resets on deactivation so the pathname-cached scene re-registers.
- Adds a hidden "Current logs query and time range" context item carrying the fully rendered LogQL and unix-ms window — shaped 1:1 to the assistant's Loki tool `query`/`start`/`end` params, with instructions to pass the timestamps through without converting.
- Bumps `@grafana/assistant` 0.1.25 → 0.1.33.

### Run it locally

1. Run local core Grafana with a plugin path that includes this plugin's `dist/` **and** a `grafana-assistant-app` checkout, with the assistant feature toggles and unsigned-plugin allowance in your `custom.ini`, plus assistant provisioning (`conf/provisioning/plugins/assistant.yaml` → `backendUrl: http://localhost:9091`).
2. In `grafana-assistant-app`: build the plugin backend (`cd apps/plugin && mage -v`), start the API stack (`docker compose --env-file .env up db api nats` — `.env` needs `ANTHROPIC_API_KEY`), and run the plugin frontend watch (`pnpm --filter @grafana/assistant-plugin run dev`).
3. In this repo: `pnpm dev`, and `docker compose -f docker-compose.local.yaml up -d loki generator` for test data.
4. Start Grafana with `GF_PLUGINS_FORWARD_HOST_ENV_VARS=grafana-assistant-app GF_DEFAULT_APP_MODE=development make run` (plus the frontend build).
5. Sanity check: `curl -su admin:admin localhost:3000/api/plugins/grafana-assistant-app/settings` — the version must match `apps/plugin/dist/plugin.json` (a stale assistant build elsewhere in the scan path can shadow it).

To inspect what actually reaches the assistant, watch the `messages?audience=user` POST in DevTools → Network — `contextItems` is the ground truth (chips only show non-hidden items).

### Scenarios to verify

- **Initial context:** load a drilldown logs page fresh, open the assistant, start a new conversation — filter chips appear without touching any filter.
- **Operators:** add `!=`/`=~` label filters, an excluded level, a metadata exclude (e.g. `pod != …`), and a fields exclude — each arrives with a clean value plus a separate `operator` field (metadata/fields items are hidden; check the payload).
- **Subscriptions:** mid-conversation, include a pattern, add/remove a line filter, and change the time range — the "Current logs query and time range" item tracks each change (`|>` stage appears, window updates).
- **Re-registration:** navigate to another app and back to drilldown, open the assistant — context is present and current (pre-fix, the pathname-cached scene stopped registering).
- **Panel action:** "Explain in Assistant" on the logs/volume panels — the data summary reflects the visible window, not the last hour; asking "what's the time range for this query" should answer from context without epoch→ISO conversion in the thinking trace.

Dev 
Here you can see !=dev-eu-west
Screenshot 2026-08-05 at 9 42 58 AM

Updated 
Here the operator != is sent automatically instead of parsded
Screenshot 2026-08-05 at 2 41 11 PM

### DIFF
--- a/src/Components/IndexScene/IndexScene.tsx
+++ b/src/Components/IndexScene/IndexScene.tsx
@@ -394,6 +394,8 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     return () => {
       clearKeyBindings();
       assistantUnregister.forEach((callback) => callback.unregister());
+      // Scenes re-activates this pathname-cached instance — reset so we re-register.
+      this.assistantInitialized = false;
     };
   }
 
@@ -543,6 +545,9 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
   private provideAssistantContext() {
     const setAssistantContext = providePageContext(`${PLUGIN_BASE_URL}/**`, []);
 
+    // Set the initial context right away — the subscriptions below only fire on changes.
+    updateAssistantContext(this, setAssistantContext);
+
     this._subs.add(
       getDataSourceVariable(this).subscribeToState(async () => {
         await updateAssistantContext(this, setAssistantContext);
@@ -561,6 +566,28 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     this._subs.add(
       getFieldsVariable(this).subscribeToState(async () => {
         await updateAssistantContext(this, setAssistantContext);
+      })
+    );
+    this._subs.add(
+      getMetadataVariable(this).subscribeToState(async () => {
+        await updateAssistantContext(this, setAssistantContext);
+      })
+    );
+    this._subs.add(
+      getLineFiltersVariable(this).subscribeToState(async () => {
+        await updateAssistantContext(this, setAssistantContext);
+      })
+    );
+    this._subs.add(
+      getPatternsVariable(this).subscribeToState(async () => {
+        await updateAssistantContext(this, setAssistantContext);
+      })
+    );
+    this._subs.add(
+      sceneGraph.getTimeRange(this).subscribeToState(async (newState, prevState) => {
+        if (newState.value !== prevState.value) {
+          await updateAssistantContext(this, setAssistantContext);
+        }
       })
     );
     this.assistantInitialized = true;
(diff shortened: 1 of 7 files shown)


src/Components/IndexScene/IndexScene.tsx
```diff
@@ -394,6 +394,8 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
     return () => {
       clearKeyBindings();
       assistantUnregister.forEach((callback) => callback.unregister());
+      // Scenes re-activates this pathname-cached instance — reset so we re-register.
```
matyax:
```suggestion
```

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -543,6 +545,11 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
   private provideAssistantContext() {
     const setAssistantContext = providePageContext(`${PLUGIN_BASE_URL}/**`, []);
 
+    // Set the initial context right away (fire-and-forget) — the subscriptions below only fire on changes.
```
matyax:
```suggestion
```

src/Components/IndexScene/IndexScene.tsx
```diff
@@ -563,6 +570,28 @@ export class IndexScene extends SceneObjectBase<IndexSceneState> {
...
+    this._subs.add(
+      getLineFiltersVariable(this).subscribeToState(async () => {
+        await updateAssistantContext(this, setAssistantContext);
+      })
+    );
+    this._subs.add(
+      getPatternsVariable(this).subscribeToState(async () => {
+        await updateAssistantContext(this, setAssistantContext);
+      })
+    );
```
matyax: Probably worth checking if the states are different to prevent unnecessary calls. Some of these can fire multiple times with no changes.

src/services/assistant.ts
```diff
@@ -38,7 +38,8 @@ export const updateAssistantContext = async (
         createAssistantContextItem('label_value', {
           datasourceUid: ds.uid,
           labelName: filter.key,
-          labelValue: `${inequalityPrefix(filter.operator)}${stripAdHocFilterUserInputPrefix(filter.value)}`,
+          labelValue: stripAdHocFilterUserInputPrefix(filter.value),
+          operator: filter.operator,
```
matyax, line 42: This is great! This has been pending on our side for quite some time.

matyax (commented):
Looks good, left some comments.

matyax (approved):

PR #2056, opened 2026-08-18
fix(LogList): show the exact log count in volume and logs panels

matyax:
Before:

imagen

After:

imagen

The tab remains using the short number to prevent it from growing too much for very high volume. The correct count can be found in Logs Volume.

### DIFF
--- a/src/Components/ServiceScene/ActionBarScene.tsx
+++ b/src/Components/ServiceScene/ActionBarScene.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import { css, cx } from '@emotion/css';
 
-import { getValueFormat, GrafanaTheme2 } from '@grafana/data';
+import { formattedValueToString, getValueFormat, GrafanaTheme2 } from '@grafana/data';
 import { SceneComponentProps, sceneGraph, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
 import { Box, Stack, Tab, TabsBar, useStyles2 } from '@grafana/ui';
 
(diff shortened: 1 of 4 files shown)


bohandley (approved):
Exact! Nice! 🚀

09jvilla:
total-log-volume



> The tab remains using the short number to prevent it from growing too much for very high volume. The correct count can be found in Logs Volume.

This makes sense and is great! 

My only question from the 'After' screenshot was why the `Total: 103` (circled in green) didn't match the Log volume value at the top (464,628). Shouldn't those quantities match?

hugohaggmark (approved):
This is a great improvement! 🎉

Just out of curiosity, are Logs and Log volume reserved strings, or should we consider translations for these as well?

matyax:
> My only question from the 'After' screenshot was why the Total: 103 (circled in green) didn't match the Log volume value at the top (464,628). Shouldn't those quantities match?

@09jvilla Those quantities should always match, or at least be approximately equal.

imagen

The logs count, in particular the one shown in the tabs, is resolved using a Loki instant metric query, which I recently realized that is pretty much always returning a different number than the one that the non-instant metric query and the logs query will get. In general those numbers always differ.

I just shared a reproduction of this with @trevorwhitney  to see if there's anything that needs to be fixed, or if we should let the user know that it is expected by some underlying reason.

09jvilla:
> I just shared a reproduction of this with @trevorwhitney to see if there's anything that needs to be fixed, or if we should let the user know that it is expected by some underlying reason.

Sounds good. Thanks Matias.

PR #2057, opened 2026-08-20
fix(LogsVolumePanel): improve accuracy of results in the panel title displayed count

matyax:
Follow up to https://github.com/grafana/logs-drilldown/pull/2056

As there are scenarios where the instant query can return inaccurate numbers, we're further updating the Logs Volume count to:

- If collapsed, show the instant query result in short format (matching the tab)

Collapsed

- If expanded, use the count from the Logs Volume metric query response, which will be accurate (and match the visualization)

Expanded

- If the logs result is below the line limit (matching a few logs), use the count from the Logs response

Small numbers

Additionally:

- When the short version is used in the Tab or Panel title, it uses Grafana's value formatter which adds a space before the unit, matching the format from the series legends.
- Fixes a bug where the ds instance held by IndexScene was out of date, affecting maxLines.

### How to test

- Select a label to go to the service details
- If there are 1493 logs and the line limit is 1000:
  - The Logs tab should show a short version (e.g. 1.4 K)
  - Logs volume show 1493, which should match the count of every log level
  - The logs visualization should show 1000 and have 1000 logs in the panel.
- Change the line limit, zoom in or out using Logs Volume, these counts should be consistent.

Get logs using Graft or in Core Grafana with `make devenv sources=loki`

### DIFF
--- a/src/Components/ServiceScene/ActionBarScene.tsx
+++ b/src/Components/ServiceScene/ActionBarScene.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import { css, cx } from '@emotion/css';
 
-import { getValueFormat, GrafanaTheme2 } from '@grafana/data';
+import { formattedValueToString, getValueFormat, GrafanaTheme2 } from '@grafana/data';
 import { SceneComponentProps, sceneGraph, SceneObjectBase, SceneObjectState } from '@grafana/scenes';
 import { Box, Stack, Tab, TabsBar, useStyles2 } from '@grafana/ui';
 
@@ -191,22 +191,12 @@ function LogsCount(
   const valueFormatter = getValueFormat('short');
 
   // The instant query (totalCount) doesn't return good results for small result sets, if we're below the max number of lines, use the logs query result instead.
-  if (totalCount === undefined && logsCount !== undefined && logsCount < maxLines) {
+  if (logsCount !== undefined && logsCount < maxLines) {
     const formattedCount = valueFormatter(logsCount, 0);
-    return (
-      <span className={cx(className, styles.logsCountStyles)}>
-        {formattedCount.text}
-        {formattedCount.suffix?.trim()}
-      </span>
-    );
+    return <span className={cx(className, styles.logsCountStyles)}>{formattedValueToString(formattedCount)}</span>;
   } else if (totalCount !== undefined) {
     const formattedTotalCount = valueFormatter(totalCount, 0);
-    return (
-      <span className={cx(className, styles.logsCountStyles)}>
-        {formattedTotalCount.text}
-        {formattedTotalCount.suffix?.trim()}
-      </span>
-    );
+    return <span className={cx(className, styles.logsCountStyles)}>{formattedValueToString(formattedTotalCount)}</span>;
   }
 
   return <span className={cx(className, styles.emptyCountStyles)}></span>;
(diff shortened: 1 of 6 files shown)


src/Components/ServiceScene/LogsVolume/LogsVolumePanel.tsx
```diff
@@ -116,6 +125,16 @@ export class LogsVolumePanel extends SceneObjectBase<LogsVolumePanelState> {
...
+      return sumLogsVolumeSeries(panelData.series);
+    }
+
+    const serviceScene = sceneGraph.getAncestor(this, ServiceScene);
+    return serviceScene.state.totalLogsCount;
+  };
+
   private setCollapsed(collapsed: boolean | undefined, panel: VizPanel) {
     if (collapsed) {
       panel.setState({
```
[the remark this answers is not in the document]
matyax: This has never happened in my tests.

src/Components/ServiceScene/LogsVolume/LogsVolumePanel.tsx
```diff
@@ -101,21 +109,48 @@ export class LogsVolumePanel extends SceneObjectBase<LogsVolumePanelState> {
...
+    const logsCount = serviceScene.state.logsCount;
+
     const indexScene = sceneGraph.getAncestor(this, IndexScene);
     const maxLines = indexScene.state.ds?.maxLines ?? LINE_LIMIT;
-    const valueFormatter = getValueFormat('locale');
-    const formattedTotalCount = totalLogsCount !== undefined ? valueFormatter(totalLogsCount, 0) : undefined;
+
+    // Potentially inaccurate, wait for logsCount
+    if (totalLogsCount !== undefined && totalLogsCount < maxLines && logsCount === undefined) {
+      return 'Log volume';
```
hugohaggmark: nit: should all these title strings be translatable?

src/Components/ServiceScene/ActionBarScene.tsx
```diff
@@ -191,22 +191,12 @@ function LogsCount(
...
   } else if (totalCount !== undefined) {
     const formattedTotalCount = valueFormatter(totalCount, 0);
-    return (
-      <span className={cx(className, styles.logsCountStyles)}>
-        {formattedTotalCount.text}
-        {formattedTotalCount.suffix?.trim()}
-      </span>
-    );
+    return <span className={cx(className, styles.logsCountStyles)}>{formattedValueToString(formattedTotalCount)}</span>;
   }
```
hugohaggmark, line 200: nit: without looking too closely at the details here, it looks like we're repeating some of the logic from src/Components/ServiceScene/LogsVolume/LogsVolumePanel.tsx. Could we extract this into a common/shared utility?

hugohaggmark (dismissed):

src/Components/ServiceScene/LogsVolume/LogsVolumePanel.tsx
matyax: Ehhh, I would be lying if I said no.

src/Components/ServiceScene/ActionBarScene.tsx
matyax: Good question. It's similar, as this one uses the instant query and Logs Volume is slightly more complex, but I don't mind living with a bit of duplication for now.

matyax:
@hugohaggmark Can I bother you with another look? Thanks!

hugohaggmark (approved):
Looks great 🎉

09jvilla:
@trevorwhitney - I see what Matias did with the collapsed vs uncollapsed behavior and reading it more closely, it makes sense to me. I think its reasonable that when collapsed you show the rounded number and that when expanded you show the more precise one. That matches with my mental model of "when I expand, I expect more details/more information."

matyax:
@09jvilla and @trevorwhitney and matches the underlying query behavior: when collapsed, we don't run a Volume query that we can use to accurately display the count, but we can closely approximate it with the instant one.

PR #2058, opened 2026-08-21
fix(VisualRefresh): Fixes background in controls header

torkelo:
Progress on https://github.com/grafana/logs-drilldown/issues/2015 

Fixes this background (also removes the box shadow, not needed and feels inconsistent as the service view logs does not have it) 

Screenshot 2026-08-21 at 14 40 07


matyax (approved):
Thanks!

